PR #53, opened 2017-12-15
Add an estimated ratio

Hyleus:


Hyleus:
Nope, because users might want to just go trough different torrents and it would be helpful to know for them if they can download it safely.

PR #59, opened 2017-12-18
Fix the URL matcher

Hyleus:


HDVinnie:
I am approving this but noting that we need to try and fix this issue upstream within the decoda package itself.

PR #69, opened 2017-12-22
Fix for MediaInfo stuff. Helper works! But View has to be tested (!)

TomArrow:
Did test the MediaInfo helper with a local script, but I don't have the environment for the whole Laravel thing, so someone will have to test whether the view works as it is. I didn't get any syntax errors, so I hope there won't be any major issues.

Changed up a couple of things, like adding the "title" property to the display and changing "codec" to "format" for subtitles (because it doesn't work well for m2ts files).

Also, I removed the blanket strtolower and added it where it was necessary (to section and property recognition and some of the property parsers). The upside of this is that a lot of MediaInfo information that uses both uppercase and lowercase is now fixed.

Added a shitton of security tests for non-existent properties etc. (both in view and helper), so there should be no more exceptions.

### DIFF
--- a/app/Helpers/MediaInfo.php
+++ b/app/Helpers/MediaInfo.php
@@ -297,7 +297,7 @@ class MediaInfo
     {
         $output = ["general"=>[],"video"=>[],"audio"=>[]];
 
-        $generalCrumbs = array("format"=>"ucfirst","duration"=>NULL);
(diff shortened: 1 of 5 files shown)


HDVinnie:
thanks ill review this today when home before pushing and double check to make sure all works.

TomArrow:
Cool stuff. Ryuu said he might check it for errors as well, so with a bit
of luck not much work will be left for ya. :)

PR #76, opened 2017-12-23
Add a source value to torrents

Hyleus:


ScroogeHD117:
Just curious, does this change the hash value of the torrent?

HDVinnie:
yes is used to generate an unique infohash

PR #102, opened 2018-01-04
Fix tip message and typos

Hyleus:


HDVinnie:
Wouldn't it be better to use validator/rules for this?
Is a config option really need for min tip amount?

Hyleus:
Yeah, I think so. Because on other sites a BON might have a different value, and it basically stops bumping the torrents with values like 1, 2, 3 if set to a high enough value so any admin can decide if they want to allow it.

Hyleus:
>Wouldn't it be better to use validator/rules for this?

What do you mean with that?

Hyleus:
OK, removed that. I've removed that commit. There's still the issue with the messages I've updated that.

PR #114, opened 2018-01-16
 Replace the URL matcher everywhere 

Hyleus:


HDVinnie:
Can't you remove the decoda facades though in the models since it's all in the helper now. No reason to have the use Decoda\Decoda call anymore in models.

PR #121, opened 2018-01-19
EN,PL lang some fixes / addings

VerioPL:
This is some cosmetic code improvements in EN lang and adding some transations for PL


HDVinnie:
Thanks

PR #122, opened 2018-01-19
Adding some new variables to views

VerioPL:
Modified Articles and Auth views, preparing langs for EN and PL


HDVinnie:
I REALLY appreciate this......this one thing I have been slacking on.......there are still many {{ trans('') }} 
variables that need to be put into all the views and main English lang files. Then can actually update all the other lang types.

VerioPL:
I want to do that on other views ;) Just gimme some free time :)

PR #165, opened 2018-01-25
(Feature) TwoStepAuth System

HDVinnie:
- Optional TwoStep user authentication that is configurable and customizable. It uses email notifications to send the user an email with a 4-digit verification code. So if they have TwStep enabled after they login with username and password they get redirected to 2 step verificsation page before getting full access to site.


app/Traits/TwoStep.php
```diff
@@ -0,0 +1,242 @@
...
+    {
+        $user = Auth::User();
+
+        if ($user) {
+
+            $twoStepAuthStatus = $this->checkTwoStepAuthStatus($user->id);
+
+            if ($twoStepAuthStatus->authStatus !== true) {
+                return false;
+            } else {
```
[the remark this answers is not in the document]
HDVinnie: resolved in 12763f7079c4fbe80ef67f5386842d0b6be15dea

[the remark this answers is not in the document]
HDVinnie: fixed in 750f635

app/Http/Controllers/Auth/TwoStepController.php
```diff
@@ -0,0 +1,225 @@
...
+
+            $this->resetActivationCountdown($this->_twoStepAuth);
+
+            $returnData = [
+                'nextUri' => session('nextUri', '/'),
+                'message' => trans('auth.titlePassed'),
+            ];
+
+            return response()->json($returnData, 200);
+        } else {
```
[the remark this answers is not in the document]
HDVinnie: fixed in 750f635

app/Traits/TwoStep.php
```diff
@@ -0,0 +1,241 @@
...
+            $twoStepAuthStatus = $this->checkTwoStepAuthStatus($user->id);
+
+            if ($twoStepAuthStatus->authStatus !== true) {
+                return false;
+            }
+            if ($this->checkTimeSinceVerified($twoStepAuthStatus)) {
+                return false;
+            }
+
+            return true;
```
Hyleus: Remove the return true line, it still uses the the return statement at the end.

config/auth.php
```diff
@@ -109,4 +109,32 @@
...
+    'verificationEmailFromName' => ' 2-Step Verification',
+
+    /*
+    |--------------------------------------------------------------------------
+    | Verification Timings Settings
+    |--------------------------------------------------------------------------
+    */
+
+    'TwoStepExceededCount'             => 3,
+    'TwoStepExceededCountdownMinutes'  => 1440,
```
Hyleus: You could replace the 1440 with 60 * 24, so it's clear that it is one day.

public/js/twostep.js
```diff
@@ -0,0 +1,194 @@
...
+          if (self.hasClass('last-input')) {
+            $('#submit_verification').focus();
+          } else {
+            self.parent().parent().next().find('input').focus();
+          }
+        }, 1);
+      }
+
+    });
+    // Check for cop and paste
```
Hyleus: Typo, should be copy and paste.

public/js/twostep.js
```diff
@@ -0,0 +1,194 @@
...
+      var specialChars = [43, 61, 186, 187, 188, 189, 190, 191, 192, 219, 220, 221, 222];
+
+      // Numbers
+      var numbers = [48, 49, 50, 51, 52, 53, 54, 55, 56, 57];
+
+      var preCombined = controls.concat(numbers);
+      var combined = preCombined;
+
+      // Allow Letter
+      for (var i = 65; i <= 90; i++) {
```
Hyleus: Replace that with a range,  and concat it with the combined array later. http://php.net/manual/en/function.range.php

public/js/twostep.js
```diff
@@ -0,0 +1,194 @@
...
+      var self = $(this);
+
+      // Keyboard Controls
+      var controls = [8, 16, 18, 17, 20, 35, 36, 37, 38, 39, 40, 45, 46, 9, 91, 93, 224, 13, 127, 27, 32];
+
+      // Special Chars
+      var specialChars = [43, 61, 186, 187, 188, 189, 190, 191, 192, 219, 220, 221, 222];
+
+      // Numbers
+      var numbers = [48, 49, 50, 51, 52, 53, 54, 55, 56, 57];
```
Hyleus: You should use here a range, no need to enumerate them manually.

app/Traits/TwoStep.php
```diff
@@ -0,0 +1,241 @@
...
+     * @param integer $length
+     * @param string $prefix
+     * @param string $suffix
+     *
+     * @return string
+     */
+    private function generateCode(int $length = 4, string $prefix = '', string $suffix = '')
+    {
+        for($i = 0; $i < $length; $i++){
+            $prefix .= random_int(0,1) ? chr(random_int(65, 90)) : random_int(0, 9);
```
Hyleus: This might be better if you have an array with codeChars, which enumerates the chars allowed, some people might want to change this later in the config. Then you could replace it then with:
```php
$prefix = codeChars[random_int(0, count(codeChars-1)];
```

config/auth.php
```diff
@@ -109,4 +109,32 @@
...
+    'TwoStepVerifiedLifetimeMinutes'   => 360,
+    'TwoStepTimeResetBufferSeconds'    => 360,
```
Hyleus: Here the same: 6 * 60.

PR #170, opened 2018-01-31
Translation en/pl of partials views

VerioPL:


HDVinnie:
codacy is being stupid and slow......PR approved.

PR #177, opened 2018-02-06
Change sorting for topics

Hyleus:
This has the effect that topics with new replies get bumped to the top.


HDVinnie:
This is no good because the updated_at column changes when the topic is viewed and view count gets updated. So it won't work for topics with new comments as your thinking.

Hyleus:
Fixed

HDVinnie:
shouldn't you also be setting the new last_reply_at in the newTopic function?

Hyleus:
OK, updated it again.

PR #224, opened 2018-03-15
(Update) Refactor General and Staff Controllers

HDVinnie:
- Remove all facades use besides mail
- Use Dependency Injection for Illuminate\Http\Request
- use helpers for auth, cache, validator, and more to rid of facades use
- use $request->input() over $request->get()
- use $request->isMethod('POST') over $request->getMethod('POST')
- general cleanup


HDVinnie:
The faceted function in the TorrentController and RequestController needs to be optimized. Right now its a lot of code and could be crunched down. Maybe something like:

```
protected function filterTorrentsByRequestParams(Request $request, Torrent $torrent, array $params = [])
    {
        $torrent = $torrent->newQuery();

        foreach ($params as $param) {
            if ($request->has($param) && $request->get($param) != null) {
                $torrent->where($param, $request->get($param));
            }
        }

        return $torrent->get();
    }
```

Cleaner?

PR #239, opened 2018-03-27
Feature/toggle chat visibility

### LINKED ISSUE #211: Add an option to toggle the chat visbility
Currently users can see the chat the whole time, but some users might prefer to hide the chat. This can be simply done by adding an option to the settings page and hiding the chatbox if it's set, and also it shouldn't obviously try to fetch new messages then.

pxgamer:
Closes #211


HDVinnie:
Thanks!

PR #265, opened 2018-04-23
Updated the captcha config to use .env keys. Defaults to test keys.

macmedia:
config/captcha.php is out of sync with git when you add keys directly to the file. Updated to use the .env file. The default also uses the test keys provided by google if no site keys are configured.

```
env('CAPTCHA_SECRETKEY')
env('CAPTCHA_SITEKEY')
```

Captcha Logo

[https://developers.google.com/recaptcha/docs/faq](url)

### DIFF
--- a/config/captcha.php
+++ b/config/captcha.php
@@ -25,10 +25,11 @@ return [
     /*
      |------------------------------------------------------------------------------------------------
      |  Credentials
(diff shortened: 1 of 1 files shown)


HDVinnie:
Thanks for the contribution.

PR #273, opened 2018-04-28
Torrent filter by uploader returns 404 error

macmedia:
## Problem
When doing a filter for torrents by a uploader (if not found), it returns:
`Symfony\\Component\\HttpKernel\\Exception\\NotFoundHttpException` when firstOrFail() dies. This returns a 404 error

Because of the 404 return in the Ajax call, the spinner never stops and it looks like a search is still running.

screen shot 2018-04-28 at 3 13 16 pm

## Possible Solution
Return a empty result array (to replace the spinner) and a count of 0 if the uploader is not found.

### DIFF
--- a/app/Http/Controllers/TorrentController.php
+++ b/app/Http/Controllers/TorrentController.php
@@ -412,7 +412,7 @@ class TorrentController extends Controller
 
         if ($request->has('uploader') && $request->input('uploader') != null) {
             $match = User::where('username', 'like', $uploader)->first();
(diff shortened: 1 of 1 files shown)


macmedia:
Do you know why the test failed on the second update? After added the space after IF

poppabear8883:
Honestly, I am unsure ...

poppabear8883:
I'm forcing it to recheck it ... we will see

PR #338, opened 2018-06-26
(Fix) Group when account age is not met

werrpy:
This removes the upper limit on upload amount.

For example, if a user had >=100TB upload and an account age > 6 months but < 1 year, they would be Member rank. This fix makes them fall into the BluLegend group.

### DIFF
--- a/app/Console/Commands/autoGroup.php
+++ b/app/Console/Commands/autoGroup.php
@@ -69,22 +69,22 @@ class autoGroup extends Command
             }
 
             //BluMember >= 1TB but < 5TB and account 1 month old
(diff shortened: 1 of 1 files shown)


HDVinnie:
This will not work. There is a reason the upper limits are in place. Your fix would actually break the system.

patkub:
Is the issue setting and saving group id twice? Because I think that already happens with bluseeder/bluarchivist.

Is it possible to fix this for users who fall outside upload limit?

werrpy:
Is the issue setting and saving group id twice? Because I think that already happens with bluseeder/bluarchivist.

Is it possible to fix this for users who fall outside upload limit?

PR #359, opened 2018-07-25
Dark profile and language menus to match header sidebar

werrpy:
Dark profile and language menus to match header / sidebar. Makes profile menu icons visible. Changed wishlist icon to a clipboard.


poppabear8883:
should be committing the results of `npm run prod` and **NOT** `npm run dev`

werrpy:
@poppabear8883 fixed

PR #363, opened 2018-08-04
Source emojione from npm

werrpy:
assets from emojione-assets package
emoji list for .textcomplete() from emojione-assets/emoji.json
uses local emojione pngs
Requires `npm run prod`


poppabear8883:
@werrpy I'm confused by the `.gitignore` entry for `public/vendor` ... That directory should be version controlled

werrpy:
Moved from `public/vendor/emojione` to `public/img/emojione`.

PR #420, opened 2018-10-14
User Purge System

HDVinnie:
User Purge System that gives the ability to disable and soft delete inactive user accounts. It is configurable and notifies user via email regarding account disable/prune warnings and more.


HDVinnie:
For those updating. Keep in mind this new system requires two new group shown here.

https://github.com/HDInnovations/UNIT3D/pull/420/commits/0807b3bfa6de4e287ba17d0f9af92ca51c356509

PR #425, opened 2018-10-21
(Fix) Submodules and (Update) Readme

MatheusRV:


HDVinnie:
Im fine with this. Im nit familiar with submodules but reading on it now. Im not sure its best to merge this yet since @Toothwitch is not done with the docker image. Im going to issue a review to get other core devs opinion on this. Thanks for the PR.

MatheusRV:
> Im fine with this. Im nit familiar with submodules but reading on it now. Im not sure its best to merge this yet since @Toothwitch is not done with the docker image. Im going to issue a review to get other core devs opinion on this. Thanks for the PR.

This facilitates the submission of commits in submodules, even if they are not ready. Only the installer is called in the readme, facilitating the life of those who are instalandom taking out the need of a new git clone. The docker is only in the deployment folder, it is not spoken about in readme so anyone who wants to use it will be able to edit the file and send commits to the correct repository without the need to send commits here. Submodules are pretty cool.

HDVinnie:
After thinking on this more Im not sure it makes sense. In order to make use of this one would have to clone the UNIT3D repository first and then run `cd deployment/installer && sudo ./install.sh`. In which the installer is just going to git clone UNIT3D again. It pretty much comes down to just cloning the installer as it is now and letting it do its thing or cloning UNIT3D. Using the installer in submodule and then having the installer clone UNIT3D again.

poppabear8883:
I see what you mean ... in that case I agree with you here. We could in the near future integrate the installer into the main repository and which case instead of cloning the repository via installer it simply moves the files to the right directory after everything else is setup.

MatheusRV:
No, the submodule will not clone again, when you make the repository clone the submodules become normal folders. It is not necessary to clone again.

HDVinnie:
@MatheusRV I think your mis-reading what I wrote. Its not the submodule but the installer that needs to be adapted first for this PR to make any sense.

poppabear8883:
@MatheusRV 

The method here would need to be re-written to not clone but rather move UNIT3D's src into the right directory ... 
https://github.com/poppabear8883/UNIT3D-INSTALLER/blob/af9d1c4e76ced4420731c93ba610bf94d6347528/src/Installer/UNIT3D/Unit3dSetup.php#L23

Ideally, it would be nice to see the clone to be there as an option that can be ran in `stand-alone` mode like `./install.sh --stand-alone` which would run the `clone` method and by default its expected to be run under the sub-module ...

If that makes any sense. You are welcome to submit a PR to the INSTALLER repository for us to review. Be sure to reference this issue in that PR though.

Thanks

MatheusRV:
No submodules for now?

Issue #238, opened 2018-03-27
Optimize MassPM Function


HDVinnie:
- UNIT3D Version: 1.6.4
- Laravel Version: 5.6.12
- PHP Version: 7.2
- Database Driver & Version: MySQL 5.7

### Description:
Current functionality is not realistic for a site with many members. Sending a MassPM to 10K members is running a massive amount of queries at once resulting in timeouts.

### Steps To Reproduce:
Send MassPM to all users on site. If site has ALOT of members you will most likely receive a NGINX TIMEOUT ERROR.

### Possible Fix:
There is a better way to approach this. 

1) We could just create a message and not send it. Then have pivot table message_user, which will hold IDs of users and read messages. So when a user logs into the system, the system will show all messages which don't have an entry in the pivot table as unread messages so instead of executing 10k queries all at once every user object will be handling its own unread messages upon user login?

2) Chunking?

HDVinnie:
closed in https://github.com/HDInnovations/UNIT3D/pull/452

LayerCorp:
I was about to open a new issue for enhancement about this, because we need to send MASS PM but with option to select usersgroups who will get the PM, but found this.

Doing this you can somehow fix this issue because you can send in parts, I really would love to have the option to select what usergroups are receiving the staff mass PM.

PR #451, opened 2018-11-13
(Update) MassPM Function

VerioPL:
Please make some review is it okay. Mass PM's are now handled as a queue job.

TODO: Configuration of queue settings

### DIFF
--- a/app/Http/Controllers/ReportController.php
+++ b/app/Http/Controllers/ReportController.php
@@ -59,7 +59,7 @@ class ReportController extends Controller
     {
         $torrent = $this->torrent->find($request->get('torrent_id'));
         $reported_by = auth()->user();
-        $reported_user = $torrent->user_id;
+        $reported_user = $torrent->user;
 
         $this->report->create([
             'type' => $request->get('type'),


VerioPL:
If accepted please close assigned ticket #238 .

VerioPL:
On my test lab (100k users) with redis queue driver now it is working much more better. In a few (less than 5) seconds foreach in controller is makings jobs, then jobs are processing in background.

HDVinnie:
Why are making another jobs migration? It already exists.
https://github.com/HDInnovations/UNIT3D/blob/master/database/migrations/2017_12_10_020753_create_jobs_table.php

HDVinnie:
If you would please spit this into 4 separate PR's. There to much going on in a single PR that are not related to each other. I suggest making the following separate ones.

```
Title: (Update) MassPM Function
Description: Change MassPM Function to use a queue job (fix for #238)
Tags: Enhancement Tag
```

```
(Update) Torrent Blades Lang 
Description: Torrent blades multilingual support EN/PL
Tags: Translation Tag
```

```
Title: (Fix) 404 in url to TMDB
Description: tmdb/movie/id not tmdb/movies/id
```

```
Title: (Fix) Report Controller
Description: revert variable change from commit 
```


One pull request per feature/fix/update/etc - If you want to do more than one thing in separate "systems", send multiple pull requests.

Issue #402, opened 2018-09-30
Report request produces error


yourguessismine:
Pressing "Report request" on request page opens modal, clicking "Report request" within the modal produces 500 error (see below).

`Trying to get property 'user' of non-object {"userId":3,"email":"email@domain.tld","exception":"[object] (ErrorException(code: 0): Trying to get property 'user' of non-object at /var/www/html/app/Http/Controllers/ReportController.php:49)`

### Issue Type
- [x] Bug Report

### End User Info
-- UNIT3D Version: 1.8.6

-- Laravel Version: 5.6.37

-- PHP Version: 7.2.9

-- Database Driver & Version: MySQL 5.7.23

-- Web Server Driver & Version: nginx 1.15.0

-- OS Driver and Version: Ubuntu 18.04.1

### Expected Behavior: 

Submit report and toater message displayed.

### Actual Behavior: 

500 error

### Steps To Reproduce The Problem:
1. Open a request
2. Click Report request at the top of the page
3. Click Report request within modal.

### Possible Fix:
Assign alternate value to `$reported_user` variable within the ReportController so if `$torrent->user` is not available it will use `torrentRequest->user`. I believe a new Requests class would need to be created.

VerioPL:
Needed refactor postReport function
- input() directive
- validation
- rebuild syntax

Assigned to me

PR #488, opened 2019-01-10
(Fix) Report Request/Report Torrent

singularity43:
Updated the report system to prepare for a rewrite.

Established relationship to requests.

Modeled the controller using UNIT3D structure.

Tested.

Closes: #402


HDVinnie:
PR looks good.However maybe before merging it should also contain the fix for reporting users (from button in profiles) which is also broken. 

@singularity43 thoughts?

singularity43:
@HDVinnie Oops, did not realize that it was broken. Was only focused on the Issue # I was trying to close.
Will fix that now.

PR #490, opened 2019-01-12
(Add) Ability to pull torrents via RSS

singularity43:
Adds user facing portion, admin portion and some adjustments to design elements so users can discover the feature.

TODO: Regenerate RID (rss key) - Will come with "refactoring" of My Account vs Security.

Cheers.


app/Http/Controllers/Staff/RssController.php
```diff
@@ -0,0 +1,227 @@
...
+    public function __construct(TorrentFacetedRepository $torrent_faceted, Toastr $toastr)
+    {
+        $this->torrent_faceted = $torrent_faceted;
+        $this->toastr = $toastr;
+    }
+
+    /**
+     * @param Toastr $toastr
+     */
+    public function setToastr($toastr)
```
HDVinnie: what is point of this?

singularity43: Probably a hangover from when I was still thinking of how to approach this.

Nuke.

HDVinnie: Cool I am going to merge both PRs. Then run the code through StyleCI since it doesn't follow UNIT3D code flow. Then do alittle manual cleanup.

singularity43: Sounds good. Let me know what I can do to fix StyleCl errors in future.

About manual cleanups, please discord me any major issues. I left some notes in what I knew was wrong.  Will keep in mind for future.

HDVinnie: Will do. Its nothing major. Just want to keep things standardized with what UNIT3D already does. Like `$rss = $user->rss()->where('is_private',0)->findOrFail($id);` to `$rss = $user->rss()->where('is_private', '=', 0)->findOrFail($id);`. I know there the same but we decided to show `'='` for clarity to noobs. 

`$this->toastr->error($error, 'Uhoh!', ['options']));` to `$this->toastr->error($error, 'Whoops!', ['options']));`

`$this->toastr->success($success, 'Congratulations!', ['options']));` to `$this->toastr->success($success, 'Yay!', ['options']));`

Just because whoops and yay is what we use everywhere already.

singularity43: No problem.  Please look at my validation though. I like to validate the long and dirty way... because I'm paranoid.  Laravel validation seems too easy.... 

About lang issues... we need to export to lang files in long run.....

(We discussed the json error thing at one point).

Issue #396, opened 2018-09-20
Torrent search enhancement 


yourguessismine:
### Issue Type
- [x] Enhancement

Presently, if you apply filters on torrents search page, click on a torrent then go back to the search page the filters are lost. It would be nice if the most recent filters were cached so they remain when going back and fourth between search results and torrent pages.

The current lazy workaround I have implemented is making every link in the search results open in a new tab so the filters in the original tab remain intact.

Also, it would be nice if somewhere in the search filter area there was an indicator showing the number of results available with current filters applied.

HDVinnie:
I goto Torrent List page. Check the following filters. `Movies`, `1080p` and `Alive`. Once results load I goto say `page 7`. Click torrent. View Torrent details page. Then click back button in browser I return to previous page with my same filters on page 7. Im thinking this might be a browser thing though. As it works fine on safari but not others. Ill have to look into this more.

yourguessismine:
Ok, thank you.

catfatwm:
search doesnt even work for me on chromium fork, vivaldi. I have to click advanced search to get the searches to appear.

HDVinnie:
There is no "advanced search". Not sure what your talking about...

HDVinnie:
@catfatwm Unrelated to UNIT3D. UNIT3D no longer has quick search. The site you speak of is a modified version.

catfatwm:
oh, i apologize.

HDVinnie:
Closed In PR https://github.com/HDInnovations/UNIT3D/pull/501

PR #501, opened 2019-01-17
(Update/Add) Torrent Views Search

singularity43:
- Add search to cards view.
- Add on page grouping (similar to other tracker pages), rather than click through.
- Add ability to hide filters ( as discussed with @HDVinnie )
- Add search to groupings view.
- Add Helper JS file for future helper type functions.
- Remove unused views.


HDVinnie:
Closes #396

PR #510, opened 2019-01-19
(Add) Filters On User History View

singularity43:
Respects @HDVinnie desire to not alter real history.

Just applies proposed compromise of filters.

JS visibility based. Not ajax.


singularity43:
Ignore earlier comments.

Had to switch to ajax to support users request of multiple filters being applied at once.

Sorry about confusion.

Cheers.

PR #590, opened 2019-02-22
(Add) User Request Sharing + Stats + Privacy Fixes

singularity43:
1: Users can now view other users requests (if not anonymous and if user doesnt hide privacy).
2: Admins/staff can see requests.
3: New stat on profile: requested and requests filled.
4: Fix missing hide from peer block (by adding new group targeted setting).
5: Add hide from online block (by adding new group targeted setting).

Close out bug + user request.


app/Http/Controllers/UserController.php
```diff
@@ -112,6 +113,9 @@ public function profile($slug, $id)
         $realdownload = $user->downloaded + $bondownload;
         $invitedBy = Invite::where('accepted_by', '=', $user->id)->first();
 
+        $requested = TorrentRequest::where('user_id','=',$user->id)->count();
+        $filled = TorrentRequest::where('filled_by','=',$user->id)->where('approved_by','is not','null')->count();
```
HDVinnie: This query is not valid. Should be:
```
$filled = TorrentRequest::where('filled_by','=',$user->id)->whereNotNull('approved_by')->count();
```

singularity43: It's dirty, but it's not not valid.

Fixing.

PR #675, opened 2019-04-14
Big update on lang Chinese

pbodq2:
1. I just spent few days on enhancing two Chinese lang folders : "cn" and "tw"
In the coming few weeks, i will go on updating them.
All these files are align to the current English version.

2. Just one little label amendment on the lang/en/user.php


HDVinnie:
Great PR @pbodq2 . Thank you.

PR #692, opened 2019-04-24
(fix) 2 bugs on Group View pagination display

pbodq2:
Should only affect Group View.


HDVinnie:
Thank you for the PR. Nice job finding this bug and providing this fix.

PR #694, opened 2019-04-27
(fix) renders missing array elements in search result

pbodq2:
Occurs when Null/0 $page value is posted back and is used for array index. Hence, not compatible with "-1" commit # 903c148391ddba75393bfb8ef0133bc871ee552d


HDVinnie:
@pbodq2 your PR has a conflict. Seems you didnt sync to master first.

pbodq2:
ok, I resync the officials to my repository first. Then, I resubmit the change.

Issue #699, opened 2019-04-28
Download without speed when using ipv6 setting


buaapyj:
Issue Type
- [X] Enhancement

End User Info
-- UNIT3D Version: Currently Running UNIT3D Nex-Gen Torrent Tracker v1.9.2

-- Laravel Version: Laravel Framework 5.8.7

-- PHP Version:

-- Database Driver & Version: mysql 5.7

-- Web Server Driver & Version: nginx，nginx version: nginx/1.15.8

-- OS Driver and Version: ubuntu 16.04

Expected Behavior:
First, I make a torret and issue the torrent from the UNIT3D website,  then I download a torrent from another pc.
If I use the ipv6 address as the tracker ip,download task without speed.
when I change the ip address as the v4. It works.
Ensure my pc is well, I added V6 address as a user manually in utorrent, then the task have speed.

Actual Behavior:
Trakcer seems to recognize the ipv6 address, but I don't know why it was not sent to the client.
![image](https://user-images.githubusercontent.com/25047090/56867954-ac12ff00-6a1e-11e9-85d1-7db81f51f710.png)

Steps To Reproduce The Problem:

First, I wonder to know if the tracker support IPV6?
If need some setting, plz teach me.

HDVinnie:
@buaapyj thanks for the report. I am going to do some testing and will get back to you.

buaapyj:
> @buaapyj thanks for the report. I am going to do some testing and will get back to you.
If you have anything that I need to do, you can let me do it.

HDVinnie:
No nothing your end. Our announce needs to be adjusted to IPV6 support.

buaapyj:
> No nothing your end. Our announce needs to be adjusted to IPV6 support.

I mean if you have any v6 test need to do

masterfo78:
unfortunately has to turn on again. We have users who have ipv6 and the credited no upload.

I have also fixed with me no ipv6 address in the peers displayed

buaapyj:
> 不幸的是必须再次开启。我们的用户拥有ipv6并且没有上传。
> 
> 我也修复了我在同行显示的没有ipv6地址

can you provide the screenshot result by click "View Peers" in torrent detail page.

masterfo78:
https://i.gyazo.com/bf5d4f354fc582e07579d4287a2eee06.png

buaapyj:
> https://i.gyazo.com/bf5d4f354fc582e07579d4287a2eee06.png

[...]

PR #714, opened 2019-05-03
Update AnnounceController.php

buaapyj:
add Ipv6 support for Tracker based on IPv6 Tracker Extension(BEP 7 on BitTorrent.org) #699


HDVinnie:
closes #699

PR #780, opened 2019-05-22
generated model phpdocs

mxss:
I've generated model phpdocs with https://github.com/barryvdh/laravel-ide-helper, this should help with IntelliSense in IDE's

### DIFF
--- a/app/Models/Album.php
+++ b/app/Models/Album.php
@@ -16,8 +16,6 @@ namespace App\Models;
 use Illuminate\Database\Eloquent\Model;
 
 /**
- * App\Models\Album
- *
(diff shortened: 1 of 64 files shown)


HDVinnie:
@mxss Thanks. I also use the IDE helper for phpstorm however never really liked the docs in the models or controllers for the class. Just have been doing the functions themselves. Just seems like alot of bloat. You think its worth it?

mxss:
@HDVinnie Yes, I think it's worth it, if there's too much docs I can remove `@method` info because `@property` docs are most helpful. That should reduce docs size in ~half

Issue #472, opened 2018-12-21
Cruddy Controllers


HDVinnie:
### Issue Type
- [x] Enhancement

Essentially, we want to limit our controllers to the 7 resourceful actions. Anything else is an opportunity to create additional controllers.

Example:

Verb | URI | Action | Route Name
-- | -- | -- | --
GET | /photos | index | photos.index
GET | /photos/create | create | photos.create
POST | /photos | store | photos.store
GET | /photos/{photo} | show | photos.show
GET | /photos/{photo}/edit | edit | photos.edit
PUT/PATCH | /photos/{photo} | update | photos.update
DELETE | /photos/{photo} | destroy | photos.destroy

singularity43:
@HDVinnie You should go a step further and require all future controllers with applicable characters to use resources.

HDVinnie:
To be completed for UNIT3D Community Edition v1.9.5
- ref https://github.com/HDInnovations/UNIT3D-Community-Edition/tree/Cruddy-By-Design

HDVinnie:
Closed in https://github.com/HDInnovations/UNIT3D-Community-Edition/releases/tag/v2.0.0

PR #788, opened 2019-05-23
moved some functions from TorrentController into TorrentCRUDController

mxss:
#472


mxss:
@HDVinnie any updates?

JackPriceBurns:
Looking at the base issue, I believe the aim is to use the `Route::resource(...` helper
https://laravel.com/docs/5.8/controllers#restful-partial-resource-routes

The `torrents` and `torrent` action should be index and show respectively.

@HDVinnie I am not missing the mark with this?

HDVinnie:
@JackPriceBurns @mxss I personally am not a fan of using Route Resources. I prefer to have each route defined separately as I feel it is easier for newer devs or end users to understand. 

I did not approve this PR yet because I wasnt sure it was done. In reality all controllers should have only  have 7 resourceful actions. Index, Create, Store, Show, Edit, Update and destroy. Along with respectful route names and schema. `Route::get('/torrents/{slug}.{id}{hash?}', 'TorrentController@show')->name('torrents.show');` not `Route::get('/torrents/{slug}.{id}{hash?}', 'TorrentController@torrent')->name('torrent');`.

All the different view for instance in the TorrentController should be split into there own resourcful controller. TorrentListController, TorrentGroupingController, TorrentCardController and so forth. Atleast thats my thinking. What you guy think?

JackPriceBurns:
@HDVinnie yeah that makes sense. If all of the views can work using the same data, then I don't see any benefit in splitting the views out into separate controllers, I feel like you'd then just duplicate the code in the controllers.

You can check the user settings as for which style to use?
Or you could check if there is a get parameter set `?view=card` and then show the correct view then?

PR #804, opened 2019-06-09
Update top_nav.blade.php

buaapyj:
change for support chinese id(when use slug the result of slug is null)
just fix the top_nav.blade.php 
there still have some other pages, I will fix them later.


HDVinnie:
ill leave this open until you change it in all places.....rather review and do it all at once.

PR #809, opened 2019-06-15
change for support chinese id

buaapyj:
change for support chinese id(when use slug the result of slug is null)

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -248,7 +248,7 @@ class UserController extends Controller
         // Activity Log
         \LogActivity::addToLog("Member {$user->username} has updated there profile.");
 
-        return redirect()->route('user_edit_profile_form', ['username' => $user->slug, 'id' => $user->id])
+        return redirect()->route('user_edit_profile_form', ['username' => $user->username, 'id' => $user->id])
             ->withSuccess('Your Account Was Updated Successfully!');
     }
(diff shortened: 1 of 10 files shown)


buaapyj:
@HDVinnie  any else change need to do?

HDVinnie:
@buaapyj in my opinion to make this proper. Since we are no longer using the true slug parameter and instead direct username we should adjust the parameter no?

`return redirect()->route('user_settings', ['slug' => $user->username, 'id' => $user->id])`

to 

`return redirect()->route('user_settings', ['username' => $user->username, 'id' => $user->id])`

Of course will have to change it as well in routes/views/controllers. If you want I can do it this week.

buaapyj:
> @buaapyj in my opinion to make this proper. Since we are no longer using the true slug parameter and instead direct username we should adjust the parameter no?
> 
> `return redirect()->route('user_settings', ['slug' => $user->username, 'id' => $user->id])`
> 
> to
> 
> `return redirect()->route('user_settings', ['username' => $user->username, 'id' => $user->id])`
> 
> Of course will have to change it as well in routes/views/controllers. If you want I can do it this week.

actually not only username have this problem , the title of torrents , the title of forum name  etc (all space using slug have this problem[can not be full chinese word]). I don't know change slug to username of id or some uniquely identifies is the right solution ? or any Slug extension to solve this problem？

PR #810, opened 2019-06-15
Fix issue with slow home page

JackPriceBurns:
Loading the home page is extremely slow when you have a lot of torrents.

This is due to a huge amount of N+1 queries.


JackPriceBurns:
Just realised, if you are a normal user, not a mod, then it does a query to check if you have permission to view each topic

PR #811, opened 2019-06-16
Slow torrent list

JackPriceBurns:
The torrents index page is actually slow to load and uses a lot of server resources, when loading the page, normally it runs around 100~ queries to generate the page, and another 100~ queries for every time you use the search bar (with 25 pagination).

If you double the pagination the amount of queries also double!

After this PR the page takes around 20~ queries to load the page. And this amount doesn't change if you increase the pagination.

I decided to leave the pagination the same as I was unsure if you wanted to increase this limit or not?

The same logic used to speed up the home page is used here.

Furthermore, I updated the movie database client to cache the data using any cache driver that laravel supports. Before you were forced to use redis, which I don't have on my local machine. (I can revert this if needed)

### DIFF
--- a/app/Http/Controllers/TorrentController.php
+++ b/app/Http/Controllers/TorrentController.php
@@ -78,7 +78,7 @@ class TorrentController extends Controller
         $user = $request->user();
         $repository = $this->faceted;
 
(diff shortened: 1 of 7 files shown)


HDVinnie:
Thanks for the contributions @JackPriceBurns

PR #814, opened 2019-06-17
Add torrent seeder

JackPriceBurns:
Not sure if you're interested in this, but I made this seeder for testing the torrents page. It just feels a shame to just delete it as it was quite helpful.

Maybe someone else might find this helpful in the future?


HDVinnie:
Delete it? This already exists using `php artisan demo:seed`. There is a command and factory that will seed a demo with fake users and torrents.

JackPriceBurns:
Oh oops! Didn't see that 😅

PR #892, opened 2019-11-06
Christ84 traduc fr

christ84:
files lang fr updated


HDVinnie:
This PR need to be resubmitted with changed files only and not the dupes in root directory.

PR #992, opened 2019-11-19
fix faq url

HVRVKVT:
should fix dead link in sitebar


HDVinnie:
This is wrong. The proper ID is 2 which matches the seeder.

https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/19734aa60a95f1c6db3e7adc0f0fdfc178599607/database/seeds/PagesTableSeeder.php#L33

PR #1014, opened 2019-12-13
Fix buffer figure

innocente205:
Buffer was being calculated solely from the upload figure divided by the config ratio. It should also take into account the amount already downloaded. Hence (uploaded - downloaded) / ratio.


innocente205:
Should have been (uploaded/ratio) - downloaded

PR #1029, opened 2019-12-19
(Update) Add delete buttons in outbox, return to outbox

innocente205:
Mainly copied and pasted from other parts of the existing code.

I would have liked to apply the same 'return to calling page' logic for the message view pages viewed from the outbox, but I couldn't find an elegant way of passing the destination to the controller without using a form and a hidden field.

I also added an 'if' clause in the Reply To A Message function to avoid sending replies to yourself from the outbox.


HDVinnie:
Thanks for the contribution @innocente205

PR #1037, opened 2019-12-23
(Update) "Slightly" Opinionated Code Refactor

HDVinnie:
- Encapsed Strings To Sprintf
- Return Type Declaration
- Add Literal Separator To Numbers
- PHPDocs
- Shorten Else If Statements
- Closure To Arrow Function
- Simplify Conditions
- Final clsses
- Count On Null
- Stringify Str Needles
- Json Throw On Error
- Regex Dash Escape
- Helper Function To Constructor Injection Refactor
+ More

While some changes in this PR may seem pointless and opinionated the point is to be php7.4 ready and keep a single code style standard.


app/Achievements/UserFilled100Requests.php
```diff
@@ -15,20 +15,25 @@
 
 use Gstt\Achievements\Achievement;
 
-class UserFilled100Requests extends Achievement
+final class UserFilled100Requests extends Achievement
```
poppabear8883: I typically do not use `final classes`, but they are considered better practice. 

I personally like leaving the classes a bit more open, because in the real world i think that `final classes` could cause more headaches then what its worth to have them ... 

just my opinion

HDVinnie:
Just to be clear this will be a major UNIT3D-Community-Edition release if agreed on and will lock UNIT3D-Community-Edition to PHP7.4+ only due to making use of:

- Typed properties
- Arrow functions
- Array spread operator
- Numeric Literal Separator
- Preloading and more

This will make the road to PHP8 smoother.

app/Notifications/UsernameReminder.php
```diff
@@ -13,21 +13,39 @@
...
+     */
+    private $urlGenerator;
 
     /**
      * Create a new notification instance.
      *
      * UsernameReminderEmail constructor.
      */
-    public function __construct()
+    public function __construct(Translator $translator, Repository $configRepository, UrlGenerator $urlGenerator
```
GrahamCampbell: Missing closing bracket and opening brace.

HDVinnie: Fixed. Thank you.

app/Achievements/UserFilled100Requests.php
GrahamCampbell: Final classes make sense for things like data classes and models. I'd keep them in, since there's no risk anyone would want to extend them, even in anger. ;)

PR #1048, opened 2020-01-06
Update bon transactions value

innocente205:
When purchases were made from the store, the cost was being recorded as the 'value' from the bonexchange table instead of the cost

Name, Cost
'100 GB Upload', '107374182400.00'
'25 GB Upload', '26843545600.00'
'100 GB Upload', '107374182400.00'
'100 GB Upload', '107374182400.00'


HDVinnie:
Good catch. Thanks for the PR. Guessing you caught this looking at the audit logs?

innocente205:
Thanks. Actually looking at the database.

innocente205:
Please note: this change breaks the reporting of 'upload added from BON store' in the users profile page.

PR #1102, opened 2020-01-31
revert top nav to static

HVRVKVT:
as discussed a while ago, this is probably the cleanest method to
make use of the anchors on pages/wiki


HDVinnie:
Tested and is causing issues with navigation

PR #1106, opened 2020-02-01
Adds magnet support

PyR8zdl:
Adds magnet links to torrent page and search results.

Enable in config/torrent.php


resources/views/torrent/results.blade.php
```diff
@@ -143,13 +143,29 @@ class="torrent-poster-img-small show-poster" alt="@lang('torrent.poster')">
                                     <i class="{{ config('other.font-awesome') }} fa-download"></i>
                                 </button>
                             </a>
+                            @if (config('torrent.magnet') == 1)
```
HDVinnie: Is this supposed to be here twice?

resources/views/torrent/results.blade.php
```diff
@@ -143,13 +143,29 @@ class="torrent-poster-img-small show-poster" alt="@lang('torrent.poster')">
...
+                            </a>
+                            @endif
                         @else
                             <a href="{{ route('download', ['id' => $torrent->id]) }}">
                                 <button class="btn btn-primary btn-circle" type="button" data-toggle="tooltip"
                                     data-original-title="@lang('common.download')">
                                     <i class="{{ config('other.font-awesome') }} fa-download"></i>
                                 </button>
                             </a>
+                            @if (config('torrent.magnet') == 1)
```
HDVinnie: In relation to comment above.

resources/views/torrent/results.blade.php
PyR8zdl: Yeah this was a silly mistake on my end. I'll clean it up. ;)

masterfo78:
Don't Go no Upload with magnet

HDVinnie:
> Don't Go no Upload with magnet

English?

masterfo78:
When I use magnet it only shows the meta in the seedbox. No download is possible.

torman365:
That is normal, downloads do work and show name will appear once completed

masterfo78:
For me in the client, only a long number ends with .meta. The download does not start. In utorrent I have the name but the file does not start there either.

torman365:
how are you adding the link to your client? you can right click the magnet link,copy  magnet link and then paste it into your clients download box where it says "torrent URL"

masterfo78:
Yes the torrent url is from the tracker

torman365:
yes but how are you adding it to your client?

masterfo78:
per klick

torman365:
try what i said, right click the magnet icon on the torrent page, then "copy link location" and then add that to your client in the "torrent URL" BOX

masterfo78:
This also does not show with just the hash number

PR #1124, opened 2020-02-12
(Refactor) if / else / elseif statements

HDVinnie:
- Change If Else Value Assign To Early Return
- Change Nested Ifs To Early Return
- Remove Always Else


app/Http/Middleware/CheckIfAlreadyVoted.php
```diff
@@ -36,12 +36,9 @@ public function handle($request, Closure $next)
 
         $poll = Option::findOrFail($request->input('option.0'))->poll;
 
-        //if we already have this user's IP stored and linked to the poll they are trying to vote on
-        if ($poll->ip_checking == 1) {
```
cbj4074: Was this conditional check removed intentionally? I'm asking only because I had assumed that all of these changes were programmatic, style-only changes and want to ensure that this isn't some kind of parser bug.

HDVinnie: This was intentional.

PR #1130, opened 2020-02-13
Adjust logic for poll.

gyakkun:
Afaic, one user should only vote on a poll **once**, after which should the remaining validation logic be done.

I extract the complete logic to function `validateVoter()`, though the meaning of "voter" here and in the db could be confused from my perspective as a non-native speaker. 621f7ba437cdca8c9db1e2ecb5cdd2a40743b5d7

Futher adjust some variable names in view and controller. 7856d27dbfa5df12676adae52a0d42a18205cae0

### DIFF
--- a/app/Http/Controllers/PollController.php
+++ b/app/Http/Controllers/PollController.php
@@ -90,8 +90,8 @@ class PollController extends Controller
         }
 
         // Operate options after validation
-        foreach ($request->input('option') as $option) {
-            Option::findOrFail($option)->increment('votes');
+        foreach ($request->input('option-id') as $id) {
+            Option::findOrFail($id)->increment('votes');
         }
 
         // Make voter after option operation completed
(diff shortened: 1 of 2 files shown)


gyakkun:
The semantic adjustment 7856d27dbfa5df12676adae52a0d42a18205cae0 should not be performed, my mistake. It crashes the request from blocks/poll view.

From my test results, one can still vote on the same poll in blocks/poll after voting, that's why I update the logic in controller 621f7ba437cdca8c9db1e2ecb5cdd2a40743b5d7, though it should not display in blocks of index page after voting (Sorry, I just noticed):
https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/fffa67ce32176f207ab49c10151b1565dbf9d6a5/resources/views/blocks/poll.blade.php#L1
I thought the reason for the test result is that I perform my test keeping the unvoted index page in my browser unclosed and make the request twice.

PR #1132, opened 2020-02-13
Update message.blade.php

ValkiryXDA:
Changing location of email header to match with title given in other. Simplify location. didnt see a point in having name in app.php and title in other.php to not tying to each other for emails. one location makes more sense? thanks

### DIFF
--- a/resources/views/vendor/mail/text/message.blade.php
+++ b/resources/views/vendor/mail/text/message.blade.php
@@ -21,7 +21,7 @@
     {{-- Footer --}}
     @slot('footer')
         @component('mail::footer')
-            © {{ date('Y') }} {{ config('app.name') }}. All rights reserved.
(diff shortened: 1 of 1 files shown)


Golumpa:
You're gonna want to do the html one as well

/resources/views/vendor/mail/html/message.blade.php

Golumpa:
Oh and the footer link

resources/views/vendor/mail/text/message.blade.php
```diff
@@ -1,8 +1,8 @@
 @component('mail::layout')
     {{-- Header --}}
     @slot('header')
-        @component('mail::header', ['url' => config('app.url')])
-            {{ config('app.name') }}
+        @component('mail::header', ['url' => config('other.url')])
```
HDVinnie: This is wrong. There is no `other.url`. It should remain `app.url`.

PR #1133, opened 2020-02-13
Update message.blade.php

ValkiryXDA:
Simplify Location to previous commit in HTML f5ccd5f919336abd82932e3a916885545785c89a

### DIFF
--- a/resources/views/vendor/mail/html/message.blade.php
+++ b/resources/views/vendor/mail/html/message.blade.php
@@ -1,8 +1,8 @@
 @component('mail::layout')
     {{-- Header --}}
     @slot('header')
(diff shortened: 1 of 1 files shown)


resources/views/vendor/mail/html/message.blade.php
```diff
@@ -1,8 +1,8 @@
 @component('mail::layout')
     {{-- Header --}}
     @slot('header')
-        @component('mail::header', ['url' => config('app.url')])
-            {{ config('app.name') }}
+        @component('mail::header', ['url' => config('other.url')])
```
HDVinnie: This is wrong. There is no `other.url`. It should remain `app.url`.

PR #1141, opened 2020-02-13
fix: Too few arguments to function Controllers\ReportController::user().

gyakkun:
Debug info:

```
Symfony\Component\Debug\Exception\FatalThrowableError
Too few arguments to function App\Http\Controllers\ReportController::user(), 2 passed and exactly 3 expected
```

Reproduce: User profile page -> Report -> Report.


HDVinnie:
Good find. Thank you!

PR #1149, opened 2020-02-14
(Tests) Update Model Factories

HDVinnie:


cbj4074:
This is cool; did you use a generator of some kind to create all of these?

I ask primarily because I'm curious if the Faker data types are based on the column types in the DB. If they are, then this probably needs quite a bit less manual review (which is good!).

HDVinnie:
> This is cool; did you use a generator of some kind to create all of these?
> 
> I ask primarily because I'm curious if the Faker data types are based on the column types in the DB. If they are, then this probably needs quite a bit less manual review (which is good!).

Yes I used this package https://github.com/mpociot/laravel-test-factory-helper

database/factories/FreeleechTokenFactory.php
```diff
@@ -6,5 +6,7 @@
 
 $factory->define(App\Models\FreeleechToken::class, function (Faker $faker) {
     return [
+        'user_id'    => $faker->randomNumber(),
+        'torrent_id' => $faker->randomNumber(),
```
cbj4074: It would be ideal to enforce foreign key constraints here. I'm curious why the automated process used `randomNumber()` here. I'll check later, but for now, I'm noting that these should really honor FKs constraints. I would expect something more like what we see elsewhere in similar usages:

```php
        'user_id' => function () {
            return factory(App\Models\User::class)->create()->id;
        },
        'torrent_id' => function () {
            return factory(App\Models\Torrent::class)->create()->id;
        },
```

database/factories/GitUpdateFactory.php
```diff
@@ -6,5 +6,7 @@
 
 $factory->define(App\Models\GitUpdate::class, function (Faker $faker) {
     return [
+        'name' => $faker->name,
+        'hash' => $faker->word,
```
cbj4074: This should probably be wrapped in an actual hash function, even if the input is a random word.

database/factories/LikeFactory.php
```diff
@@ -1 +1,19 @@
...
+
+$factory->define(App\Models\Like::class, function (Faker $faker) {
+    return [
+        'user_id' => function () {
+            return factory(App\Models\User::class)->create()->id;
+        },
+        'post_id' => function () {
+            return factory(App\Models\Post::class)->create()->id;
+        },
+        'subtitle_id' => $faker->randomNumber(),
```
cbj4074: FK?

database/factories/NotificationFactory.php
```diff
@@ -1 +1,15 @@
 <?php
+
+/* @var $factory \Illuminate\Database\Eloquent\Factory */
+
+use Faker\Generator as Faker;
+
+$factory->define(App\Models\Notification::class, function (Faker $faker) {
+    return [
+        'type'            => $faker->word,
+        'notifiable_id'   => $faker->randomNumber(),
```
cbj4074: FK?

database/factories/PeerFactory.php
```diff
@@ -1 +1,29 @@
 <?php
+
+/* @var $factory \Illuminate\Database\Eloquent\Factory */
+
+use Faker\Generator as Faker;
+
+$factory->define(App\Models\Peer::class, function (Faker $faker) {
+    return [
+        'peer_id'     => $faker->word,
```
cbj4074: FK?

database/factories/PersonalFreeleechFactory.php
```diff
@@ -1 +1,11 @@
 <?php
+
+/* @var $factory \Illuminate\Database\Eloquent\Factory */
+
+use Faker\Generator as Faker;
+
+$factory->define(App\Models\PersonalFreeleech::class, function (Faker $faker) {
+    return [
+        'user_id' => $faker->randomNumber(),
```
cbj4074: Shouldn't this reference a FK?

database/factories/PrivateMessageFactory.php
```diff
@@ -1 +1,20 @@
...
+        'sender_id' => function () {
+            return factory(App\Models\User::class)->create()->id;
+        },
+        'receiver_id' => function () {
+            return factory(App\Models\User::class)->create()->id;
+        },
+        'subject'    => $faker->word,
+        'message'    => $faker->text,
+        'read'       => $faker->boolean,
+        'related_to' => $faker->randomNumber(),
```
cbj4074: FK?

database/factories/SeedboxFactory.php
```diff
@@ -1 +1,15 @@
...
+
+use Faker\Generator as Faker;
+
+$factory->define(App\Models\Seedbox::class, function (Faker $faker) {
+    return [
+        'user_id' => function () {
+            return factory(App\Models\User::class)->create()->id;
+        },
+        'name' => $faker->name,
+        'ip'   => $faker->word,
```
cbj4074: Isn't there a proper Faker IP address type?

database/factories/TagTorrentFactory.php
```diff
@@ -6,7 +6,8 @@
 
 $factory->define(App\Models\TagTorrent::class, function (Faker $faker) {
     return [
-        'tag_name' => function () {
+        'torrent_id' => $faker->randomNumber(),
```
cbj4074: FK?

database/factories/TorrentFactory.php
```diff
@@ -0,0 +1,59 @@
...
+use Faker\Generator as Faker;
+
+$factory->define(App\Models\Torrent::class, function (Faker $faker) {
+    return [
+        'name'            => $faker->name,
+        'slug'            => $faker->slug,
+        'description'     => $faker->text,
+        'mediainfo'       => $faker->text,
+        'bdinfo'          => $faker->text,
+        'info_hash'       => $faker->word,
```
cbj4074: Use proper hash here?

database/factories/TorrentFactory.php
```diff
@@ -0,0 +1,59 @@
...
+        'mal'          => $faker->word,
+        'igdb'         => $faker->word,
+        'type'         => $faker->word,
+        'resolution'   => $faker->randomNumber(),
+        'stream'       => $faker->boolean,
+        'free'         => $faker->boolean,
+        'doubleup'     => $faker->boolean,
+        'highspeed'    => $faker->boolean,
+        'featured'     => $faker->boolean,
+        'status'       => $faker->randomNumber(),
```
cbj4074: FK?

database/factories/TorrentRequestBountyFactory.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Faker\Generator as Faker;
+
+$factory->define(App\Models\TorrentRequestBounty::class, function (Faker $faker) {
+    return [
+        'user_id' => function () {
+            return factory(App\Models\User::class)->create()->id;
+        },
+        'seedbonus'   => $faker->randomFloat(),
+        'requests_id' => $faker->randomNumber(),
```
cbj4074: FK?

database/factories/TorrentRequestClaimFactory.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+/* @var $factory \Illuminate\Database\Eloquent\Factory */
+
+use Faker\Generator as Faker;
+
+$factory->define(App\Models\TorrentRequestClaim::class, function (Faker $faker) {
+    return [
+        'request_id' => $faker->randomNumber(),
```
cbj4074: FK?

database/factories/VoterFactory.php
```diff
@@ -0,0 +1,17 @@
...
+
+$factory->define(App\Models\Voter::class, function (Faker $faker) {
+    return [
+        'poll_id' => function () {
+            return factory(App\Models\Poll::class)->create()->id;
+        },
+        'user_id' => function () {
+            return factory(App\Models\User::class)->create()->id;
+        },
+        'ip_address' => $faker->word,
```
cbj4074: Use proper Faker IP method?

HDVinnie:
> On the whole, this looks to be a welcome addition.
> 
> Without being intimately familiar with every database table and attendant column, it's difficult to vet every factory carefully, but we can fine-tune as we flesh-out the test suite.
> 
> I do notice that there are quite a few columns that look like they should refer to proper foreign keys, but do not, so I think it's worth checking the actual tables/migrations to ensure that the keys are present where appropriate.
> 
> Ultimately, I recommend reviewing each comment and either fixing or resolving as appropriate.

I figured we would merge this as a base and then fine tune as we go with tests?

PR #1158, opened 2020-02-17
(Refactor) Encapsed Strings To Sprintf

HDVinnie:
### `Encapsed Strings To Sprintf`

Convert enscaped {$string} to more readable and preferred sprintf logic

```diff
 final class SomeClass
 {
     public function run(string $format)
     {
-        return "Unsupported format {$format}";
+        return sprintf('Unsupported format %s', $format);
     }
 }
```

### DIFF
--- a/app/Console/ConsoleTools.php
+++ b/app/Console/ConsoleTools.php
@@ -27,8 +27,7 @@ trait ConsoleTools
 
     private function cyan($line)
     {
-        $this->io->writeln(sprintf('
-<fg=cyan>%s</>', $line));
+        $this->io->writeln(sprintf('<fg=cyan>%s</>', $line));
     }
 
     private function white($line)
@@ -38,26 +37,22 @@ trait ConsoleTools
 
     private function magenta($line)
     {
-        $this->io->writeln(sprintf('
-<fg=magenta>%s</>', $line));
+        $this->io->writeln(sprintf('<fg=magenta>%s</>', $line));
     }
 
     private function green($line)
     {
-        $this->io->writeln(sprintf('
-<fg=green>%s</>', $line));
+        $this->io->writeln(sprintf('<fg=green>%s</>', $line));
     }
 
     private function red($line)
     {
-        $this->io->writeln(sprintf('
-<fg=red>%s</>', $line));
+        $this->io->writeln(sprintf('<fg=red>%s</>', $line));
     }
 
     private function blue($line)
     {
-        $this->io->writeln(sprintf('
-<fg=blue>%s</>', $line));
+        $this->io->writeln(sprintf('<fg=blue>%s</>', $line));
     }
 
     private function done()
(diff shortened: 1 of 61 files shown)


app/Models/Poll.php
```diff
@@ -120,9 +120,9 @@ public function setTitleAttribute($title)
     public function makeSlugFromTitle($title)
     {
         $slug = strlen($title) > 20 ? substr(Str::slug($title), 0, 20) : Str::slug($title);
-        $count = $this->where('slug', 'LIKE', "%$slug%")->count();
+        $count = $this->where('slug', 'LIKE', sprintf('%%s%', $slug))->count();
```
cbj4074: This might be okay, but it looks like the type of thing that we should verify. I'd be a bit surprised if additional escaping isn't necessary.

app/Notifications/NewBon.php
```diff
@@ -68,7 +68,7 @@ public function toArray($notifiable)
         return [
             'title' => $this->sender.' Has Gifted You '.$this->transaction->cost.' BON',
             'body'  => $this->sender.' has gifted you '.$this->transaction->cost.' BON with the following note: '.$this->transaction->comment,
-            'url'   => "/users/{$this->transaction->senderObj->username}",
+            'url'   => sprintf('/users/%s', $this->transaction->senderObj->username),
```
cbj4074: This may not be the place to address this, these URLs should be generated using calls to `route()` or similar. Many other Notification classes have these hard-coded URL segments. It might be worth creating an Issue or PR to fix this before we forget.

app/Models/Poll.php
cbj4074: Here's an example of why I say this, in Tinker:

```php
>>> $slug = 'foo';
=> "foo"
>>> sprintf('%%s%', $slug);
=> "%s"
```

And without the additonal `%s` instances:

```php
>>> sprintf('%s', $slug);
=> "foo"
```

app/Notifications/NewBon.php
HDVinnie: agree

app/Console/ConsoleTools.php
```diff
@@ -27,32 +27,37 @@ trait ConsoleTools
 
     private function cyan($line)
     {
-        $this->io->writeln("\n<fg=cyan>$line</>");
+        $this->io->writeln(sprintf('
```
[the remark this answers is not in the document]
HDVinnie: Fixed and reported.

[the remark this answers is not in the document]
cbj4074: @HDVinnie It looks like you fixed this. Thanks!

app/Models/Poll.php
HDVinnie: I actually overlooked this and prefer how was from the start. I dont believe sprintf() should be used like this in a where statement. Agree?

app/Notifications/NewBon.php
cbj4074: Sounds good. I notice that many routes do not have names, either, so you'll have to do add those first.

app/Models/Poll.php
cbj4074: Agreed; I think it was fine to begin with.

PR #1168, opened 2020-02-20
Flesh-out ChatController tests

cbj4074:
This PR expands upon previous work to fill-in the test cases that Shift scaffolded programatically.

At this point, all of the `ChatController` tests are passing, although a handful of them should be further improved with assertions against the returned data structures (as I've done for the majority of them).


cbj4074:
Note that all the `sprintf()` calls can be replaced with `route()` calls once the API routes have proper names.

cbj4074:
@HDVinnie For some reason, a couple of tests are failing on Travis-CI while they pass locally. Let's hold-off on merging this for now.

cbj4074:
Despite every effort to discern the source of the failure on Travis-CI, the problem remains elusive. Accordingly, I've modified the failing tests so that they are marked "Incomplete", and therefore skipped, until we're able to determine a root-cause.

Let's go ahead and merge.

PR #1173, opened 2020-02-24
tests: Delete conflicting DemoFactory

cbj4074:
This factory was causing unexpected behavior with the UserFactory;
basically, this factory was taking precedence whenever
factory(User::class) was called.

Oddly, this factory was effective only in the Travis-CI environment and
not anybody's respective local environment. Presumably, some nuance
with the filesystem coupled with Composer's autoloader affects the
order in which the respective factories are registered in different
environments.

In any case, this factory is no longer useful and is hereby deleted,
which fixes a handful of test methods that heretofore had to be marked
as incomplete.


cbj4074:
@HDVinnie Indeed, removing that factory fixed everything in Travis. 👍

PR #1174, opened 2020-02-24
tests: Implement TorrentController API tests

cbj4074:
As part of this effort, a dedicated "torrents" storage location was
added to filesystems.php, which affects the torrent file storage
location. Accordingly, the torrent files in existing installations will
need to be moved (or symlinked) upon upgrade. The primary reason for
this change is mentioned in the Laravel docs: "This convention will
keep your publicly accessible files in one directory that can be easily
shared across deployments when using zero down-time deployment systems
like Envoyer."


database/factories/TorrentFactory.php
```diff
@@ -26,34 +26,34 @@
...
-        'mal'          => $faker->word,
-        'igdb'         => $faker->word,
+        'imdb'         => $faker->randomNumber(),
+        'tvdb'         => $faker->randomNumber(),
+        'tmdb'         => $faker->randomNumber(),
+        'mal'          => $faker->randomNumber(),
+        'igdb'         => $faker->randomNumber(),
         'type'         => $faker->word,
-        'resolution'   => $faker->randomNumber(),
+        //'resolution'   => $faker->randomNumber(),
```
HDVinnie: Im confused where it got these....the ones you commented out that is. I don't ever remember pushing those anywhere. Strange.

cbj4074: I agree; I have no idea from where Shift initially obtained these keys. The words `resolution` and `bdinfo` appear nowhere in the database schema nor any migrations. Very odd.

Should we just delete them?

cbj4074: Looking at the commit history, you first committed those; I assumed you got them from a Shift-generated file set. In any case, let me know if you want me to trash everything that is commented.

HDVinnie: yeah remove them

cbj4074:
Rather than introduce a BC-break by moving the torrent file storage location, I'm going to preserve it and make the BC-breaking change in a 3.x major release branch.

I'll push a commit that fixes that shortly.

app/Http/Controllers/API/TorrentController.php
```diff
@@ -79,15 +81,21 @@ public function store(Request $request, Torrent $torrent)
...
+
+        $fileName = sprintf('%s.torrent', uniqid()); // Generate a unique name
+        Storage::disk('torrents')->put($fileName, Bencode::bencode($decodedTorrent));
 
         // Find the right category
         $category = Category::withCount('torrents')->findOrFail($request->input('category_id'));
 
         // Create the torrent (DB)
-        $torrent = new Torrent();
+        $torrent = app()->make(Torrent::class);
```
HDVinnie: Why the change here?

cbj4074: It was a drive-by refactor. Generally speaking, creating a new instance with the `new` keyword should be avoided wherever possible in favor of resolving the class from the service container. And in fact, this should actually be `resolve(Torrent::class)`, which is short-hand for what's there now.

In this specific instance, the change does not affect functionality, but it makes the code more robust. Even though `Torrent` is "just a simple Eloquent model", and is unlikely to change or ever require a constructor argument, using the container like this adds the ability to hook into the service lifecycle and, for example, interact with a `Torrent` instance that is resolved from the container before it's given to its consumer.

In short, `resolve()` should always be favored over a plain-old `new ...`.

HDVinnie: Got it, thanks for explaining. Sounds like a global refactor for the future.

PR #1175, opened 2020-02-24
(Tests) Staff Controllers

HDVinnie:


HDVinnie:
As discussed these two tests are failing on the update function for Unique constraints even though I have defined uniqueness in the factories.

cbj4074:
Ahh. The issue is that you're *creating* the Chatroom with the factory, and then POSTing to the route, which attempts to create it again.

Just `create()` to `make()`:

```php
$chatroom = factory(Chatroom::class)->make();
```

tests/Feature/Http/Controllers/Staff/ChatRoomControllerTest.php
```diff
@@ -0,0 +1,101 @@
...
+use GroupsTableSeeder;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Tests\TestCase;
+
+/**
+ * @see \App\Http\Controllers\Staff\ChatRoomController
+ */
+class ChatRoomControllerTest extends TestCase
+{
+    use RefreshDatabase;
```
cbj4074: I suggest omitting this trait here because we use it in the parent class, `TestCase`. I say this specifically because if we ever need to disable the trait while debugging tests, it's a lot easier to do it in one place (the parent class) than have to hunt-down all the child classes that might use it.

HDVinnie:
Success
```
PHPUnit 8.5.2 by Sebastian Bergmann and contributors.

...................................................               51 / 51 (100%)

Time: 42.73 seconds, Memory: 60.60 MB

OK (51 tests, 360 assertions)
```

PR #1179, opened 2020-02-26
(Fix) Wrong condition in torrent/groupings.blade.php

howardlau1999:
The condition `$t->meta->title` will throw an Exception if `$t->meta` is non-object


HDVinnie:
Thanks for the contribution @howardlau1999

PR #1183, opened 2020-02-27
tests: Add Announce and Article controller tests

cbj4074:


tests/Feature/Http/Controllers/AnnounceControllerTest.php
```diff
@@ -0,0 +1,48 @@
...
+
+        factory(Torrent::class)->create([
+            'info_hash' => bin2hex(md5('foo')),
+            'status'    => 1, // Approved
+        ]);
+
+        $response = $this->get(route('announce', [
+            'passkey'    => $user->passkey,
+            'info_hash'  => md5('foo'),
+            'peer_id'    => 'bar',
```
HDVinnie: Should be 20-byte string used as a unique ID for the client, generated by the client at startup.

tests/Feature/Http/Controllers/AnnounceControllerTest.php
```diff
@@ -0,0 +1,48 @@
...
+    public function announce_returns_an_ok_response()
+    {
+        $this->seed(GroupsTableSeeder::class);
+
+        $user = factory(User::class)->create([
+            'can_download' => true,
+        ]);
+
+        factory(Torrent::class)->create([
+            'info_hash' => bin2hex(md5('foo')),
```
HDVinnie: Should be a 20-byte SHA1 hash

tests/Feature/Http/Controllers/AnnounceControllerTest.php
```diff
@@ -0,0 +1,48 @@
...
+            'status'    => 1, // Approved
+        ]);
+
+        $response = $this->get(route('announce', [
+            'passkey'    => $user->passkey,
+            'info_hash'  => md5('foo'),
```
HDVinnie: Possibly better to make like 27 from `factory(Torrent::class)->create([.....` to `$torrent = factory(Torrent::class)->create([......` and then reference `'info_hash'  => $torrent->info_hash,` here instead?

cbj4074: Indeed; I've made this change. (The flow is a little different than as suggested because the info hash is `bin2hex()`'d in the database, but not when the client sends it.)

tests/Feature/Http/Controllers/AnnounceControllerTest.php
cbj4074: The actual value is inconsequential unless we assert against it, but it can't hurt to use a realistic value if for no other reason than documenting it (in code) for use in a future test that actually checks the value.

tests/Feature/Http/Controllers/AnnounceControllerTest.php
cbj4074: As with the previous comment, we don't check the requirement in the controller, so the actual value is irrelevant, but we'll add that to the test so we can assert against it in a future test.

PR #1185, opened 2020-02-28
tests: Add BonusController tests

cbj4074:
@HDVinnie Can you take a look at the TODO?

If you run the tests a handful of times (I'd say 10 or so), that test should fail.

Any idea which factory value needs to be specified to fix it? Maybe the Bon value?


cbj4074:
It seems pretty clear that `$this->doItemExchange($user->id, $id)` is returning `false` in the failure cases.

It looks like there are only 3 reasons for which that function would return false, but it'll take me a while to work through the logic whereas I'm hoping you can spot the issue right away, given your knowledge of the business rules.

HDVinnie:
> @HDVinnie Can you take a look at the TODO?
> 
> If you run the tests a handful of times (I'd say 10 or so), that test should fail.
> 
> Any idea which factory value needs to be specified to fix it? Maybe the Bon value?

Will look into it for sure and report back asap.

cbj4074:
That issue should now be resolved.

PR #1188, opened 2020-03-01
(Fix) Replace dots with spaces in upload title

werrpy:
This fixes issues with replacing dots with spaces in upload title. It keeps dots between H.264, and audio channel (ex. FLAC 2.0).

You can test it here: https://regex101.com/r/tz9esE/2/
In the test string add some titles, one per line. In the substitution drop down set it to a single space character.

This replaces dots with spaces that are between:

1. letter and letter
2. number and letter
3. . 4 numbers (.year)
4. 4 numbers and 3 or 4 numbers followed by i or p (year and resolution, 2020.1080p)
5. S . 2 numbers . 3 or 4 numbers followed by i or p (season and resolution, S01.720p)
6. 2 letters . number . number (DD.2.0 or AAC.2.0 or DD+.2.0 or DTS-X.7.1)

Each of the cases is separated by a vertical bar | (aka "OR") in the regex
"Letter" here means any non-digit
Cases 4 and 5 are actually only needed for 720p, since Case 3 already handles removing a dot before four numbers (ex. .2020 is same as .1080), but I think including 1080 (4 numbers) here makes it more robust.

### DIFF
--- a/resources/js/unit3d/helper.js
+++ b/resources/js/unit3d/helper.js
@@ -30,40 +30,24 @@ class uploadExtensionBuilder {
         let name = document.querySelector('#title');
         let torrent = document.querySelector('#torrent');
         if (!name.value) {
-            let fileEndings = ['.mkv.torrent', '.mp4.torrent', '.torrent'];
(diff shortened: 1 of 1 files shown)


HDVinnie:
thanks @werrpy

PR #1191, opened 2020-03-01
Fix error in check/delete of torrent after failing validation

staticza:
Check / delete of torrent file was referencing the incorrect disk.  It was referring to `torrent` instead of `torrents`.


HDVinnie:
@staticza thanks for the contribution.

cbj4074:
@staticza Great spot! Thank you for the careful review.

Unfortunately, these lines are not yet covered by tests, so this issue was overlooked.

We're covering the "best case scenario" execution path first, and then we'll come back and try to add coverage for the various failure scenarios.

Thanks again for the keen eye.

PR #1195, opened 2020-03-03
tests: Add CommentController test

cbj4074:


database/factories/CommentFactory.php
```diff
@@ -7,7 +7,7 @@
 $factory->define(App\Models\Comment::class, function (Faker $faker) {
     return [
         'content'    => $faker->text,
-        'anon'       => $faker->randomNumber(),
+        'anon'       => (int) $faker->boolean(),
```
cbj4074: @HDVinnie Can you confirm that this is supposed to be a boolean value?

database/factories/CommentFactory.php
```diff
@@ -20,7 +20,7 @@
         'playlist_id' => function () {
             return factory(App\Models\Playlist::class)->create()->id;
         },
-        'collection_id' => $faker->randomNumber(),
+        //'collection_id' => $faker->randomNumber(),
```
cbj4074: Can we eliminate this? It seems like this column may have been deleted since the Shift stubs were generated.

database/factories/TorrentRequestFactory.php
```diff
@@ -36,8 +36,8 @@
             return factory(App\Models\User::class)->create()->id;
         },
         'approved_when' => $faker->dateTime(),
-        'type_id'       => function () {
-            return factory(App\Models\Type::class)->create()->id;
-        },
+        //'type_id'       => function () {
+        //    return factory(App\Models\Type::class)->create()->id;
+        //},
```
cbj4074: Same as previous; can we delete these?

HDVinnie: yes

database/factories/CommentFactory.php
HDVinnie: yes

database/factories/CommentFactory.php
HDVinnie: boolean is correct

Issue #1192, opened 2020-03-02
[Bug] Side navigation bar no longer works as intended on Firefox


shlandturtle:
**Describe the bug**
Side navigation bar is constantly open (and can't be collapsed) when using Firefox. This issue arrive during one of the last 62 commits.
On Chrome and Chrome-based browsers this problem does not exist.

**To Reproduce**
1. Go to site using Firefox
2. Side navigation is wide open
3. Go to site using Chrome or it's derivative
4. Side navigation works as before

**Expected behavior**
Side nav should collapse and work on FF too.

**Screenshots**

**UNIT3D (please complete the following information):**
- UNIT3D Version: 2.2.7 (on commit 95033bd100e5cfc7d64f2df0f6340671de091646)
- Laravel Version: 6.17.0
- PHP Version: 7.4.2
- Database Driver & Version: mysql 5.7.29
- Web Server Driver & Version: nginx  1.15.8
- OS Driver and Version: Ubuntu 18.04

**Desktop (please complete the following information):**
 - OS: All of them.
 - Browser: Firefox 73
 - Version:

**Additional context**

**Possible Fix?**
Have a idea on how to fix? Write it here.

HDVinnie:
Not able to replicate on FireFox 72.0.2 (64-bit)

[![Image from Gyazo](https://i.gyazo.com/c6b0ec49a3e8af7759db96955db37ab7.gif)](https://gyazo.com/c6b0ec49a3e8af7759db96955db37ab7)

I can set side nav in settings just fine.
https://unit3d.site/users/UNIT3D/settings

HDVinnie:
I have discovered the issue.

HDVinnie:
It appears the PR by @werrpy https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/1188 uses lookbehinds aka `?<=` in regex. News to me as well but it appears look behinds are not supported in FF, Safari and others. 

For reference: 
https://bugzilla.mozilla.org/show_bug.cgi?id=1225665
https://caniuse.com/#feat=js-regexp-lookbehind

Have a solution @werrpy ? Otherwise will need to revert.

HDVinnie:
closed in https://github.com/HDInnovations/UNIT3D-Community-Edition/commit/63b84ffa679b2d54bbf43f9951a4b7a150f8b686

Issue #1193, opened 2020-03-02
[Bug] Forum post tipping does not work, or tips unrelated post instead


shlandturtle:
**Describe the bug**
Trying to tip a users post on forum does not do anything (FF) or alternatively the tipper gets redirected to another random post and tip goes to that instead (Chrome and derivatives). Seems like the ID's are somehow mixed up now.
For example topics/91?page=7#post-1175 redirects to and tips topics/13?page=1#post-91
This problem has been introduced during the last 62 commits.

**To Reproduce**
1. Go to a random post
2. Click on Tip This Poster
3. Nothing happens, or alternatively another post elsewhere gets tipped.

**Expected behavior**
Tip working and tipping the correct post.

**Screenshots**

**UNIT3D (please complete the following information):**
- UNIT3D Version: 2.2.7 (on commit 95033bd100e5cfc7d64f2df0f6340671de091646)
- Laravel Version: 6.17.0
- PHP Version: 7.4.2
- Database Driver & Version: mysql 5.7.29
- Web Server Driver & Version: nginx  1.15.8
- OS Driver and Version: Ubuntu 18.04

**Desktop (please complete the following information):**
 - OS:
 - Browser: Firefox 73, Chrome 80.0, Edge 80.0
 - Version:

**Additional context**

**Possible Fix?**
Have a idea on how to fix? Write it here.

HDVinnie:
This is linked to https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/1192#issuecomment-593559196

shlandturtle:
This was introduced in PR #1185 
This is caused by passing topic id in resources/views/forum/topic.blade.php#L48 and using that ID to look for post in app/Http/Controllers/BonusController.php#L489
Hence a post that has id of the topic you make a tip in gets tipped.
An easy fix would be to change BonusController.php line 489 to:
$post = Post::with('topic')->findOrFail($request->input('post'));
Proper fixing requires specifying the post in topic.blade.php and passing the correct ID, but this is beyond my skills.

cbj4074:
Ahh, indeed, that change in https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/8e569dd8d439e84618417659dbc52d295a4bf434/app/Http/Controllers/BonusController.php#L489 seems to be the cause.

[...]

PR #1196, opened 2020-03-03
(Fix) Use separate capture groups instead of negative look-behind

werrpy:
This re-writes the regex to use separate capture groups for each case, instead of negative look-ahead. Tested by defining the function in a firefox console on a new tab


werrpy:
#1192, #1193

HDVinnie:
Close #1192  #1193

HDVinnie:
@werrpy upon further testing this doesn't appear to fix support for Firefox and Safari.

PR #1197, opened 2020-03-03
(Add) Send notifications instead of chat announce for posts at staff area

shlandturtle:
Select whether to send notifications to staff members instead of announcing posts in chatbox for a given forum area, determined by ID. In future could/should be improved by having a no_announce flag for forum areas for example.

### DIFF
--- a/app/Notifications/NewPost.php
+++ b/app/Notifications/NewPost.php
@@ -74,6 +74,14 @@ class NewPost extends Notification implements ShouldQueue
             ];
         }
 
+        if ($this->type == 'staff') {
+            return [
+                'title' => $this->poster->username.' Has Posted In A Staff Forum Topic',
+                'body'  => $this->poster->username.' has left a new post in Staff Topic '.$this->post->topic->name,
+                'url'   => sprintf('%s?page=%s#post-%s', route('forum_topic', ['id' => $this->post->topic->id]), $this->post->getPageNumber(), $this->post->id),
+            ];
+        }
+
         return [
             'title' => $this->poster->username.' Has Posted In A Topic You Started',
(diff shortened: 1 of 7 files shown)


app/Notifications/NewPost.php
```diff
@@ -74,6 +74,14 @@ public function toArray($notifiable)
             ];
         }
 
+        if ($this->type == 'staffarea') {
+            return [
+                'title' => $this->poster->username.' Has Posted In A Staff Forum Topic',
+                'body'  => $this->poster->username.' has left a new post in Staff Topic '.$this->post->topic->name,
+                'url'   => sprintf('/forums/topics/%s?page=%s#post-%s', $this->post->topic->id, $this->post->getPageNumber(), $this->post->id),
```
cbj4074: If the route has a name (which this one does), it's ideal to use the route helper, to avoid hard-coding any portion of the URL, e.g.:

```php
'url'   => sprintf('%s?page=%s#post-%s', route('forum_topic', ['id' => $this->post->topic->id]), $this->post->getPageNumber(), $this->post->id),
```

HDVinnie:
> @HDVinnie I've reviewed this and I don't see any issues with the code (aside from the super-nitpicky style change regarding string concatenation and using `sprintf()` over the `.` operator).
> 
> Do you have any concerns with the business logic or addition of the feature itself?

Im fine with it. Just needs the styleci action to pass before merging.

shlandturtle:
Removed the extra line StyleCI complained about. I do not understand what the codecov error is about.

cbj4074:
Thanks! Don't worry about the Codecov error.

PR #1201, opened 2020-03-03
(Fix) Typos related to chat system announces

shlandturtle:
1) Nerdbot registrations command
2) Escape % in 100% freeleech.

### DIFF
--- a/app/Bots/NerdBot.php
+++ b/app/Bots/NerdBot.php
@@ -294,11 +294,11 @@ class NerdBot
     {
         $registrations = cache()->get('nerdbot-users');
         if (!$registrations || $registrations == null) {
-            $users = User::where('created_at', '>', $this->current->subDay())->count();
-            cache()->put('nerdbot-users', $users, $this->expiresAt);
+            $registrations = User::where('created_at', '>', $this->current->subDay())->count();
+            cache()->put('nerdbot-users', $registrations, $this->expiresAt);
         }
 
-        return sprintf('In The Last 24 Hours %s Users Have Registered To ', $users).config('other.title').'!';
+        return sprintf('In The Last 24 Hours %s Users Have Registered To ', $registrations).config('other.title').'!';
     }
 
     /**
--- a/app/Http/Controllers/TorrentController.php
+++ b/app/Http/Controllers/TorrentController.php
@@ -1556,13 +1556,13 @@ class TorrentController extends Controller
             $torrent->free = '1';
 
             $this->chat->systemMessage(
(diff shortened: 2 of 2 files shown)


app/Repositories/ChatRepository.php
```diff
@@ -313,9 +313,9 @@ public function systemMessage($message, $bot = null)
         if ($bot) {
             $this->message($systemUserId, $this->systemChatroom(), $message, null, $bot);
         } else {
-            $systemBotId = Bot::where('slug', 'systembot')->first()->id;
+            $systemBotId = Bot::where('command', 'systembot')->first()->id;
 
-            $this->message($systemUserId, $this->systemChatroom(), $message, null, $systemBotId);
+            $this->message($systemBotId, $this->systemChatroom(), $message, null, $systemBotId);
```
cbj4074: It looks like this change breaks several tests.

The method signature is `public function message($user_id, $room_id, $message, $receiver = null, $bot = null)`, and this change passes a Bot ID instead of a User ID.

It's entirely possible that https://github.com/HDInnovations/UNIT3D-Community-Edition/commit/aa32a893fa7767f91b4dcefcf76016179b6e8066 introduced a regression, in which case we should take a look and update the tests to describe the correct behavior if necessary.

Did you notice a specific regression due to the aforementioned commit @shlandturtle ?

app/Repositories/ChatRepository.php
```diff
@@ -313,9 +313,9 @@ public function systemMessage($message, $bot = null)
         if ($bot) {
             $this->message($systemUserId, $this->systemChatroom(), $message, null, $bot);
         } else {
-            $systemBotId = Bot::where('slug', 'systembot')->first()->id;
+            $systemBotId = Bot::where('command', 'systembot')->first()->id;
```
HDVinnie: Why this change?

app/Repositories/ChatRepository.php
shlandturtle: That was unneeded change which I did due to not realizing where the issues was. After the commit you linked all chat announces stopped working, but it was due to the "'slug', 'systembot'", not this edit. Removed ChatRepository changes from this commit now. My bad.

app/Repositories/ChatRepository.php
shlandturtle: Because it appears that in some earlier version of DB seeder or something, a slug with just "system" was created, with "command" systembot. Hence there was no bot named "Systembot", causing all chat announces to error. However, I notice at least the current seeder does slug "systembot", so might have been just my mixup or something. Using command instead of slug fixed this issue, but I guess it can just as well be fixed by editing the slug in the DB since that's what it should be. Hence I've removed this change from this commit. My bad.

HDVinnie:
Thanks for the contribution @shlandturtle

PR #1205, opened 2020-03-04
fix: correct units in various files

stonedbovines:
Change (K|M|G|T)B to (K|M|G|T)iB in various files, where incorrect


HDVinnie:
> I'm in favor of correcting the units, if necessary, but before we do that, we should ensure that the values are correct to begin with. I'm referring to the original value, `1073741824000`.
> 
> One tebibyte (TiB) is `1099511627776` bytes, and one terabyte is `1000000000000` bytes, so I'm wondering from where the base value originated.

Yeah Im not positive what happened there. I would recommend they are corrected within this PR. There are a few spots.

cbj4074:
Okay, in that case, it would probably be best to make the byte count a configuration value in `config/unit3d.php` and then reference it as `config('unit3d.bytes-per-tebibyte)`.

The *best* way to fix this would be to create a `bytes(int $number, string $unit)` method, but the first method would at least open the door for this more complete solution.

stonedbovines:
Ah, that base value is 1024 x1024 x 1024 x 1000
I hadn't even considered that the numbers might need fixing

cbj4074:
> Ah, that base value is 1024 x1024 x 1024 x 1000

Good call! 😆

Are you interested in trying to implement this "the right way" in this PR? Or would adding a new class and method that handles all the conversions we might need more than you'd bargained for?

To @HDVinnie 's point, if we don't fix it now, we probably never will... or at least it'll be harder to go back, dig-up this PR, determine everywhere that needs to be touched for a second time, and fix it proper.

Ultimately, if you're not up for tackling this @stonedbovines , my preference would be just to leave this open for a few days, and I'll fix it as soon as I have a chance.

stonedbovines:
> > Ah, that base value is 1024 x1024 x 1024 x 1000
> 
> Good call! 😆
> 

Are people not familiar with kilogibibytes? 😁 

> Are you interested in trying to implement this "the right way" in this PR? Or would adding a new class and method that handles all the conversions we might need be more than you'd bargained for?

"Quick and dirty" I could probably do. The right way is outside my skill set. I'd prefer to see it done properly. Who knows? I might even learn something from that.

cbj4074:
Okay, let's leave this open and I'll implement a solution soon.

HDVinnie:
Some spots I have found so far that have hardcoded bytes: @cbj4074 

- [ ] app/Console/Commands/AutoBonAllocation.php
- [ ] app/Console/Commands/AutoGroup.php
- [ ] config/other.php
- [ ] database/seeds/BonExchangeTableSeeder.php
- [ ] app/Http/Controllers/BonusController.php

Note the function used to convert bytes to human format in all the views is here:

- [ ] app/Helpers/StringHelper.php
- [ ] app/Helpers/MediaInfo.php
- [ ] app/Helpers/SystemInformation.php

cbj4074:
@HDVinnie Thanks, that list is very helpful.

Let's tackle the second list in a separate PR; the scope of this one has already grown quite large.

cbj4074:
@stonedbovines I tidied-up your commits (squashed them into a single commit), added the remainder of the features we discussed, and force-pushed to your development branch.

If you're not familiar with this workflow, you will need to `git reset --hard origin/development` to be able to push anything else to that branch. Or, you can just trash your branch and fork, if you don't need it after this. :)

PR #1233, opened 2020-03-09
(Fix) Pruning commands

shlandturtle:


app/Console/Commands/AutoDisableInactiveUsers.php
```diff
@@ -51,14 +51,13 @@ public function handle()
 
             $current = Carbon::now();
 
-            $matches = User::whereIn('group_id', [config('pruning.group_ids')]);
+            $matches = User::whereIn('group_id', [config('pruning.group_ids')])->get();
 
             $users = $matches->where('created_at', '<', $current->copy()->subDays(config('pruning.account_age'))->toDateTimeString())
-                ->where('last_login', '<', $current->copy()->subDays(config('pruning.last_login'))->toDateTimeString())
-                ->get();
+                ->where('last_login', '<', $current->copy()->subDays(config('pruning.last_login'))->toDateTimeString());
```
HDVinnie: don't you need `get();` here at end?

shlandturtle: No. I believe it's because of it being a collection and not eloquent.
Having get() there causes an error:
> TypeError: Too few arguments to function Illuminate/Support/Collection::get(), 0 passed in Psy Shell code on line 3 and at least 1 expected

However I think that it's recommended to have something there so I added all(). That works the same.

HDVinnie: Yes your correct I overlooked that.

PR #1237, opened 2020-03-10
(Enhancement) Only softdelete unused accounts when pruning

shlandturtle:
Check if user has contributed in any way, and only delete if not. Makes pruned group essentially same than banned.


HDVinnie:
@shlandturtle I will review this soon after I finish up some other things.

PR #1242, opened 2020-03-11
fix: Fix Tipping regression

cbj4074:
57212ca2631307c15ad6b9327129b82df6a25e84 introduced the following change:

https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/52c3d557944f2cba286573df3fa42f617918ce4e/app/Http/Controllers/BonusController.php#L489

This broke Tipping as described in:

https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/1193


cbj4074:
I can't easily test the front-end at the moment, but my recommendation is that we merge this, because it won't make the problem any worse, and could very well fix it.

PR #1248, opened 2020-03-21
(Enhancement) Adds torrents/edit API endpoint

PyR8zdl:
This adds torrents/edit API endpoint.
It will edit the torrents data if param is passed and the user is owner.

### DIFF
--- a/app/Http/Controllers/API/TorrentController.php
+++ b/app/Http/Controllers/API/TorrentController.php
@@ -20,6 +20,7 @@ use App\Helpers\TorrentTools;
 use App\Http\Resources\TorrentResource;
 use App\Http\Resources\TorrentsResource;
 use App\Models\Category;
+use App\Models\FeaturedTorrent;
 use App\Models\TagTorrent;
 use App\Models\Torrent;
 use App\Models\TorrentFile;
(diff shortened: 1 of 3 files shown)


PyR8zdl:
I still want to add an option to delete the torrent. Still working on the best way to do this.

app/Http/Controllers/API/TorrentController.php
```diff
@@ -241,6 +278,70 @@ public function show($id)
...
+     * @param int     $id
+     *
+     * @return TorrentResource
+     */
+    public function edit(Request $request, $id)
+    {
+        $user = $request->user();
+        if ($user->group->is_owner) {
+            $torrent = Torrent::findOrFail($id);
+            if ($request->filled('name')) {
```
PyR8zdl: @HDVinnie For these if statements I only want to set $torrent->x if $request->filled('x') with no else. I don't want to touch the data if the request is supplied.

I found this stackoverflow, which one should I adopt to? I know PHP7.4 will be required soon but in the meantime I don't want to break anything for others who haven't updated.
https://stackoverflow.com/a/36933323/12882632

PyR8zdl:
I'm working on torrent delete in API however I will not be able to test till Wednesday.
It's mostly a copy-paste from `/app/Http/Controllers/TorrentController.php @deleteTorrent`

Then I will start work on user API endpoint.

app/Http/Controllers/API/TorrentController.php
HDVinnie: its fine how it is since these are only ifs and not if else.

PR #1254, opened 2020-03-29
(Add) missingTorrentFiles.CSV

clandestine8:
**Route:** `/users/{user}/downloadHistoryTorrents`

When _**Downloading All**_ torrents from `/users/{user}/torrents`, if a torrent file is missing, instead of issuing an error and preventing the download, it will output into a CSV called `missingTorrentFiles.csv` into the Zip File along with the successfully located torrent files. 

This also fixes the issue where users wouldn't be able to download All Torrents if One or More files are missing.

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -2027,9 +2027,9 @@ class UserController extends Controller
                 }
             }
             if ($failCount > 0) {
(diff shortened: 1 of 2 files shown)


HDVinnie:
@clandestine8 style needs attention.

clandestine8:
@HDVinnie Fixed

PR #1261, opened 2020-04-01
(Update) Test Email command

PyR8zdl:
Changed Test Email command to use owner email set in env instead of using mysql query to pluck email using hardcoded default owner user ID.

This is needed incase a tracker decides to change owners. They can easily change the owner info in env file and test emails will just work.

### DIFF
--- a/app/Console/Commands/TestMailSettings.php
+++ b/app/Console/Commands/TestMailSettings.php
@@ -14,7 +14,6 @@
 namespace App\Console\Commands;
 
 use App\Mail\TestEmail;
-use App\Models\User;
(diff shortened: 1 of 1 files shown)


HDVinnie:
Its considered bad practice to make a direct call to the `.env` file. You should instead call the `other.php` config which access's the `.env` var.

```
    /*
    |--------------------------------------------------------------------------
    | Site email
    |--------------------------------------------------------------------------
    |
    | Email address to send emails
    |
    */

    'email' => env('DEFAULT_OWNER_EMAIL', 'unit3d@none.com'),
```

PR #1285, opened 2020-04-19
(Update) profile.blade.php

innocente205:
For review:
The current set of modal windows do not allow line breaks to be entered. Rather than change the rules for all these, this amendment routes pm requests to the main "pm.send" blade.


HDVinnie:
might as well remove the modal code too

PR #1286, opened 2020-04-19
(Add) Empty Inbox button to the Private Message inbox 

clandestine8:
Created function & route to provide the option of deleting all item in your inbox at once. 

Updated view to provide this feature to the end user.


clandestine8:
Sorry include commits not intended for this pull request

PR #1287, opened 2020-04-19
(Add) Empty Inbox button to the Private Message inbox

clandestine8:
Created function & route to provide the option of deleting all items in your inbox at once.

Updated view to provide this feature to the end-user.


clandestine8:
StyleCI wants me to break existing code which I didn't touch so I'm not going to do that

PR #1302, opened 2020-04-23
(Refactor) Album Controller : Operations priority might differ from w…

VictoRD11:
(Refactor) Album Controller: Operations priority might differ from what you expect.

### DIFF
--- a/app/Http/Controllers/AlbumController.php
+++ b/app/Http/Controllers/AlbumController.php
@@ -17,7 +17,6 @@ use App\Models\Album;
 use App\Services\Clients\OmdbClient;
 use Carbon\Carbon;
 use Illuminate\Http\Request;
-use Illuminate\Http\Response;
 use Illuminate\Support\Str;
 use Image;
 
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/AlbumController.php
```diff
@@ -135,7 +136,7 @@ public function destroy(Request $request, $id)
         $user = $request->user();
         $album = Album::findOrFail($id);
 
-        abort_unless($user->group->is_modo || $user->id === $album->user_id && Carbon::now()->lt($album->created_at->addDay()), 403);
+        abort_unless($user->group->is_modo || ($user->id === $album->user_id && Carbon::now()->lt($album->created_at->addDay())), Response::HTTP_FORBIDDEN);
```
HDVinnie, line 139: Why are we using `Response::HTTP_FORBIDDEN` here over `403`.

VictoRD11: I think it's easier to read the code, I can return it to 403.

HDVinnie: `403` is used commonly through out the codebase so it preferred. Unless going to refactor all usages.

VictoRD11: Ok, reverted to number.

PR #1305, opened 2020-04-24
Update AutoHighspeedTag.php

innocente205:
Is this a suitable fix for the error:

Creating default object from empty value {"exception":"[object] (ErrorException(code: 0): Creating default object from empty value at /var/www/html/app/Console/Commands/AutoHighspeedTag.php:54)

### DIFF
--- a/app/Console/Commands/AutoHighspeedTag.php
+++ b/app/Console/Commands/AutoHighspeedTag.php
@@ -51,9 +51,11 @@ class AutoHighspeedTag extends Command
 
             foreach ($torid as $id) {
                 $torrent = Torrent::where('id', '=', $id)->first();
-                $torrent->highspeed = 1;
-                $torrent->save();
-
+                if (isset($torrent)) { 
+                    $torrent->highspeed = 1;
+                    $torrent->save();
+                }
+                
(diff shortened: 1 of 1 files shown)


HDVinnie:
Yes proper fix but StyleCI needs attention. Also please read the https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/master/CONTRIBUTING.md

innocente205:
I'm afraid I don't understand "StyleCI needs attention", and I have read the Contributing page, but as I only edit on line in Github I don't think some of it applies. If you could be a little more specific I'd be happy to amend, otherwise I'd better leave it to the experts.

HDVinnie:
> I'm afraid I don't understand "StyleCI needs attention", and I have read the Contributing page, but as I only edit on line in Github I don't think some of it applies. If you could be a little more specific I'd be happy to amend, otherwise I'd better leave it to the experts.

For sure....below you can see the checks. StyleCI is failing due to code style. https://github.styleci.io/analyses/J2bd3E?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

As for linking guide look at PR template. (PREFIX)

PR #1311, opened 2020-04-26
(Fix) WishController.php

innocente205:
Check for IMDB entry to avoid 500 error


HDVinnie:
Im going to merge this for now. In the end the wish system needs to be refactored though. We only use TMDB anymore.

PR #1341, opened 2020-05-14
(Update) Subtitles System

PyR8zdl:
1. ~~Fixed typo in return route.~~ Reverted.
2. Made the use of filesystem consistent.
3. Removed unneeded copy to tmp directory.
4. Simplified mime-type and file return.

### DIFF
--- a/app/Http/Controllers/SubtitleController.php
+++ b/app/Http/Controllers/SubtitleController.php
@@ -120,7 +120,7 @@ class SubtitleController extends Controller
         ]);
 
         if ($v->fails()) {
-            return redirect()->route('subtitles.create', ['torrent_id' => $request->input('torrent_id')])
+            return redirect()->route('subtitle.create', ['torrent_id' => $request->input('torrent_id')])
                 ->withErrors($v->errors());
         }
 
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/SubtitleController.php
```diff
@@ -120,7 +120,7 @@ public function store(Request $request)
         ]);
 
         if ($v->fails()) {
-            return redirect()->route('subtitles.create', ['torrent_id' => $request->input('torrent_id')])
+            return redirect()->route('subtitle.create', ['torrent_id' => $request->input('torrent_id')])
```
HDVinnie, line 123: This should remain plural. 
```
        // Subtitles System
        Route::group(['prefix' => 'subtitles'], function () {
            Route::name('subtitles.')->group(function () {
                Route::get('/', 'SubtitleController@index')->name('index');
                Route::get('/create/{torrent_id}', 'SubtitleController@create')->where('id', '[0-9]+')->name('create');
                Route::post('/store', 'SubtitleController@store')->name('store');
                Route::post('/{id}/update', 'SubtitleController@update')->name('update');
                Route::delete('/{id}/delete', 'SubtitleController@destroy')->name('destroy');
                Route::get('/{id}/download', 'SubtitleController@download')->name('download');
                Route::get('/filter', 'SubtitleController@faceted');
            });
        });
```

Routes are plural and view returns are not.

PyR8zdl: Ah, I confused them the routes with views. I will change that back.

HDVinnie: Not a problem. Looks good otherwise.

PR #1369, opened 2020-06-01
(Refactor) PHP Announce

HDVinnie:
- WIP
- StyleCI will be addressed at end before merging


HDVinnie:
> This may dupe the peerlist when in `&compact=0` mode, because the ip filter only work in compact mode.
> In my opinon, when client in `&compact=0`, tracker should return all peer in `$rep_dict['peers']` and no `$rep_dict['peers6']` used.

@Rhilip is there clients that dont use compact in this age? Wouldn't it be better to just reject / throw an exception when a user tries to announce using an client that doesn't support compact? My guess is any clients that dont by default are EOL?

app/Http/Controllers/AnnounceController.php
```diff
@@ -48,446 +66,443 @@ class AnnounceController extends Controller
...
-            //info('Client Attempted To Connect To Announce Without A Passkey');
-            return \response(Bencode::bencode(['failure reason' => 'Please Call Passkey']))->withHeaders(['Content-Type' => 'text/plain']);
+    /**
+     * @param Request $request
+     *
+     * @throws \App\Exceptions\TrackerException
+     *
+     * @return void
+     */
+    protected function checkClient(Request $request): void
```
HDVinnie, line 132: So did some logging and am seeing so far that these two torrent versions are sending the `accept-language` header.

Client Headers: ["uTorrent/15399 CFNetwork/978.2 Darwin/18.7.0 (x86_64)","en-us",null,null,null]	
Client Headers: ["uTorrent/3220(28500)","en-CA",null,null,null]

Issue #1472, opened 2020-10-29
[Bug]on Media HUB you cannot click the other pages.


Gideon75:
**Describe the bug**
[Bug]Pages do not work in any section of the media HUB

**To Reproduce**
1. Go to MediaHUB
2. Click on a section
3. Scroll down at the end of the page
4. you cannot click the other pages

**Screenshots**
[https://gyazo.com/ec3c740a19b306e1a4c7fb3d2eff9172]
**UNIT3D (please complete the following information):**

OS
Currently Running
Linux
PHP
Currently Running
php7.4.3
DATABASE
Currently Running
8.0.21-0ubuntu0.20.04.4
LARAVEL
Currently Running
Ver. 8.11.2
RAM
Total: 62.6 GiB
Free: 23.19 GiB Used: 39.41 GiB
DISK
Total: 436.34 GiB
Free: 359.25 GiB Used: 77.08 GiB
LOAD
Average: 1.37
Estimated

PR #1473, opened 2020-10-30
fix: missing livewire assets/configs

leighmacdonald:
Generated via:
`php artisan vendor:publish --tag=livewire:config
php artisan vendor:publish --tag=livewire:assets`


HDVinnie:
Closes #1472

PR #1476, opened 2020-10-31
update: Allow configuring of echo-server address

leighmacdonald:
Enables custom laravel-echo-server address without hard coding. Uses .env config.


HDVinnie:
Ill merge this soon. Have to update installer stub at same time so no conflicts.

PR #1479, opened 2020-11-02
fixup seeders and cleaned them up as well

iamuser00110001:
I installed unit3d and immediatly saw some problems in the seeders. After doing a second ```migrate --seed``` it crashed so i fixed that.


HDVinnie:
@iamuser00110001 thanks for your contribution.

PR #1551, opened 2021-01-02
(Fix) Mass PM Attributes

Spidi:
Due to the refactor a couple of days ago there is now a mismatch between constructor params and the attributes used in the handle function


HDVinnie:
@Spidi thanks for the PR

Issue #1448, opened 2020-10-01
[Bug] Quoting a comment that contains a quote results in badly formatted posts


Tzuriel25:
**Describe the bug**
When you add the text to quote after the quote (which is where it goes by default) the new text appears to the right of the original.  If it's long, it starts to scroll under but it looks really bad.

**To Reproduce**
On Demo site: https://unit3d.site/forums/topics/2?page=1#post-4
1. Go to Forums
2. Create a Topic with some text
3. Quote that comment
4. Quote the quoted comment
5. Formatting error as above
**Expected behavior**

**Screenshots**
Original being quoted looks fine:
![image](https://user-images.githubusercontent.com/43972449/94866782-9cfb6200-0405-11eb-9749-d8582b1893f4.png)

Quoting that post with a new comment formats badly:
![image](https://user-images.githubusercontent.com/43972449/94866832-b3a1b900-0405-11eb-8f9e-18827a8ce4f9.png)

**UNIT3D (please complete the following information):**
- UNIT3D Version: Currently Running UNIT3D Community Edition (Nex-Gen Torrent Tracker) v3.0.0
- Laravel Version: Ver. 7.28.3
- PHP Version: php7.4.3
- Database Driver & Version: 8.0.21-0ubuntu0.20.04.4
- Web Server Driver & Version: 
- OS Driver and Version: UBUNTU 20.04

**Desktop (please complete the following information):**
 - OS: WIndows 10
 - Browser: Chrome
 - Version: 85.0.4183.121 (Official Build) (64-bit)

TPS-Zappateers:
example: Pressing "quote" button on a post containing;

> [quote=@author#1]1st comment[/quote]
> 2nd comment

results in;

> [quote=@author#2][list]                          [*]                                                     [b][i][/i] Quoting @author#2 :[/b]                              1st comment                                                                              [/list]
> 2nd comment[/quote]

where I would expect to see;

> [quote=@author#2][quote=@author#1]1st comment[/quote]
> 2nd comment[/quote]
> 3rd Comment

So, author#1 is lost, and [list] & an empty [i] is introduced.
Depending on the length of the 3rd comment, it can appear as tzuriel explains, or can be lost completely.

Really dificult to explain any kind of work-a-round to members for this bug!

PR #1552, opened 2021-01-02
fix: formatting of multiple nested quotes

Spidi:
This aims to solve the issues in #1448 

Rather than making it more complicated than it needs to be for now.. simply add the raw bbcode that's pulled from the database as a data attribute and then when quoting simply push that to the comment box rather than the text content of the html node. 

I'm happy for people to test this out though if they wish. I'm also not 100% sure what the replacement of '@ here' was about but i'm assuming it was some sort of side effect of getting the data from html?

### DIFF
--- a/resources/views/forum/topic.blade.php
+++ b/resources/views/forum/topic.blade.php
@@ -124,7 +124,7 @@
                                 </span>
                             </aside>
 
-                            <article class="col-md-9 post-content">
+                            <article class="col-md-9 post-content" data-bbcode="{{ $p->content }}">
                                 @joypixels($p->getContentHtml())
                             </article>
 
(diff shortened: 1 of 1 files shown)


HDVinnie:
closes #1448

Tzuriel25:
I updated our topic.blade.php with this today.  Is this dependent on the other code that has been added to topic.blade.php from v5?  Because it didn't fix anything and might have made it worse

Spidi:
> I updated our topic.blade.php with this today. Is this dependent on the other code that has been added to topic.blade.php from v5? Because it didn't fix anything and might have made it worse

@Tzuriel25 i pulled the latest code before doing this so i'm unsure if it required anything else in V5. What is it you are seeing instead?

Tzuriel25:
Yep.  It works fine on the demo site, it seems.  I will need to upgrade to the current codebase.  Thanks!

PR #1553, opened 2021-01-02
Fix: unique tables

JamesPSS:
This is fix so that username can't be the same. 
+ some other tables that need to be uniqu.

You can change the username if you staff to anything whitout checks and the system don't handle that well. 

It will be smart that mails is unique to but with now when you create a new site the system use the same mails 3 times so i did not add the unique mail.


HDVinnie:
@JamesPSS I agree and thanks for the PR. The database schema will need tweaking soon with uniques and proper indexing.

PR #1563, opened 2021-01-04
(Update) Torrent Featuring

JamesPSS:
This will make the button to revoke feature work.

+ add .ftpconfig to the  .gitignore  so my atom edtior will work with sftp.


HDVinnie:
Thanks for the PR. In future PRs try to have all checks passing. While styleci will auto fix once merge its best to try and correct it directly in the PR before merged.

PR #1575, opened 2021-01-07
(Fix) Pages Model

JamesPSS:
This fix page model so it return the right html. 

It use the BBCodeConverter before that is not working anymore. I did not find any files else using it so you can remove it if you don't need it anymore.

### DIFF
--- a/app/Models/Page.php
+++ b/app/Models/Page.php
@@ -13,7 +13,8 @@
 
 namespace App\Models;
 
-use App\Helpers\BBCodeConverter;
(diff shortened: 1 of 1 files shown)


HDVinnie:
This is not a fix? Your removing the possibility you use Markdown in pages. The point of this is so staff can use Markdown and BBCODE in there pages.

JamesPSS:
Okay. then i will fixe it. But do you need to use markdown in the articles too?.

PR #1668, opened 2021-03-14
(Feature) Add Staff & User Ticket Notifications

Obi-Wana:
Adds Notifications (similar to new PM's) for new & Updated Tickets.
A new Migration has been created adding a "user_read" & "staff_read" column to the Ticket DB Table.

### DIFF
--- a/database/migrations/2021_03_14_093812_add_read_column_tickets_table.php
+++ b/database/migrations/2021_03_14_093812_add_read_column_tickets_table.php
@@ -1,0 +1,34 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddReadColumnTicketsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('tickets', function (Blueprint $table) {
+            $table->tinyInteger('user_read')->nullable()->after('staff_id');
+            $table->tinyInteger('staff_read')->nullable()->after('user_read');
+        });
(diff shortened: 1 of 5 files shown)


resources/views/partials/top_nav.blade.php
```diff
@@ -58,9 +76,16 @@
                     </a>
                 </li>
             @endif
+
```
HDVinnie, line 79: Double check this PR please. Theres code that shouldn't belong.

Obi-Wana:
@HDVinnie Removed the code that shouldn't belong into the top-nav.

database/migrations/2021_03_14_093812_add_read_column_tickets_table.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('tickets', function (Blueprint $table) {
+            $table->tinyInteger('user_read')->nullable()->after('staff_id');
```
HDVinnie, line 17: Would it not be more appropriate if these two new columns where a boolean with a default of 0 or false which would also allow you to get rid of these lines. https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/149f3a3c09dcac2cfbc2a0cc9f436e51ebca24b4/app/Http/Controllers/CommentController.php#L556-L561 ?

Obi-Wana: @HDVinnie I have used the table "private_messages" as template, where the "read" column is a tinyint. besides having less code, i dont see much of a performance improvement in using a bool with default 0.
However, I'm not that experienced yet. Decision's up to you.

PR #1676, opened 2021-03-17
An Option to cleanup the Menu a bit

LewdNeko-chan:
Because mobile devices cannot access some menu points, because its out of view.
Here is an Example of the new Sidemenu:
image


LewdNeko-chan:
I am not very sure, why the greetings bot fails... but I am sure, that this works on the local testinstance, I am working with @Alkl58.

PR #1713, opened 2021-04-11
(Fix) Metadata: placeholders for actor photos, make them clickable

Oha-you:
Fixes two minor issues:
1. Broken actor images if TMDB doesn't have any photos (e.g. https://www.themoviedb.org/tv/111027-i-love)
2. Make actor photos clickable, same as names (easier to click, no need to hover names only)

Before:

After:


Oha-you:
Looks like Octokit doesn't have enough permissions to comment here (since it's my first PR), so the first check failed:

> Thank you for submitting your first pull request. Please make sure it follows our Contributing Template and you have tested the code you are issuing this pull request on.

Honestly, I haven't checked the fix on a live tracker, since the changes are very basic. I'm sure there will be no issues.

PR #1715, opened 2021-04-13
(Update) About Me in Profile: don't limit to 500 chars, add wysibb, remove badge style

Oha-you:
I see no logical reason why About Me column should be limited only to 500 chars.
Uploaders want to use this field to specify their upload schedule and other useful info.
Forum signature doesn't have such limitations (though it makes more sense to limit it).

Other fixes:
- added wysibb support
- removed unnecessary labels that made whole text inside textarea bold
- removed badge style since it applies unnecessary `text-align: center` and `white-space: nowrap` styles

Tested using current master branch on Ubuntu 20.04.2 (used UNIT3D Installer).

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/database/migrations/2021_04_13_200421_update_about_column_on_users_table.php
+++ b/database/migrations/2021_04_13_200421_update_about_column_on_users_table.php
@@ -1,0 +1,30 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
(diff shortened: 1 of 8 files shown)


HDVinnie:
@Oha-you Thanks for the contribution.

PR #1726, opened 2021-04-14
(Add) Validate if files inside torrents have valid filenames

### LINKED ISSUE #1666: [Feature] Check if Torrent contains invalid filenames (important for Windows) Low Priority
**Describe the bug**
A user could potentially upload a Torrent, where files with filenames like "NUL", "PRN" or "CON" (and many more) are inside.
The problem is, that downloading these Torrents on a Windows Host will either result in the Torrent Client not downloading it, or the Torrent Client creating these "forbidden" files.  **Deleting these "forbidden" files on a Windows Host is not possible**, meaning the user would need to boot into a Linux Live system and delete them manually.

The probability of this happening is low, but it could (hopefully) be easily mitigated. Thats why I would say that this is a low priority.

**To Reproduce**
1. Create a Torrent on a Linux Host containing filenames like "NUL" (without extension)
2. Upload it to UNIT3D
3. Download the Torrent on a Windows Host

**Expected behavior**
This behavior was expected

**Screenshots**
Image

**Desktop (please complete the following information):**
 - OS: any Windows Host
 - Browser: na
 - Version: na

**Additional context**
Here some additional information: [Github](https://gist.github.com/doctaphred/d01d05291546186941e1b7ddc02034d3#file-ntfs-filenames-txt) and [Microsoft Docs](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file)

**Possible Fix?**
As it seems that you already parse the contents of a Torrent file, you could add additional filename checks.

Oha-you:
Fixes #1666.

Found and fixed a possible bug in bool logic where torrent extension is checked.
`$requestFile->getError() != 0` always returned `true` in my case so the part after `&&` was never checked.

Also caught an error when `Bencode::get_meta` was called on non-torrent files.
Added a try/catch block since I noticed `API/TorrentController.php` had it already.

Attaching a .zip with examples of torrent files containing bad filenames (for testing):
- [bad_torrents.zip](https://github.com/HDInnovations/UNIT3D-Community-Edition/files/6313925/bad_torrents.zip)

### DIFF
--- a/app/Helpers/TorrentTools.php
+++ b/app/Helpers/TorrentTools.php
@@ -183,4 +183,24 @@ class TorrentTools
 
         return $fileContent;
     }
+
+    /**
+     * Check if the filename is valid or not.
+     *
+     * @param $filename
+     *
+     * @return bool
+     */
+    public static function isValidFilename($filename)
+    {
+        $result = true;
+        if (\strlen($filename) > 255 ||
+            \preg_match('#[/?<>\\:*|"\x00-\x1f]#', $filename) ||
+            \preg_match('#(^\.+|[\. ]+)$#', $filename) ||
+            \preg_match('#^(con|prn|aux|nul|com[0-9]|lpt[0-9])(\..*)?$#i', $filename)) {
+            $result = false;
+        }
+
+        return $result;
+    }
 }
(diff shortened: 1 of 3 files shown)


Oha-you:
Note: used these sources as a reference for regexes:
- https://github.com/parshap/node-sanitize-filename/blob/master/index.js
- https://github.com/sindresorhus/filename-reserved-regex/blob/main/index.js

PR #1746, opened 2021-04-29
(Fix) Profile: optimize SQL query for uploads count

Oha-you:
Solves the problem of Profile pages not being rendered for users with lots of uploads (12500+).
<details>

</details>

Probably the initial MySQL query hits some limit and fails (without logging any errors).
It may be fixable by editing server settings, but since we need only the counter, it can be optimized. No need to do a costly `select * from torrents`.

<details>
<summary>Before</summary>

</details>
<details>
<summary>After</summary>

</details>

There are probably other places where such `->count()` calls can be replaced with better queries.
Laravel Debugbar can help identify them. If I have time, I'll work on it in future.

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -52,14 +52,11 @@ class UserController extends Controller
      */
     public function show($username)
     {
-        $user = User::with(['privacy', 'history'])->where('username', '=', $username)->firstOrFail();
+        $user = User::with(['privacy', 'history'])->withCount('torrents')->where('username', '=', $username)->firstOrFail();
 
         $groups = Group::all();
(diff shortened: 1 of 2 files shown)


HDVinnie:
You could in theory just eager load the count here https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/efcb5db45bd3de5b924271fb3cecc3bce9d52d93/app/Http/Controllers/UserController.php#L55 by just adding a withCount(‘torrents’) so we don’t have a whole new variable and query. Then just call $torrents_count in profile. Double check my syntax though as I’m on my mobile lol

Oha-you:
@HDVinnie heck yes! I like this way much more. Tried to use `->withCount()` at first, but failed to find where to insert in properly, haha.

This made the query even faster. And solves the blank page too:

PR #1748, opened 2021-05-05
Show if Peer is Connectable in Peerlist

Alkl58:
**This PR should be checked before merging.**

Greetings,

I thought it would make a nice addition to see if the port of the peers are open or not.

Note that I am not familiar with the best practices of PHP / Laravel.

Another thing came into my mind is, if this could cause issues server side, by trying to connect to the port of the client each time the peer list is being opened.

Also I am not sure, whether or not the connection is being properly terminated `fclose($fp)`.

Perhaps a better implementation would make more sense, where the connectability is checked once per user, which is then stored in some database temporarily. (which then can be seen in the user profile or something...)


leighmacdonald (changes requested):
The feature itself is certainly useful, but a couple thoughts that should be addressed before i would consider it:

1. I think you need to be very careful when adding a feature like this due to the nature of the things people typically do with this software. If you are connecting blindly like this, you will expose your origin server if you are not proxying your outgoing requests through some sort of TCP proxy/relay/vpn. Most admins are probably only going to be running a reverse http proxy and not something like this. Its a not entirely trivial thing as far as added complexity for setup. At a minimum i think this should at least be documented, very clearly, so the admins running the software can make the decision themselves knowing the caveats/risks.

2. This logic should not happen in the view, or even in the controller really as it will murder (perceived) performance waiting for all the connections to complete before serving the users request. Which is basically what you alluded to in your last paragraph and is mostly correct. Using some sort of background service (laravels cron thing maybe?) is probably the best way to handle this. This way the result can be cached. Probably just need to add a `connectable` column on `Peer` and the same in the database. It should be stored by peer because its possible for users to run multiple clients, so instead it should be checked periodically, per unique peer_id,

3. Try not to use @ to hide warnings. Though i realize php can sometimes make it required. `socket_connect` is probably better than `fsockopen`, though ive not really done socket stuff in php. We very much will want to set a reasonable timeout for this. Something like this (untested)

```php
                $socket  = socket_create(AF_INET, SOCK_RAW, 1);
                socket_set_option($socket, SOL_SOCKET, SO_RCVTIMEO, array('sec' => $timeout, 'usec' => 0));
                $is_connectable = socket_connect($socket, $host, null);
```

Issue #1542, opened 2020-12-29
[Bug] Torrent Description Linux Line-breaks not applied


Obi-Wana:
**Describe the bug**
If i post a torrent using mechanize (& unit3d API) the line breaks within the description are not applied and the whole content is on one single line.
The content for posting is copy & pasted from some nfo data.

**To Reproduce**

**Expected behavior**
Line breaks are applied as on the nfo where i copied the text from.

**Screenshots**
[https://i.imgur.com/j1kTpOQ.png](https://i.imgur.com/j1kTpOQ.png)
Screenshot from unit3d.site: [https://unit3d.site/torrents/2](https://unit3d.site/torrents/2)

**UNIT3D (please complete the following information):**
- UNIT3D Version: 4.0.0 & 5.0.0
- Laravel Version:
- PHP Version: 7 & 8
- Database Driver & Version:
- Web Server Driver & Version:
- OS Driver and Version:

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser:
 - Version:

**Additional context**

**Possible Fix?**
Have a idea on how to fix? Write it here.

Oha-you:
Just checked this behavior both at https://unit3d.site/ from Windows and locally from Ubuntu: **can't reproduce**.

Copied text with LF line endings, pasted to torrent descriptions and forums posts (with BBCode off and on): each time line breaks were present.

I think issue is no longer relevant, can be closed. Ping: @HDVinnie

PR #1757, opened 2021-05-12
(Fix) Convert unix-style line breaks to <br> for BBcode

Oha-you:
Bug was mentioned in #1542, I just didn't understand it correctly.
If you write torrent descriptions on Ubuntu, for example, they are not correctly converted to HTML (`\n` lines are left instead of `<br>`).
Adding an additional rule after `\r\n` fixes this issue.


HDVinnie:
@all-contributors please add @Oha-you for code

PR #1765, opened 2021-05-18
(Fix) Set Light theme as default for new users

Oha-you:
I know it's subjective but Light theme is more suitable for newly registered users, IMO.
Those who like dark styles can always change it later in their profiles.
Also Light theme is the only one that uses randomized images in footer (other themes have `background-image: none`).

### DIFF
--- a/config/other.php
+++ b/config/other.php
@@ -156,7 +156,7 @@ return [
     | 7 = Dark Teal Theme
     | 8 = Dark Yellow Theme
     */
-    'default_style' => 2,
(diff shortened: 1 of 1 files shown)


hucstar:
Ah，I personally like dark style. I think it is  so natural that have not notice it😁. but  still thank you for your contribution！In addition ，I wonder if The guru have any thoughts on the bug of Album to fix it?

https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/1457！

PR #1767, opened 2021-05-18
(Fix) A batch of front-end fixes for torrents and forum posts

Oha-you:
See individual commits to know what's fixed (I didn't squash them on purpose).
Tested on a live tracker (deployed 4 days ago), so there shouldn't be any bugs.


HDVinnie:
@Oha-you much appreciated....will make it easier for me to review tonight before merging.

PR #1774, opened 2021-05-23
(Fix) Remove line breaks from BBcode, add noreferrer to Linkify

Oha-you:
Fixes these issues:
1. Extra `<br>` tags are added to forum quotes and youtube embeds (which breaks formatting and Linkify behavior)
2. It's best to use "noopener noreferrer" attributes to prevent official websites to see where the links have been posted


Oha-you:
Noticed one more improvement that should be done to Linkifier, closing for now.

PR #1830, opened 2021-07-08
(Add) Flush Peers for Users

Alkl58:
**Please only merge it if tested beforehand.**

This PR Adds the option for users to flush their own "ghost peers".

grafik

There is a limit of 2 flushes per day.

Announce has to be at least 70 minutes old.

A new column `own_flushes` is created in the `users` table with the type `tinyInteger`.

Auto Reset is done with the command `auto:reset_user_flushes` which is being executed daily.

**It's also ok if you don't merge this.**


HDVinnie:
LGTM

PR #1831, opened 2021-07-09
Fix for .env.testing

imanghafoori1:
laravel has no `file` driver for  broadcasting, I think it was meant to be the `log` driver.
setting `file` driver causes error.


HDVinnie:
Good catch. Thank you.

PR #1832, opened 2021-07-09
(Fix) function_exists

imanghafoori1:
discovered by the laravel-microscope


HDVinnie:
Curious about the logic behind this. Why is camelCase a issue vs underscore case?

imanghafoori1:
They must be exactly the same strings, for obvious reasons.
You are trying to avoid redefining a function name by that ifs

PR #1838, opened 2021-07-13
(Refactor) Routes

HDVinnie:
- register routes using static ::class references instead of strings and remove any namespace prefixes from the RouteServiceProvider. This convention was adopted by default in Laravel 8. Converted controller action strings to array tuples for improved code completion and static analysis.


HDVinnie:
@clandestine8 performed some new updates to fix conflicts with imports.

clandestine8:
Yeah I think it is better this way. Nice and clear what is happening as well.

clandestine8 (approved):
Looks good to me. 👍

PR #1844, opened 2021-07-17
[Improvement] Set default value for personal_release

Obi-Wana:
The Column "torrents.personal_release" is set to be "notNull" and should IMHO have a default value of "0" to prevent errors.
Eg. it does have an impact when running artisan tests.


HDVinnie:
Totally agree. Don’t know how I missed that.

PR #1852, opened 2021-07-31
(Feature) Add Internals System for Staff

Obi-Wana:
This "internals system" allows staff members to create new internal groups. Users can then be assigned to those groups (user -> edit).

All users, which belong to the same group, will then be displayed as one group within the "internals" page.

[Internals Page](https://ibb.co/Btp2nGx)
[Staff Page](https://ibb.co/vX2YKXm)

### DIFF
--- a/app/Http/Controllers/PageController.php
+++ b/app/Http/Controllers/PageController.php
@@ -58,9 +58,10 @@ class PageController extends Controller
      */
     public function internal(): \Illuminate\Contracts\View\Factory | \Illuminate\View\View
     {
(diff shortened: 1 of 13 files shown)


Obi-Wana:
WIP on CI errors

HDVinnie:
Been thinking on this....what if a user belongs to one or more internal groups? Maybe instead oof having internal_id on users table it might be better to have an internal_users pivot table?

Obi-Wana:
@HDVinnie I've not had a use case for this yet, so did not think much about it yet.

Do you think it's needed? Tho, having a pivot table would be a more clean solution anyway, isnt it?

PR #1869, opened 2021-08-11
Apply early return rule

imanghafoori1:
flattens the code using early return rule


HDVinnie:
@imanghafoori1 Did you use a tool for IDE for this as the formatting is way off.

imanghafoori1:
Sorry, I will fix it.

imanghafoori1:
In fact the refactoring process is done by the laravel-microscope and the `php artisan check:early` command

PR #1884, opened 2021-09-03
(Update) Condense and rearrange torrent general info

Roardom:
Rearranges the torrent general info layout to be more condensed.

Changes:
- Hovering over "n days ago" shows you the ISO timestamp
- Hovering over the icons show the full name, so "Featured for x days", "Freeleech", etc
- Hovering over the size gives you the estimated ratio after download
- Hovering over the freeleech token buttons shows you how many fl tokens you have
- Hovering over the seeder, leecher, times figures shows "n seeders", n leechers", "n times" respectively
- Seed and leech counts links to the peers page
- Times count link to the history page
- Moderation buttons moved to dedicated panel
- Moved torrent name front and center

Before:
rearrange_before

After:
rearrange_after

### DIFF
--- a/resources/views/torrent/torrent.blade.php
+++ b/resources/views/torrent/torrent.blade.php
@@ -121,11 +121,14 @@
         </div>
 
         <div class="meta-general box container">
-            <div class="panel panel-chat shoutbox torrent-general">
+            <div class="panel panel-chat shoutbox torrent-general" x-data="{ show: false }">
                 <div class="panel-heading">
-                    <h4><i class="{{ config("other.font-awesome") }} fa-info"></i> @lang('torrent.general')</h4>
+                    <h4 style="cursor: pointer;" @click="show = !show">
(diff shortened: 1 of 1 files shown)


clandestine8:
I think the icons don't provide enough context for what they represent. Adding labels or additional context to the icons would help. Users shouldn't have to hover to be able to read the information on the screen. 

I would like to see the expanded view preserved behind a toggle, especially since there is information that is not carried over to the new layout. This could also allow for further condensing as the less important details could be hidden in the expanded view. 

We should also get some feedback from the sponsors since this is a key area of the project.

Roardom:
The legacy layout can now be toggled by clicking on the "General" header similar to how the raw MediaInfo works

clandestine8 (approved):
Now with the toggle added, I think we are good with this.

HDVinnie (approved):
LGTM

PR #1929, opened 2021-09-29
Attempt to read property "name" on array

maxpayneovh:
Attempt to read property "name" on array (View: /var/www/html/resources/views/category/show.blade.php)


HDVinnie:
Im going to close this because I am going to refactor this view to be like the torrent list page so its consistent.

PR #1957, opened 2021-10-22
(Add) RSS filter: bookmark

ReileenKawahara:
An RSS filter that looks for torrents in user's bookmarks.
Allows for the creation of "Bookmarks" RSS feed, just leave other filters empty.

#1624


ReileenKawahara:
@HDVinnie Done. Sorry I forgot about that one.

HDVinnie:
> @HDVinnie Done. Sorry I forgot about that one.

All good thanks for the contribution

PR #2013, opened 2021-11-22
(Add) 25% and 50% promo v0.2

Obi-Wana:
Will add 25% and 50% Freeleech promo.

- Searchable in Livewire (Graveyard & Torrent List search)
- Added Buttons for Staff (FL buttons have been updated to be blocked when 1 promo is already active, prevent combination of 25 and 50 and 100% FL)
- Add migration for new DB fields
- Add math in Announce Jobs
- Did also replace some "tab" with 4 spaces to match PSR2 Style guide:

>Code MUST use 4 spaces for indenting, not tabs.
>Moved some labels such as FL, DU, etc. to be all in the same order on graveyard & list view.

- Did rearrange the discount icons a bit in order to have them all in the same order in graveyard-list, torrent-list and torrent.blade.php

Tested with 50% & 25% Promo heavy_check_mark


Obi-Wana:
@HDVinnie I'm not sure if this is really needed:
https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/3079e0d7fdc8f56b043fc1b3dce6555b0b4433cf/app/Models/Torrent.php#L478

For 25% and 50% FL. I did copy it from above, but it does not have any function yet.

app/Models/Torrent.php
```diff
@@ -463,4 +467,32 @@ public function isFreeleech($user = null)
...
     }
+
+    /**
+     * Torrent Is Promo 50%.
+     *
+     * @param null $user
+     *
+     * @return bool
+     */
+    public function isPromo50($user = null)
```
HDVinnie, line 478: `$user = null` is not needed here. `$pPromo50 = $user ? $user->group->is_freeleech || PersonalFreeleech::where('user_id', '=', $user->id)->first() : false;` can be removed as well. `return $this->free || \config('other.freeleech') || $pfree;` should just be `return $this->promo50`

app/Models/Torrent.php
```diff
@@ -463,4 +467,32 @@ public function isFreeleech($user = null)
...
+    }
+
+    /**
+     * Torrent Is Promo 25%.
+     *
+     * @param null $user
+     *
+     * @return bool
+     */
+    public function isPromo25($user = null)
```
HDVinnie, line 492: `$user = null` is not needed here. `$pPromo25 = $user ? $user->group->is_freeleech || PersonalFreeleech::where('user_id', '=', $user->id)->first() : false;` can be removed as well. `return $this->free || \config('other.freeleech') || $pfree;` should just be `return $this->promo25`

Obi-Wana:
@HDVinnie So it's gonna be just like this then

https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/2013/commits/9e9b3e46baf6d0ccb6e5891900048ce57769efff

Obi-Wana:
BTW, I did not add the Promotions to the API yet, as it is with the "Featured" and "FreeLeech" option.

Obi-Wana:
> > @HDVinnie So it's gonna be just like this then
> > [9e9b3e4](https://github.com/HDInnovations/UNIT3D-Community-Edition/commit/9e9b3e46baf6d0ccb6e5891900048ce57769efff)
> 
> LGTM. Im going to do a final test before merging once home from work tonight. Thanks for the PR.
> 
> Im thinking the TorrentController at this point is getting way too big. I might pull out the `buff/promo` functions to a new `TorrentBuffController` to clean it up.


Sure, please give it a try before merging it. Would be a shame if it breaks the announcements.
Wouldn't hurt to split it, maybe some people might add other Buff-Option like XX% discount in the future.

Obi-Wana:
Did you guys had the time to test it, yet?

Gideon75:
I'm interested in and I could have time to try it, what changes do I have to do?

K4Li:
Tested this and gives 500 error on  both 25% and 50% promos.

Method App\Http\Controllers\TorrentController::grant25 does not exist. {"userId":560,"exception":"[object] (BadMethodCallException(code: 0): Method App\\Http\\Controllers\\TorrentController::grant25 does not exist. at /var/www/html/vendor/laravel/framework/src/Illuminate/Routing/Controller.php:68)

Also disables the ability to revoke promos if "featured" is activated. 

In our case:

            $torrent->free = '0';
            $torrent->doubleup = '0';

FL and Double Up are not auto set and we manually apply as needed. If Featured is revoked, manually applied FL or DU are removed also.

Obi-Wana:
@K4Li Thank you for your review.

When a torrent gets featured, it applies free,DU & featured on DB level (set to "1" each).
As the Featured Banner says "all featured files are 100% Free & DU", IMO we do not want to disable any of those. So the only option needed would be to "Unfeature" it and then add other Promos but "Feature".

Unfeaturing a file, will remove all previous applied BUFFs, as it is currently:
https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/cb0ce6e48275f13ea15278cca497a741ef6b6d63/app/Http/Controllers/TorrentController.php#L943
We can surely improve this, but maybe within another PR.

The "revoke" buttons get disabled in some cases when they are not needed. Tho may still show "revoke" as for example when featured, "FL" is also applied to it.

K4Li:
> @K4Li Thank you for your review.
> 
> When a torrent gets featured, it applies free,DU & featured on DB level (set to "1" each). As the Featured Banner says "all featured files are 100% Free & DU", IMO we do not want to disable any of those. So the only option needed would be to "Unfeature" it and then add other Promos but "Feature".
> 
> Unfeaturing a file, will remove all previous applied BUFFs, as it is currently:
> 
> https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/cb0ce6e48275f13ea15278cca497a741ef6b6d63/app/Http/Controllers/TorrentController.php#L943
> 
> 
> We can surely improve this, but maybe within another PR.
> The "revoke" buttons get disabled in some cases when they are not needed. Tho may still show "revoke" as for example when featured, "FL" is also applied to it.

But in our case, we have featured set to NOT auto apply double up and fl so we have the ability to set the promo ourselves. In the event any tracker wants to "feature" a torrent to put it in the slider, but only apply 50%, as an example, this is not possible. I just don't think removing featured should remove manually applied promos and should not disable the ability to revoke promos. I'm going to work on this.... thanks for the pr!

Obi-Wana:
@K4Li 
Talking on discord about this right now. I'll do the Featured / FL / DU block the way the outcome of discord is.
If the outcome does not expect your wishes, it's a simple task to adjust it in your own fork.

A new PR for the topic "re-activate FL/DU after featured period has ended" would be great for all tho!

K4Li:
> @K4Li Talking on discord about this right now. I'll do the Featured / FL / DU block the way the outcome of discord is. If the outcome does not expect your wishes, it's a simple task to adjust it in your own fork.
> 
> A new PR for the topic "re-activate FL/DU after featured period has ended" would be great for all tho!

Yes, I agree. MOST aren't going to change "Featured" so it is best to keep it the way it is. It is easy enough to remove featured from the "elseif" Thanks again for the PR!

kis2a:
add this in upload page and sitpot ;)

Obi-Wana:
PR will be refactored with other new features before merged.
ToDo:
- Move all BUFF features to a new Controller
- % FL slider

Obi-Wana:
Going to close this one and open a new one.

PR #2042, opened 2021-12-07
(Fix) 500 Server Error if User Warning is Soft Deleted

Alkl58:
Issue introduced in 33f49016e7c4257ac2abb92c1f6fd3a36016808b (security: cross-site request forgery)


HDVinnie:
Good catch! Thanks.

PR #2048, opened 2021-12-14
(Update) Move Hardcoded Controller sentences to Lang files & German / French Translations 

Alkl58:


Alkl58:
Re-Checked and no technical issues found.

PR #2049, opened 2021-12-14
(Add) Torrent Client Stats

Alkl58:
Would recommend to test it... 

Not gonna lie, the job / command `AutoStatsClients.php` has pretty ugly looking code, but it works on prod.

A job to calculate the clients and users runs every 24h:
```php
$schedule->command('auto:stats_clients')->daily();
```

A new route has been added:
```php
Route::get('/user/clients', [App\Http\Controllers\StatsController::class, 'clients'])->name('clients');
```

Stats read from cache (see `StatsController.php`):
```php
$clients = [];

if (\cache()->has('stats:clients')) {
     $clients = \cache()->get('stats:clients');
}

return \view('stats.clients.clients', ['clients' => $clients]);
```

On stats page:

grafik

When there is nothing in the cache / not run yet:

grafik

Working behavior:

grafik

### DIFF
--- a/resources/views/stats/clients/clients.blade.php
+++ b/resources/views/stats/clients/clients.blade.php
@@ -24,6 +24,7 @@
             <hr>
             <div class="row col-md-offset-1">
                 @foreach ($clients as $key => $value)
+                    @php if (\strlen($key) > 26) { $key = substr($key, 0, 23).'...'; } @endphp
                     <div class="well col-md-3" style="margin: 10px;">
                         <div class="text-center">
                             <h3 class="text-success">{{ $key }}</h3>
--- a/resources/views/stats/clients/clients.blade.php
+++ b/resources/views/stats/clients/clients.blade.php
@@ -1,0 +1,37 @@
+@extends('layout.default')
+
+@section('title')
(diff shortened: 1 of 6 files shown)


resources/views/stats/clients/clients.blade.php
```diff
@@ -0,0 +1,37 @@
...
+    </li>
+@endsection
+
+@section('content')
+    <div class="container">
+        <div class="block">
+            <h2>Clients</h2>
+            <hr>
+            <div class="row col-md-offset-1">
+                @foreach ($clients as $key => $value)
```
HDVinnie, line 26: maybe should add some default sorting for agent alphabetically?

Alkl58: We could put sorting in `StatsController.php` to keep the blade file clean.

It seems, that PHP has a neat sorter for sorting by keys in an array: https://www.php.net/manual/en/function.ksort.php

However I am thinking about adding a string length limiter in the blade file, to avoid too long Agent strings, which might result in poor looking boxes.

HDVinnie: sounds good to me

PR #2051, opened 2021-12-17
(Chore) Reformat Views

HDVinnie:
- @clandestine8 @Obi-Wana I think my IDE settings were borked.....this should be proper now if want to review before merge.


Obi-Wana:
Took a quick look and looks much better now.

PR #2074, opened 2022-01-10
(Refactor) Warnings System

HDVinnie:
- To be explained


costaht:
Hi guys, I know I'm very late but, I updated from 6.3 to 8.2 recently and I just realized that the HnR users are only losing their download privileges and not being banned. 
So my question is: will a user ever be banned by reaching the maximum number of warnings or all bans became manual tasks from this PR on?

Thanks

PR #2078, opened 2022-01-18
(Add) Limit BBCode Font size in User Profile ("About Me")

Alkl58:
This limits the maximum BBCode font size to 99 with simple regex.

Why this addition? : Because there are very "interesting" users which will set it to ridiculous values.

Maybe this can be done better, but it's enough at first.


HDVinnie:
LGTM

kiam001:
Same problem in the shoutbox

PR #2174, opened 2022-03-28
Patch 1

innocente205:
Correct db count and remove extra lookup

### DIFF
--- a/app/Http/Controllers/StatsController.php
+++ b/app/Http/Controllers/StatsController.php
@@ -187,7 +187,7 @@ class StatsController extends Controller
     public function seeders(): \Illuminate\Contracts\View\Factory|\Illuminate\View\View
     {
         // Fetch Top Seeders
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/StatsController.php
```diff
@@ -187,7 +187,7 @@ public function downloaded(): \Illuminate\Contracts\View\Factory|\Illuminate\Vie
     public function seeders(): \Illuminate\Contracts\View\Factory|\Illuminate\View\View
     {
         // Fetch Top Seeders
-        $seeders = Peer::with('user')->select(DB::raw('user_id, count(*) as value'))->where('seeder', '=', 1)->groupBy('user_id')->latest('value')->take(100)->get();
+        $seeders = Peer::with('user')->select(DB::raw('user_id, count(distinct info_hash) as value'))->where('seeder', '=', 1)->groupBy('user_id')->latest('value')->take(100)->get();
```
HDVinnie, line 190: Would be better off using `count(distinct torrent_id)` instead of string `count(distinct info_hash)`

PR #2233, opened 2022-05-18
Improve & Repair Announce Routes

clandestine8:
- Move Announce to its own Middleware Group
- Add Sensible Throttling to prevent overloading servers
- Fix Min Interval Check
- Fix Connectable Check


HDVinnie (approved):
LGTM

PR #2235, opened 2022-05-18
Fix: Cache Key Prefix

clandestine8:
I was building the cache key like this 

`$key = strtolower(config('app.name')).'_cache:peers:connectable:'.$tmp_ip.'-'.$this->port.'-'.$this->agent;`

I should have been building it like this 

`$key = config('cache.prefix').':peers:connectable:'.$tmp_ip.'-'.$this->port.'-'.$this->agent;`

UNIT3D was building the prefix like this

` 'prefix' => env('CACHE_PREFIX', Str::slug(env('APP_NAME', 'laravel'), '_').'_cache'),` 

But built using an ENV which is not used by default so this has also be fixed. This would be an issue in a multi-tenant environment


clandestine8:
Wrong branch

PR #2243, opened 2022-05-29
(Fix) Bug in Upload Form Logic

clandestine8:
Upload form was not properly revealing all attributes and did not set default attributes in cases where the data was not available from TMDB (such as TVDM id missing). This would result in the form being unresponsive as the error was also hidden from users. There are other general fixes and improved controller handling in the form as well.

Should be merged to Master and 6.x.x as the bug is being experiences with end users in the wild. 

Please review.


HDVinnie (approved):
LGTM

PR #2246, opened 2022-05-31
(Fix) top navigation, navigation tabs, and breadcrumbs

Roardom:
- Various improvements to the top navigation bar accessibility for different screen sizes and users with long-numbered stats and longer than average usernames.
- Changed header to be sticky instead of the top navigation bar and secondary navigation bar being individually fixed. Also fixes the issue where breadcrumbs are overlap the page content if using mobile layout.
- Fixes a few incorrectly-assigned classes.
- Changes the breadcrumbs to an ordered list, and the navigation tabs to an unordered list.
- Fixes breadcrumbs being slightly off-center.
- Fixes not being able to scroll inside the hamburger menu on mobile.

### DIFF
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -19,5 +19,6 @@
 @import 'components/quick_search';
 @import 'components/user-active';
 @import 'components/user-torrents';
+@import 'layout/header';
 @import 'layout/secondary-nav';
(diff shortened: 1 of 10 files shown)


clandestine8:
> LGTM
> 
> @clandestine8 any input before merge?

LGTM2

PR #2249, opened 2022-05-31
(Fix) torrent filter type hints

Roardom:
Adds proper type hinting to the torrent filter trait.


clandestine8 (approved):
Everything seems consistent. Good to go

PR #2331, opened 2022-08-09
(Update) optimize highspeed tag

Roardom:
Something like this should hopefully improve performance. From my testing, it works correctly, however it requires further testing to verify that the query run time compared to the previous version with a significant amount of seedboxes and peers is indeed more performant. To note: the previous query wasn't actually correct as it tagged all torrents that a user with a seedbox was seeding, regardless if it was actually the ip of the seedbox that was seeding the torrent.


HDVinnie:
Tested and more performant.

PR #2334, opened 2022-08-09
(Add) Bon store product edit

Roardom:
Provides the ability for staff to edit bon store items and their price.


HDVinnie:
Maybe this should be limited to admin + only?

PR #2373, opened 2022-08-27
(Fix) Peer progress bar

Roardom:
Currently, the progress bar on the peers percentage complete page uses an algorithm that will provide the wrong value if a user resets their torrent session (i.e. restarts their client). Calculating the progress based on `left` is more accurate in cases like these.

Progress is also rounded differently at either end of 0 and 100 so that a user who downloads 1 byte is considered at 1% and a user who has 1 byte remaining is considered at 99%. This helps in cases where a peer is missing a single piece and they won't show as having (a rounded) 100%.

### DIFF
--- a/resources/views/torrent/peers.blade.php
+++ b/resources/views/torrent/peers.blade.php
@@ -83,29 +83,17 @@
                                                 {{ $p->user->username }}</span></a>
                                 </td>
                             @endif
-                            @if ($p->seeder == 0)
-                                <td>
-                                    <div class="progress">
-                                        <div class="progress-bar progress-bar-striped active" role="progressbar"
-                                             aria-valuenow="{{ ($p->downloaded / $torrent->size) * 100 }}"
-                                             aria-valuemin="0"
-                                             aria-valuemax="100"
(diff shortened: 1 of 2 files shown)


resources/views/torrent/peers.blade.php
```diff
@@ -83,29 +83,25 @@ class="{{ $p->user->group->icon }}" data-toggle="tooltip"
...
-                                </td>
-                            @elseif ($p->seeder == 1)
-                                <td>
-                                    <div class="progress">
-                                        <div class="progress-bar progress-bar-striped active" role="progressbar"
-                                             aria-valuenow="100" aria-valuemin="0" aria-valuemax="100"
-                                             style="width: 100%;">
-                                            100%
-                                        </div>
+                            @php
```
HDVinnie, line 86: I would avoid using @php @endphp. I know it used in a few places in UNIT3D already but it is considered bad practice and they will be removed eventually so dont really want to add more.

Roardom: How about moving it into the controller?

PR #2416, opened 2022-09-30
(Add) Cheating detection

Roardom:
Records a "balance" of every torrent which is calculated from the sum of the upload minus the sum of the download.

Adds a staff page for viewing the most "cheated on" torrents, sorted by the proportion of the torrent's filesize that was cheated on.

### DIFF
--- a/database/migrations/2022_09_29_182332_alter_torrents_table.php
+++ b/database/migrations/2022_09_29_182332_alter_torrents_table.php
@@ -1,0 +1,21 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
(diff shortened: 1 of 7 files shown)


database/migrations/2022_09_29_182332_alter_torrents_table.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('torrents', function (Blueprint $table) {
+            $table->bigInteger('balance');
```
HDVinnie, line 17: These should probably be nullable

PR #2448, opened 2022-11-03
(Refactor) Torrent moderation buttons

Roardom:
Changes the moderation buttons to be cruddy. Also checks to see if the torrent's moderation status has changed since the page was loaded, and errors if so.


HDVinnie:
nice one

PR #2411, opened 2022-09-20
(Update) Make bonus point allocation more customizable

Roardom:
WIP

Todo:
- Figure out bon allocation query
- Update user earnings page to display new earnings
- Change multiplier column to accept decimals

image

PR #2453, opened 2022-11-10
(Add) Freeleech based on size threshold

LostRager:
Adds the ability to activate or deactivate automatic freeleech on torrents uploaded that meet a certain threshold set by the owner.

### DIFF
--- a/app/Http/Controllers/API/TorrentController.php
+++ b/app/Http/Controllers/API/TorrentController.php
@@ -207,6 +207,10 @@ class TorrentController extends BaseController
 
             return $this->sendError('Validation Error.', $v->errors());
         }
+        
+        if (config('torrent.size_freeleech') == true && $torrent->size >= \config('torrent.size_threshold')){
+            $torrent->free = 100;
+        }
 
         // Save The Torrent
         $torrent->save();
--- a/app/Http/Controllers/TorrentController.php
+++ b/app/Http/Controllers/TorrentController.php
@@ -555,6 +555,10 @@ class TorrentController extends Controller
             return \to_route('upload_form', ['category_id' => $category->id])
                 ->withErrors($v->errors())->withInput();
         }
+        
+        if (config('torrent.size_freeleech') == true && $torrent->size >= \config('torrent.size_threshold')){
+            $torrent->free = 100;
+        }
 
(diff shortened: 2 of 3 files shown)


HDVinnie:
I think this would be better off as a dashboard option where you can set conditions for FL or DU on torrents automatically. Using this as a base would be great. https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/2411

@Roardom thoughts?

Obi-Wana:
How about having it like this in a fist step and refactor it to a dashboard later?

HDVinnie:
I honestly feel like this is what we're trying to avoid anymore. Im not trying to push out contributors as there very welcome and much needed but in this case there is a more flexable and complete way to build this system out. If we merge as is then its just something we will have to revisit again and do a complete refactor....

Obi-Wana:
You got an argument with this, indeed. Rgr, then lets touch it once and do it the clean way.

FuNFooKaS:
could this be fixed up for V7?

LostRager:
If you want it to work exactly how its laid out in the pr.

You can f.ex add this here:

```
if (config('torrent.size_freeleech') == true && $torrent->size >= \config('torrent.size_threshold')){
    $torrent->free = 100;
    $torrent->save();
}
```

https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/6ed14c67142092a7a7aafbc4b680e89e776550be/app/Http/Controllers/TorrentController.php#L377

and this f.ex. here:

    if (config('torrent.size_freeleech') == true && $torrent->size >= \config('torrent.size_threshold')){
        $torrent->free = 100;
    }

https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/6ed14c67142092a7a7aafbc4b680e89e776550be/app/Http/Controllers/API/TorrentController.php#L244

FuNFooKaS:
This is not working in  latest unit3d?

LostRager:
> This is not working in latest unit3d?

This is not needed anymore. https://unit3d.dev/dashboard/automatic-torrent-freeleeches

FuNFooKaS:
I liked it, i for the life of me can not seem to get a proper regex to take in the new one.

PR #2493, opened 2022-12-22
(Update) Peers table

Roardom:
Size reductions:

- Converts `peer_id` from utf-encoded hex to binary, removing 20 bytes per peer.
- Removes `md5_peer_id` which isn't used, removing 32 bytes per peer.
- Removes `info_hash` which isn't used (uses `torrent_id` relation instead), removing 40 bytes per peer.
- Removes `connectable` which isn't used (used in redis instead), removing 1 byte per peer.
- Changes `ip` from utf-encoded dot notation to binary, saving 7 bytes per peer for ipv4, much more for ipv6.
- No null values anymore, removing at least 1 byte per peer.

Total savings: 100 bytes per peer. At 1 million peers, 100MB is saved. Assuming a common agent size of 16 characters, the size of the table is reduced by just over half.

Even though 4 billion peers are very unlikely, mysql does not recycle ids, and each new `started` event will increase the id, which significantly increases the chance of running out of ids, so it was kept as an 8-byte int.

Took advantage of requiring the usage of `INET6_NTOA` in all the controllers/views and optimized the clients table on the profile page to use `group by` instead of looping over all the peers.

### DIFF
--- a/app/Console/Commands/AutoHighspeedTag.php
+++ b/app/Console/Commands/AutoHighspeedTag.php
@@ -49,7 +49,7 @@ class AutoHighspeedTag extends Command
             ->leftJoinSub(
                 Peer::distinct()
                     ->select('torrent_id')
-                    ->whereRaw("ip IN ('".$seedboxIps->implode("','")."')"),
+                    ->whereRaw("INET6_NTOA(ip) IN ('".$seedboxIps->implode("','")."')"),
                 'highspeed_torrents',
                 fn ($join) => $join->on('torrents.id', '=', 'highspeed_torrents.torrent_id')
             )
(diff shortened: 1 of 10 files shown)


app/Console/Commands/AutoHighspeedTag.php
```diff
@@ -49,7 +49,7 @@ public function handle(): void
             ->leftJoinSub(
                 Peer::distinct()
                     ->select('torrent_id')
-                    ->whereRaw("ip IN ('".$seedboxIps->implode("','")."')"),
+                    ->whereRaw("INET_NTOA6(ip) IN ('".$seedboxIps->implode("','")."')"),
```
HDVinnie, line 52: Shouldn't this be `INET6_NTOA(ip)`.

Roardom: Indeed. Good eye. Changed now.

PR #2520, opened 2023-01-02
(Update) User invites views

Roardom:
Update view to new css design

### DIFF
--- a/resources/views/user/invite/index.blade.php
+++ b/resources/views/user/invite/index.blade.php
@@ -19,96 +19,89 @@
     @include('user.buttons.user')
 @endsection
 
-@section('content')
-    <div class="container-fluid">
-        <div class="block">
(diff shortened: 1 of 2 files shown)


resources/views/user/invite/index.blade.php
```diff
@@ -19,96 +19,90 @@
...
+                            <td>
+                                <menu class="data-table__actions">
+                                    <li class="data-table__action">
+                                        <form
+                                            action="{{ route('invites.send', ['id' => $invite->id]) }}"
+                                            method="POST"
+                                            x-data
+                                        >
+                                            @csrf
+                                            @method('DELETE')
```
HDVinnie, line 75: This added by mistake?

Roardom: Indeed. Fixed now.

PR #2553, opened 2023-01-21
(Update) Support for new mediainfo languages in parser

milkdawg:
Goal of this it re-add country flags to subtitle portion of mediainfo pane/parser with language output changes made in mediainfo v22.12.  Overall pretty simple changes but relevant info is 

- Not comprehensive,  but should cover most bases and only did real world examples I could find instead of guessing what mediainfo output would be.  If this gets merged will keep a list of ones I see and will come back for another round.
- Yes Chinese and Mandarin are both valid, and (Simplied) is a typo on mediainfo's side, not mine
- Set latin american spanish to mexico as larged spanish speaking country in world.
- Re-added canadian flag as can.png.  If you want a different filename I understand but ca, cn, cd are taken, this was the exact flag that was in repo prior
- Yes default country/variant for Arabic is 001
- Mapped Bengali to Bengladesh as I saw it, was already here making changes and flag was in repo

There's potential to re implement function to do partial matches, but wasn't going to go down that route unless specifically requested


HDVinnie:
Thanks for the contribution.

PR #2569, opened 2023-01-31
(Update) Add more mediainfo languages to flag mappings

milkdawg:
Saw prep work start for next release so figured I'd get this in as I haven't seen anything not covered yet in the last week or so outside of these.

- Maps `fil (PH)` to Philippines flag
- Maps `English (CA)` to US flag.  Figured better to map this to US instead of having both French and English point to Canadian flag.  If you disagree I'm more than willing to fix.
- Replaced some tabs with spaces for more consistent viewing

I think failing unit test is due to me originally setting base branch to `7.x.x`, this helpers file is nowhere in stack trace.


Roardom:
> Maps English (CA) to US flag. Figured better to map this to US instead of having both French and English point to Canadian flag. If you disagree I'm more than willing to fix.

I'd almost say Canadian English is more similar to British English. Perhaps you can use the Quebec flag for French Canadian. Even though French is spoken a lot in other places in Canada other than just Quebec, imo it makes more sense the other way around. Unfortunately Canada doesn't have a "French Canadian" flag that covers the whole country, only each province has their own separate English and French flags.

milkdawg:
No problem.  
- added Quebec flag with a width of 100px, to line up with other flags in relevant directory, did not see it there already
- Mapped `English (CA)` to generic canadian flag
- Mapped `French (CA)` to quebec flag

PR #2590, opened 2023-02-08
(Update) Rendered bbcode/markdown/html styles

Roardom:
Updated the styles used for rendered bbcode, markdown and html. Base styles inspiration sourced from https://github.com/sindresorhus/github-markdown-css/blob/main/github-markdown.css, although removed all unnecessary styles we don't support and changed it around to support our bbcode.


resources/sass/components/_bbcode-rendered.scss
```diff
@@ -0,0 +1,616 @@
+.bbcode-rendered.bbcode-rendered {
```
Oha-you, line 1: Just wondering, was there a reason to increase specificity by repeating `.bbcode-rendered.bbcode-rendered` here?

Roardom: Yes. If I recall correctly, it was the existing panel styles, or another location the component was placed into that inherited styles that needed to be overridden.

PR #2629, opened 2023-03-06
(Update) new credit system and meta info redesign

Roardom:
Better handling of "People". Only save specific occupations of people instead of everything tmdb returns.

Also refactors the movie/tv/game metadata on the torrent/request/similar page to a new design.

Also refactors the torrent and similar pages to the new sass design, including a more sleek mediainfo panel.

TODO:
- mediahub.person.show needs an update so to filter movies/tv by occupation.


Roardom:
This PR looks like it will also fix #2633

Roardom:
Also fixes #2637

PR #2679, opened 2023-04-09
(Update) Optimize api

Roardom:
- Remove unnecessary relation fetching (comments count and thanks count) since they're not returned by the api.
- Use cursor pagination for higher efficiency (Requires users parsing the `links.next` key for finding the link to the next page).
- Cache results for 5 minutes

Reasoning:
- Jackett calls currently use 9% total database execution time.
- The pagination in the api currently uses an additional 23% database execution time.
- Extra discussion in #2322

### DIFF
--- a/app/Http/Controllers/API/TorrentController.php
+++ b/app/Http/Controllers/API/TorrentController.php
@@ -343,41 +343,59 @@ class TorrentController extends BaseController
             && $field[-1] === '/'
             && @preg_match($field, 'Validate regex') !== false;
 
-        $torrents = Torrent::with(['user:id,username,group_id', 'category', 'type', 'resolution'])
-            ->when($request->filled('name'), fn ($query) => $query->ofName($request->name, $isRegex($request->name)))
-            ->when($request->filled('description'), fn ($query) => $query->ofDescription($request->description, $isRegex($request->description)))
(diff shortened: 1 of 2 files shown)


Roardom:
Update: This change has resulted in jackett from using 32% of the total database execution time, to only 4.3%

Roardom:
> > Update: This change has resulted in jackett from using 32% of the total database execution time, to only 4.3%
> 
> Fun fact... by removing [meta.total](http://meta.total) you broke every existing Jackett installation, not because of the optimizations.

From my understanding in https://github.com/HDInnovations/UNIT3D-Community-Edition/discussions/2322#discussioncomment-5566593, the 2 lines (https://github.com/Jackett/Jackett/blob/adb9b97ecffd018fb4095add09599c717306fc78/src/Jackett.Common/Definitions/blutopia-api.yml#L95-L96) simply have to be removed from the definition. We even tested this ourselves and it worked fine once those two lines were removed. Existing installations can adapt to the new change over time.

Roardom:
Indeed we notice less queries than we had before, but only about half as much, while using a much lower load.

Also note that this is the development branch and as far as I know, only a single tracker uses it in production. Things can still be changed before merged into the stable branch if absolutely necessary.

PR #2700, opened 2023-04-16
(Update) Optimize navbar

Roardom:
Cache the seeding and leeching count of users for 60 seconds. Also, use a less resources intensive query to calculate it.

### DIFF
--- a/resources/views/partials/top_nav.blade.php
+++ b/resources/views/partials/top_nav.blade.php
@@ -252,13 +252,13 @@
             <li class="ratio-bar__seeding" title="{{ __('torrent.seeding') }}">
                 <a href="{{ route('users.peers.index', ['user' => auth()->user()]) }}">
                     <i class="{{ config('other.font-awesome') }} fa-upload"></i>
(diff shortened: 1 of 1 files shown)


JoshyPHP:
@Roardom Have you considered using a single query for both counts? If I read the schema correctly, there's an index on `user_id` but none on `seeder`, which means that the request is not satisfiable using only the index and the rows have to be read. If you query both the number of rows where `user_id = ?` and the value of `SUM(seeder)` for those rows, you should be calculate both values at once. Does that make sense?

Roardom:
> If you query both the number of rows where user_id = ? and the value of SUM(seeder) for those rows, you should be calculate both values at once

Thanks for the feedback. Unfortunately, doing it that way wouldn't allow a `distinct` without using a subquery, but the performance might be worth it over the inaccuracy. I'll see what I can do.

JoshyPHP:
Oh right, I didn't notice the `distinct()` in that query so I had not realized that a user could have several `peers` rows for a given torrent.

PR #2723, opened 2023-04-25
(Update) Livewire comments

HDVinnie:
- add back chat messages

### DIFF
--- a/app/Http/Livewire/Comment.php
+++ b/app/Http/Livewire/Comment.php
@@ -28,12 +28,15 @@ use App\Achievements\UserMadeTenComments;
 use App\Models\User;
 use App\Notifications\NewComment;
 use App\Notifications\NewCommentTag;
+use App\Repositories\ChatRepository;
 use Illuminate\Support\Facades\Notification;
 use Livewire\Component;
 use voku\helper\AntiXSS;
 
 class Comment extends Component
 {
+    protected ChatRepository $chatRepository;
+
     public $comment;
 
     public $anon = false;
@@ -59,6 +62,11 @@ class Comment extends Component
     public $editState = [
         'content' => '',
     ];
+
+    final public function boot(ChatRepository $chatRepository): void
+    {
+        $this->chatRepository = $chatRepository;
+    }
 
     final public function mount(): void
     {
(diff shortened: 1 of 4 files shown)


Roardom (commented):
- Collection comments error because of lack of user_id property
- Torrent request results in "x has left a comment on torrentrequest y" without a space.
- Replies to parent comments aren't considered.
- Ticket comments need to be filtered out.

HDVinnie:
> * Collection comments error because of lack of user_id property
> * Torrent request results in "x has left a comment on torrentrequest y" without a space.
> * Replies to parent comments aren't considered.
> * Ticket comments need to be filtered out.

All sorted besides `torrentrequest`. Not sure best way to handle that one.

Roardom:
> All sorted besides `torrentrequest`. Not sure best way to handle that one.

Maybe Str::snake() and replace underscores with spaces?

app/Http/Livewire/Comment.php
```diff
@@ -142,14 +150,46 @@ final public function postReply(): void
...
+
+        // Auto Shout
+        $profileUrl = href_profile($this->user);
+
+        $modelUrl = match (strtolower(class_basename($this->comment->commentable_type))) {
+            'article'    => href_article($this->comment->commentable_type),
+            'collection' => href_collection($this->comment->commentable_type),
+            'playlist'   => href_playlist($this->comment->commentable_type),
+            'request'    => href_request($this->comment->commentable_type),
+            'torrent'    => href_torrent($this->comment->commentable_type),
```
Roardom, line 169: `->commentable_id` instead of `->commentable_type`

app/Http/Livewire/Comment.php
```diff
@@ -142,14 +150,46 @@ final public function postReply(): void
...
+            default      => "#"
+        };
+
+        if (strtolower(class_basename($this->comment->commentable_type)) !== 'ticket') {
```
Roardom, line 173: Probably should add code so that if a user replies to their own parent comment in a ticket, the staff member assigned the ticket still gets notified.

Roardom (approved):

PR #2747, opened 2023-05-01
(Bug) Fix torrent free removal

Obi-Wana:
Since the view for Freeleech (option "No" equals value 0) uses 0, this must be a valid option in the TorrentBuffController.


Roardom:
Thanks, been meaning to get this one for awhile now. However, the rest of the function needs to be updated as well. Since the if statement relies on the value never being 0, there are cases where the chat message will be sent incorrectly.

PR #2748, opened 2023-05-01
(Bug) Fix Usericon spacing in forum posts

Obi-Wana:
This will ensure that username icons, such as isOnline & personal message, are the same style as on user profile blades. Also, fixed the spacing and padding on forum posts for those two icons.

### DIFF
--- a/resources/views/components/forum/post.blade.php
+++ b/resources/views/components/forum/post.blade.php
@@ -133,7 +133,7 @@
             :anon="false"
             :user="$post->user"
         >
-            <x-slot:appended-icons>
+            <x-slot:appendedIcons>
                 @if ($post->user->isOnline())
                     <i class="{{ config('other.font-awesome') }} fa-circle text-green" title="Online"></i>
                 @else


resources/views/components/forum/post.blade.php
```diff
@@ -133,7 +133,7 @@ class="post__author"
             :anon="false"
             :user="$post->user"
         >
-            <x-slot:appended-icons>
+            <x-slot:appendedIcons>
```
Roardom, line 136: The documentation states that kebab case should be used in html when referring to component properties: https://laravel.com/docs/10.x/blade#casing

Roardom (commented):
I'm unable to reproduce the bug you describe on a fresh instance of the 7.x.x branch, and the spacing remains the same after the change.

7.x.x:
image


This PR:
image

Obi-Wana:
@Roardom

I saw that "bug" on BLU, where there was no spacing at all in between username, isOnline and the envelope. However, as I checked back today, the bug does not occur anymore. Don't know if something has changed or not.
Since you could not reproduce it on a new install, feel free to close.

------- Original Message -------

Roardom:
Thanks for the PR anyhow, I think the changes made in this commit were the true fix to the issue: https://github.com/HDInnovations/UNIT3D-Community-Edition/commit/6e64801e8809edd6dcb82ca514f390141929d925#diff-0d6724e04169318c104612caed708ac22b506b5d6b1a69a7d7d29ba97ca2dc2c

PR #2799, opened 2023-05-22
(Fix) High speed torrents

HDVinnie:


app/Console/Commands/AutoHighspeedTag.php
```diff
@@ -43,7 +43,10 @@ class AutoHighspeedTag extends Command
      */
     public function handle(): void
     {
-        $seedboxIps = Seedbox::pluck('ip')->filter(fn ($ip) => filter_var($ip, FILTER_VALIDATE_IP));
+        $seedboxIps = Seedbox::all()
+            ->where('seeder', '=', 1)
```
Roardom, line 47: `->where('seeder', '=', 1)` has to go down below on the peers subquery instead of here.

Roardom (changes requested):
Oops

Roardom (approved):

PR #2811, opened 2023-06-03
(Fix) Tickets Search

HDVinnie:


Roardom (commented):
Should probably just use the third parameter of `->when()` as `else` instead of handling opposite cases. Probably the same thing for the non-modo case as well.

PR #2815, opened 2023-06-03
(Update) Password rules

HDVinnie:
- `Password::min(12)->mixedCase()->letters()->numbers()->uncompromised()` is now used across UNIT3D. Previously was only used in User Password Controller.


Roardom (changes requested):
ResetPasswordController should probably be updated too, although not sure what's happening in there. It looks like it's skipping a few things.

HDVinnie:
closing in favor of #2831

PR #2820, opened 2023-06-13
(Add) Invite retraction

Roardom:
Allow users and staff to retract unused invites.


Roardom:
These should use some sort of soft deletion instead, but still kept and not permanently deleted.

PR #2829, opened 2023-06-17
(Fix) Migration models

Roardom:
Should never have used eloquent models in migrations. Also fix up a few other bad migrations in the past if people haven't migrated yet. All torrent_id possible values were triple checked but user_id values still need to be checked to make sure nothing is deleted accidentally.


Roardom:
The following tables should not have been migrated in this fashion for the alter_user_id migration:

- `bon_transactions`:  `sender`, `receiver` (0 has a special meaning)
- `reports`: `reported_user` (0 has a special meaning)

Tables of records that probably should be deleted instead of updated to system user if the user itself doesn't exist anymore:

- `achievement_progress`
- `bookmarks`
- `clients`,
- `failed_login_attempts`,
- `freeleech_tokens`,
- `graveyard`,
- `likes`
- `notifications`,
- `peers`,
- `personal_freeleech`
- `rss`: private only
- `subscriptions`,
- `thanks`,
- `twostep_auth`,
- `user_activations`
- `user_audibles`,
- `user_echoes`,
- `user_notes`,
- `user_notifications`,
- `user_privacy`,
- `warnings`: `user_id` only
- `watchlist`: `user_id` only,
- `wishes`,

database/migrations/2022_11_24_032521_update_requests_table.php
```diff
@@ -11,28 +11,28 @@
      */
     public function up(): void
     {
-        TorrentRequest::query()
+        DB::table('requests')::query()
```
HDVinnie, line 14: cannot use query here. same with line 21, 28, 35

database/migrations/2022_12_23_103322_update_requests_table.php
```diff
@@ -18,7 +17,7 @@ public function up(): void
             $table->foreign('torrent_id')->references('id')->on('torrents')->cascadeOnUpdate()->cascadeOnDelete();
         });
 
-        TorrentRequest::query()
+        DB::table('requests')->query()
```
HDVinnie, line 20: cannot use query here.

PR #2837, opened 2023-06-18
(Add) Torrent activity icons

HDVinnie:


espiman:
shouldn't history as not_completed be seeder=0?

Thanks

PR #2852, opened 2023-07-03
(Fix) Safe defaults for forum permissions on group creation

Roardom:
Too often a new group is created and forum permissions are forgotten to be updated.


HDVinnie:
good call

PR #2910, opened 2023-07-18
(Add) Redis history batching

Roardom:
Although now currently functional, #2565 unfortunately doesn't offer any performance improvements. This is most likely because `history` seems to be the main bottleneck, not `peers`. This PR follows the same code structure, but for the history records instead. It also uses the new Laravel feature to have commands be ran more often than once per minute.


HDVinnie:
NOTE: This should not be used. 7.x.x is not released yet so those running it in beta beware this PR while merged still needs work!

PR #2965, opened 2023-07-27
(Fix) Request rejection rejector fetching

Roardom:
- `->username` isn't a property of `->user()` (relation rather than object)
- Changed the variable names to be more clear.
- Fetch the relations before clearing the fields needed to fetch the relation otherwise the notification will have null as the sender

### DIFF
--- a/app/Http/Controllers/RequestFillController.php
+++ b/app/Http/Controllers/RequestFillController.php
@@ -64,17 +64,18 @@ class RequestFillController extends Controller
     {
         abort_unless($request->user()->id === $torrentRequest->user_id || $request->user()->group->is_modo, 403);
 
+        $filler = $torrentRequest->filler;
+        $requester = $torrentRequest->user;
+        $approver = $request->user();
+
         $torrentRequest->update([
             'filled_by'   => null,
             'filled_when' => null,
             'torrent_id'  => null,
         ]);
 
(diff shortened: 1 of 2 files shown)


Oha-you:
Note: same changes are needed for [ApprovedRequestFillController.php#L85](https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/master/app/Http/Controllers/ApprovedRequestFillController.php#L85), otherwise request creator's name is revealed to the filler when the creator approves a fill.

Sorry, I'm too lazy to create a proper issue.

Roardom:
> Note: same changes are needed for [ApprovedRequestFillController.php#L85](https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/master/app/Http/Controllers/ApprovedRequestFillController.php#L85), otherwise request creator's name is revealed to the filler when the creator approves a fill.
> 
> Sorry, I'm too lazy to create a proper issue.

The line you reference creates a new NewRequestFillApprove notification, which only says `anonymous` for the request creator's name if the torrent request is anon:

https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/c904aea8beac627666c8c9e120672a4987b6d5a1/app/Notifications/NewRequestFillApprove.php#L58-L59

I also tested by creating a new anonymous request with one account, filling it using another account without anonymous ticked, and the filler received a notification saying "An anonymous user has approved your fill of Requested Torrent test".

PR #3040, opened 2023-08-08
(Update) Inline announce job

Roardom:
This prevents duplicate queries from being made since laravel jobs by default refetch Models from the database if they are an argument. The main issue with doing this before was that the Peer/History inserts/updates are expensive, but we now only do an O(1) insert into a redis queue and upsert them all at once later. User and Torrent updates are occasional enough that they shouldn't be an issue, and PersonalFreeleech and FreeleechToken queries should have small enough tables and be cached fast enough that there shouldn't be any issues either.


Roardom:
This current iteration won't work. The delay for the port connectivity check has to be async from the announce otherwise the 1 second delay will bring everything offline. Options include moving the port connectivity check to an hourly scheduled job, or to maintain the job but serialize the models as primitive arrays.

Roardom:
Peer connectivity checks are now moved to an hourly scheduled job: #3113

PR #3056, opened 2023-08-14
(Update) Use foreign key constraints for forum and topics ids

Roardom:
Also rely on parent_id = null, instead of parent_id = 0 to determine if the forum is a category.


HDVinnie:
Tests are failing

Roardom:
> Tests are failing

Fixed.

PR #3069, opened 2023-08-18
(Add) Search by torrent size

Roardom:
The use case for this filter is ordering grouped torrents by times_completed, and wanting to filter out low-size torrents since they're often bon farmed and skew the numbers.


shapelessvoid (approved):
LGTM

PR #3071, opened 2023-08-19
(Add) Torrent posters search view

Roardom:
Allow only seeing tmdb posters when browsing torrents.


HDVinnie:
Looks to be some larastan issue to sort before merge

PR #3079, opened 2023-08-23
(Update) Split torrent tmdb into movie_id and tv_id

Roardom:
This should allow further improvements such as eager loading movies and tv using the framework's `with` function instead of doing it manually every time.

Results in grouped torrent search being approximately ~20% faster.


Roardom:
The code has changed too much to bother rebasing. It would be easier starting from scratch.

PR #3107, opened 2023-09-01
fix: avatar remove old

### LINKED ISSUE #2482: [Bug] Old profile picture is saved if new profile picture is a different format
### UNIT3D Version

6.03

### PHP & Platform

8.1.2 - Ubuntu 22.04 x64

### Laravel verion

9.2.3

### Have you done this?

### Expectation

When uploading a new profile picture of a different media format, the old profile picture should no longer exist.

### Description

When uploading a profile picture with a different media format, the old profile photo is not removed from storage.

### Stack trace & logs

_No response_

### Are you a UNIT3D sponsor?

No, don't give priority to this

ReileenKawahara:
Removes avatar's old file format when need be.
Fixes #2482

### DIFF
--- a/app/Http/Controllers/User/UserController.php
+++ b/app/Http/Controllers/User/UserController.php
@@ -157,6 +157,7 @@ class UserController extends Controller
             }
 
             $avatar = $user->username.'.'.$image->getClientOriginalExtension();
+
             if ($user->image != $avatar) {
                 $oldAvatar = $user->image;
                 $user->image = $avatar;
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/User/UserController.php
```diff
@@ -155,7 +156,12 @@ public function update(Request $request, User $user): \Illuminate\Http\RedirectR
                 $image->move(public_path('/files/img/'), $filename);
             }
 
-            $user->image = $user->username.'.'.$image->getClientOriginalExtension();
+            $avatar = $user->username.'.'.$image->getClientOriginalExtension();
+
+            if ($user->image != $avatar) {
```
Roardom, line 161: I believe you should be able to safely use `!==` here. Correct me if I'm wrong.

ReileenKawahara: You're right, thanks for pointing it out.

Roardom (approved):

PR #3115, opened 2023-09-04
(Fix) Tmdb service types

Roardom:
Haven't tested yet.


Roardom:
Tested now, seems to work, however still no fail safe for when tmdb is down.

PR #3136, opened 2023-09-11
(Add) Ip blocking system

HDVinnie:


database/migrations/2023_09_10_234654_create_blocked_ips_table.php
```diff
@@ -0,0 +1,21 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('blocked_ips', function (Blueprint $table): void {
+            $table->id();
+            $table->unsignedInteger('user_id');
```
Roardom, line 12: Probably should block ranges instead of IPs directly. Blocked IPs aren't always associated with a user, so the user_id should be nullable as well.

app/Models/BlockedIp.php
Roardom, line 1: Probably want the inverse HasMany relationship on the User model. Would allow adding a blocked IP from the user's profile, and displaying a list of the user's blocked IPs.

database/migrations/2023_09_10_234654_create_blocked_ips_table.php
HDVinnie: So the user_id here is not related to a blocked member but is just to show what staff member initiated the block on said IP.

Roardom: Ah, that makes sense. Ranges would still be useful however if easy enough.

Audionut: Wildcards would be nice also. xxx.xxx.xxx.* for instance, to block the /24

PR #3139, opened 2023-09-11
(Update) User registration

HDVinnie:


Roardom (approved):
Would be nice to remove these if statements with `Rule::when()` in a future PR

PR #3142, opened 2023-09-15
(Add) Blacklist to disallow uploads of certain types by certain groups

Obi-Wana:
Adds a blacklist allowing to add groups and specific types (selection from `unit3d.types`) to be blacklisted, or all types if none is selected.


Audionut:
What about having a token based bypass system? There can be certain circumstances where a certain group may be allowed.

Having a one-use token type system would allow staff to give a user a token, allowing the user to upload a group who would otherwise be blacklisted.

PR #3143, opened 2023-09-15
(Add) Top 10 failed login IPs

Obi-Wana:
auth_ip_log

### DIFF
--- a/app/Http/Livewire/FailedLoginSearch.php
+++ b/app/Http/Livewire/FailedLoginSearch.php
@@ -26,6 +26,7 @@ class FailedLoginSearch extends Component
     public string $username = '';
 
     public string $userId = '';
+
     public string $ipAddress = '';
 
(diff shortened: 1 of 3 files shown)


Audionut:
Would be super convenient if the reported IP's were clickable, directing to the relevant search `ipAddress=xxx.xxx.xxx.xxx`

Would be nice to have a count of the different user ids that an IP has failed on also, as this would differentiate between users who may consistently fail to login, vs IP's targeting various user ids.

Obi-Wana:
> Would be super convenient if the reported IP's were clickable, directing to the relevant search `ipAddress=xxx.xxx.xxx.xxx`
> 
> Would be nice to have a count of the different user ids that an IP has failed on also, as this would differentiate between users who may consistently fail to login, vs IP's targeting various user ids.

Going to keep this in mind. Will add an updated version if I have time.

PR #3194, opened 2023-10-18
(Fix) Torrent card posters for other meta

Roardom:


Gideon75:
hi, the fix doesn't work for me, I keep seeing the white posters in the poster view of the NO-META torrents
[https://i.gyazo.com/4ade428f3c27226755e1080a96d85c7b.mp4](url)

Roardom:
> fix work for me, but NO-META torrents are not seen in the "poster" and "Group" view

Thanks for confirming it works. Poster and group views depend on there being meta for the torrent. Without meta, there's nothing to group by. The same goes for music meta and game meta. Refactoring all of these to store their relevant works would have to be done first (see #2691).

PR #3221, opened 2023-11-06
(Fix) Bad Spacing and Special Characters in Torrent Icons

ThyThal:
It is a bit much to have all in one line of code I would suggest to have it spaced in each tag to make it more readable.
Also fixed a formatting error for refundable torrent.

### DIFF
--- a/resources/views/components/partials/_torrent-icons.blade.php
+++ b/resources/views/components/partials/_torrent-icons.blade.php
@@ -56,10 +56,10 @@
                 'fa-calendar-star' => ! $alwaysFreeleech && $torrent->fl_until !== null,
             ])
             title="{{ 
-                ($personalFreeleech ? __('torrent.personal-freeleech') . ($tagsMet++ >= 1 ? "\n" : '') : '') . 
-                ($torrent->freeleechTokens_exists ? __('torrent.freeleech-token') . ($tagsMet++ >= 1 ? "\n" : '') : '') . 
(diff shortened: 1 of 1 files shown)


Roardom (changes requested):
I like the idea, but the new logic doesn't work. For example, if a user has personal freeleech, is in a group with freeleech, and the torrent is set to 100% free, it displays like this:

```
Personal FreeleechSpecial Freleeech

100% free
```

Roardom (changes requested):
Spacing is proper now, thanks!

However, it just dawned on me that this could be a whole lot cleaner using `implode()` and `array_keys()`. Would you mind implementing it like this instead? (Sorry for the back and forth)

```blade
            title="{{
                implode("\n", array_keys([
                    __('torrent.personal-freeleech') => $personalFreeleech,
                    __('torrent.freeleech-token')    => $torrent->freeleechTokens_exists,
                    __('torrent.special-freeleech')  => auth()->user()->group->is_freeleech,
                    __('torrent.global-freeleech')   => config('other.freeleech'),
                    $torrent->free . '% ' . __('common.free') . ($torrent->fl_until !== null ? ' (expires ' . $torrent->fl_until->diffForHumans() . ')' : '') => $torrent->free > 0,
                ], true))
            }}"
```

ThyThal:
Sure no problems, I never did web development or used php so I am not familiar with this kind of stuff but I will keep it in mind for the future!

Roardom (approved):
Thanks!

PR #3224, opened 2023-11-06
(Update) Improved Magnet Button for Torrent Card

ThyThal:
The Torrent Card now contains the magnet button if the magnets are enabled.

image
image

### DIFF
--- a/resources/views/torrent/partials/buttons.blade.php
+++ b/resources/views/torrent/partials/buttons.blade.php
@@ -1,30 +1,33 @@
 <menu class="torrent__buttons form__group--short-horizontal">
-    <li class="form__group form__group--short-horizontal">
-        @if ($fileExists = file_exists(public_path().'/files/torrents/'.$torrent->file_name))
-            @if (config('torrent.download_check_page') == 1)
-                <a
-                    class="form__button form__button--filled form__button--centered"
-                    href="{{ route('download_check', ['id' => $torrent->id]) }}" role="button"
-                >
-                    <i class='{{ config("other.font-awesome") }} fa-download'></i> {{ __('common.download') }}
-                </a>
-            @else
-                <a
-                    class="form__button form__button--filled form__button--centered"
-                    href="{{ route('download', ['id' => $torrent->id]) }}"
-                >
-                    <i class='{{ config("other.font-awesome") }} fa-download'></i> {{ __('common.download') }}
-                </a>
-            @endif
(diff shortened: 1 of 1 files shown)


ThyThal:
For some reason in my testing the "Use a Freeleech Token" breaks its text even tho that line is unchanged.
https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/127c240383ba39efd95929b5ae483262629b029a/resources/views/torrent/partials/buttons.blade.php#L43
image

Roardom (changes requested):
The magnet link config is only intended for allowing magnet links as fallback when the .torrent files don't exist (rather than not allowing downloads at all). Very few clients support private magnets. Private magnets also allow downloading torrent files without announcing to the tracker which shouldn't be encouraged. However, it would be good to change the `else` for if the file exists, to `elseif(config('torrent.magnet')` so that it only shows the magnet if the config is enabled.

On that note, if you're up for it, you could potentially add another config option for forcing the magnet button display all the time in a way that's backwards compatible with the current config option.

Roardom:
> For some reason in my testing the "Use a Freeleech Token" breaks its text even tho that line is unchanged.
> 
> https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/127c240383ba39efd95929b5ae483262629b029a/resources/views/torrent/partials/buttons.blade.php#L43
> 
> image

This is broken, yes. Ideally, we should remove all the html from the localization files

https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/v7.0.3/lang/en/torrent.php#L74

ThyThal:
I do not quite get what you are trying to say, but I get that it could be a good idea to create a new config key.

My implementation keeps the original functions and on top of that if the magnet in the config is enabled it will show both buttons. What I did was separate the buttons so they can be independent from each other and not override it.

Roardom:
So, the issue with your code is that generally private trackers do not want to allow users to use magnet links, because trackers want to track each time a user downloads a torrent, and a magnet link bypasses that. The magnet link should therefore only be used when there's no other alternative. That's why there should always be an option to only allow .torrent file downloads.

While looking at this code with this PR, I noticed a related bug, which is that the site will always show a magnet link if the .torrent file doesn't exist, instead of only allowing the .magnet link if the config allows it. I was hoping you could account for this in your PR.

Because this is an unwanted feature for tracker operators most of the time, that's the reason why it makes sense to make this feature separate from the default config. If you desire this feature, you can implement it with a different config key, that way it can be turned off by default, and have a note in its config file about how it easing bypassing of the torrent download logging.

Issue #2422, opened 2022-10-01
[Bug] Undefined array key 0


LostRager:
### UNIT3D Version

6.3.0

### PHP & Platform

8.1.8

### Laravel verion

9.20.0

### Have you done this?

- [X] I am willing to share my stack trace and logs
- [ ] I can suggest a fix as a Pull Request

### Expectation

not sure what the problem is

### Description

not sure what the problem is

### Stack trace & logs

```shell
Undefined array key 0 {"exception":"[object] (ErrorException(code: 0): Undefined array key 0 at /var/www/html/vendor/laravel/framework/src/Illuminate/Collections/Collection.php:1683)
[stacktrace]
#0 /var/www/html/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(257): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError()
#1 /var/www/html/vendor/laravel/framework/src/Illuminate/Collections/Collection.php(1683): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->Illuminate\\Foundation\\Bootstrap\\{closure}()
#2 /var/www/html/app/Http/Controllers/Auth/ActivationController.php(38): Illuminate\\Support\\Collection->offsetGet()
#3 /var/www/html/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): App\\Http\\Controllers\\Auth\\ActivationController->activate()
#4 /var/www/html/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(45): Illuminate\\Routing\\Controller->callAction()
#5 /var/www/html/vendor/laravel/framework/src/Illuminate/Routing/Route.php(268): Illuminate\\Routing\\ControllerDispatcher->dispatch()
#6 /var/www/html/vendor/laravel/framework/src/Illuminate/Routing/Route.php(211): Illuminate\\Routing\\Route->runController()
#7 /var/www/html/vendor/laravel/framework/src/Illuminate/Routing/Router.php(725): Illuminate\\Routing\\Route->run()
#8 /var/www/html/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(141): Illuminate\\Routing\\Router->Illuminate\\Routing\\{closure}()
[...]

PR #3242, opened 2023-11-15
(Update) Laravel Log Viewer

LostRager:
Fixes the "Undefined array key 0" error. #2422
Adding Clear Current Log function + Adding Clear Latest Log button.
Adding Delete All Logs function + Adding Clear All Logs button.


app/Http/Livewire/LaravelLogViewer.php
```diff
@@ -55,7 +55,9 @@ final public function getEntriesProperty(): \Illuminate\Contracts\Pagination\Len
         $logString = '';
 
         foreach ($this->logs as $log) {
```
Roardom, line 57: Does `foreach ($this->logs ?? [] as $log)` or `foreach ($this->logs ?: [] as $log)` fix the issue? If so, could you try using it instead of the `isset()`?

app/Http/Livewire/LaravelLogViewer.php
```diff
@@ -94,6 +96,25 @@ final public function getEntriesProperty(): \Illuminate\Contracts\Pagination\Len
...
+    final public function clearLatestLog()
+    {
+        $latestLogFile = $this->logFiles->first();
+
+        if ($latestLogFile) {
+            File::put($latestLogFile->getPathname(), '');
+        }
+    }
+
+    final public function deleteAllLogs()
```
Roardom, line 108:
```suggestion
    final public function deleteAllLogs(): void
```

app/Http/Livewire/LaravelLogViewer.php
```diff
@@ -94,6 +96,25 @@ final public function getEntriesProperty(): \Illuminate\Contracts\Pagination\Len
         return new LengthAwarePaginator($currentEntries, $groupedEntries->count(), $this->perPage, $this->page);
     }
 
+    final public function clearLatestLog()
```
Roardom, line 99:
```suggestion
    final public function clearLatestLog(): void
```

Roardom (changes requested):

app/Http/Livewire/LaravelLogViewer.php
LostRager: ?? and ?: did not work sadly.

Roardom: Darn, let's not worry about this then. Probably best to move it back to how you had originally implemented it.

app/Http/Livewire/LaravelLogViewer.php
Roardom, line 1: You'll probably need to add a docblock right above the class to clear the last CI warning:

```php
/**
 * @property \Illuminate\Support\Collection $logFiles
 * @property \Illuminate\Pagination\LengthAwarePaginator $entries
 */
class LaravelLogViewer extends component
```

Roardom (changes requested):

PR #3243, opened 2023-11-15
(Remove) Google fonts

Roardom:
Why, just why.


HDVinnie:
`resources/sass/main/aboutus.scss` is not even used anymore...

Roardom:
How on earth is the network call for this font still showing up in dev console then?

Roardom:
Unless you're meaning we should just get rid of aboutus.scss completely, which was on my list of things to do next.

HDVinnie:
I mean the file is there and compiled in aboutus.scss but its not actually used anywhere anymore since your css/views refactor. So this PR could just remove the aboutus.scss and remove the import in app.scss

PR #3246, opened 2023-11-16
(Add) Apikey logging

Roardom:


HDVinnie (changes requested):
`'users.apikeys.edit'` in test needs to be new `'users.apikeys.index'`

PR #3255, opened 2023-11-18
Add TOTP status to user profile

Obi-Wana:
Adds a status icon, if TOTP is enabled, to user profiles.

### DIFF
--- a/resources/views/user/profile/show.blade.php
+++ b/resources/views/user/profile/show.blade.php
@@ -634,6 +634,14 @@
                     <dd>{{ $user->id }}</dd>
                     <dt>{{ __('common.email') }}</dt>
                     <dd>{{ $user->email }}</dd>
+                    <dt>2FA Enabled</dt>
+                    <dd>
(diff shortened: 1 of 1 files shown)


resources/views/user/profile/show.blade.php
```diff
@@ -634,6 +634,14 @@ class="form__button form__button--text"
                     <dd>{{ $user->id }}</dd>
                     <dt>{{ __('common.email') }}</dt>
                     <dd>{{ $user->email }}</dd>
+                    <dt>2FA Enabled</dt>
+                    <dd>
+                        @if ($user->two_factor_confirmed_at != NULL)
```
Roardom, line 639:
```suggestion
                        @if ($user->two_factor_confirmed_at !== null)
```

PHP generally uses lowercase for null. Also best practice to check for null explicitly instead of falsey values.

Roardom (changes requested):

Roardom (approved):

PR #3257, opened 2023-11-18
(Update) Chatbox.vue

LostRager:
Auto add [img] to image links and [url] for urls in chat.

Not sure if this is a wanted PR with the livewire chat coming up in the future, but i thought i might share it anyhows.


Roardom:
Thanks, but indeed all of this code will be removed very soon.

PR #3265, opened 2023-11-22
(Fix) Missing parameter user in subforum

LostRager:


Roardom:
Probably would be a better idea to try and figure out why that variable is null instead of working around the error.

Roardom:
Do you have it eager loaded on https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/83dd50607196da9c323e021380afb83bcf9f3ee4/app/Http/Livewire/SubscribedForum.php#L27 and https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/83dd50607196da9c323e021380afb83bcf9f3ee4/app/Http/Livewire/ForumTopicSearch.php#L64?

If so, perhaps the `last_post_user_id` field is empty on your forum, or belongs to a user that no longer exists. In which case that issue should be fixed instead.

LostRager:
> Do you have it eager loaded on
> 
> https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/83dd50607196da9c323e021380afb83bcf9f3ee4/app/Http/Livewire/SubscribedForum.php#L27
> 
> and
> https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/83dd50607196da9c323e021380afb83bcf9f3ee4/app/Http/Livewire/ForumTopicSearch.php#L64
> 
> ?
> If so, perhaps the `last_post_user_id` field is empty on your forum, or belongs to a user that no longer exists. In which case that issue should be fixed instead.

Yes both are there. 

Reverted change and tested again with `?? 0` to find the offending part. Its a user that got pruned. When i click into a topic the person links to system and when i click into post its listed as system. So that seems to work correctly. 

Just forum index that has the issue.

Roardom:
Ah, if it's a pruned user, the solution is the following:

Change
```php
->with(['lastestPoster'])
```
into
```php
->with(['lastestPoster' => fn ($query) => $query->withTrashed()])
```

LostRager:
> Ah, if it's a pruned user, the solution is the following:
> 
> Change
> 
> ```
> ->with(['lastestPoster'])
> ```
> 
> into
> 
> ```
> ->with(['lastestPoster' => fn ($query) => $query->withTrashed()])
> ```

hmm, changing it in the SubscribedForum.php did not work,
but changing it in the ForumController.php, did work.
`'forums.latestPoster' => fn ($query) => $query->withTrashed(),`

Roardom:
The benefit is being able to remove `:username` to change all user routes to use the id instead of the username: https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/83dd50607196da9c323e021380afb83bcf9f3ee4/routes/web.php#L378

LostRager:
> The benefit is being able to remove `:username` to change all user routes to use the id instead of the username:
> 
> https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/83dd50607196da9c323e021380afb83bcf9f3ee4/routes/web.php#L378

Ahh that makes sense, thanks for explaining.

PR #3292, opened 2023-12-12
(Update) Fortify Auth

HDVinnie:


app/Providers/FortifyServiceProvider.php
```diff
@@ -149,42 +152,56 @@ public function boot(): void
 
             $user = User::query()->where('username', $request->username)->first();
 
+            if ($user && !Hash::check($request->password, $user->password)) {
```
Roardom, line 155: We shouldn't be hashing the password twice. If we did, we should consider increasing the default bcrypt value. This should be memoized with the following if statement.

app/Providers/FortifyServiceProvider.php
```diff
@@ -149,42 +152,56 @@ public function boot(): void
...
                     $request->session()->invalidate();
 
                     throw ValidationException::withMessages([
-                        Fortify::username() => trans('auth.not-activated'),
+                        Fortify::username() => __('auth.not-activated'),
                     ]);
                 }
 
                 // Check if user is banned
-
```
Roardom, line 165: Personally, I add an empty line after a comment, when the comment describes code that include empty lines within it, but not going to force anything here.

config/auth.php
```diff
@@ -116,4 +116,18 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define the amount of minutes before an email verification
+    | times out and the user is prompted to re-verify their email via the
+    | verify screen. By default, the timeout lasts for one day.
+    |
+    */
+    'verification' => [
+        'expire' => 1440,
+    ],
```
Roardom, line 132: Where is this used? Couldn't find any documentation about it.

Roardom (commented):
Gave it a quick lookover.

config/auth.php
HDVinnie: https://github.com/laravel/framework/blob/7907bfb50165b3951403e2d9882b4f8e0ee26b9e/src/Illuminate/Auth/Notifications/VerifyEmail.php#L85

Not documented for included in default laravel auth config

PR #3315, opened 2023-12-28
(Update) Clean up stylesheets

Roardom:
More work on finishing cleaning up the styles to match the intended 7-1 sass architecture.


Roardom:
I think we should be good to go now.

PR #3329, opened 2023-12-31
(Update) Block system reserved ports

Roardom:
They have a very high chance to be fake and not connectable.


HDVinnie:
Should probably remove all the BLACK_PORTS const that are below 1024

Roardom:
> Should probably remove all the BLACK_PORTS const that are below 1024

Well spotted. Done.

PR #3362, opened 2024-01-08
(Refactor) Deprecated rules

HDVinnie:
- Illuminate\Contracts\Validation\Rule is deprecated

### DIFF
--- a/app/Rules/EmailBlacklist.php
+++ b/app/Rules/EmailBlacklist.php
@@ -1,13 +1,24 @@
 <?php
+/**
+ * NOTICE OF LICENSE.
+ *
+ * UNIT3D Community Edition is open-sourced software licensed under the GNU Affero General Public License v3.0
+ * The details is bundled with this project in the file LICENSE.txt.
+ *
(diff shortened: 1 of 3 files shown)


app/Rules/EmailBlacklist.php
```diff
@@ -46,11 +59,8 @@ protected function shouldUpdate(): void
     {
         $autoupdate = config('email-blacklist.auto-update');
 
-        try {
-            if ($autoupdate && !cache()->has(config('email-blacklist.cache-key'))) {
-                EmailBlacklistUpdater::update();
-            }
-        } catch (InvalidArgumentException) {
+        if ($autoupdate && !cache()->has(config('email-blacklist.cache-key'))) {
+            EmailBlacklistUpdater::update();
```
Roardom, line 63: We haven't hit the CI error caused by the email blacklist updater for awhile, but if we start seeing it again, we'll know this try-catch was what's been fixing it.

Roardom (approved):

PR #3397, opened 2024-01-17
(Refactor) Command performance

HDVinnie:


app/Console/Commands/AutoUpsertHistories.php
```diff
@@ -64,19 +64,14 @@ public function handle(): void
...
-            if ($histories === false) {
-                break;
-            }
-
-            $histories = array_map('unserialize', $histories);
+        $allHistories = Redis::connection('announce')->command('LPOP', [$key, $historyCount]);
+        $allHistories = array_map('unserialize', $allHistories);
+        $chunks = array_chunk($allHistories, $historiesPerCycle);
 
+        foreach ($chunks as $histories) {
```
Roardom, line 74: It's better to keep redis queries short and simple as it's single-threaded and can not handle requests concurrently. It should also be compared to false, otherwise it causes errors if no user announces between intervals.

app/Console/Commands/AutoUpsertPeers.php
Roardom, line 1: Ditto

app/Console/Commands/AutoUpsertAnnounces.php
Roardom, line 1: Ditto

app/Console/Commands/AutoWarning.php
```diff
@@ -61,14 +61,16 @@ public function handle(): void
...
+                ->whereIn('user_id', $hitrun->pluck('user.id'))
+                ->get();
+
             foreach ($hitrun as $hr) {
                 if (!$hr->user->group->is_immune && $hr->actual_downloaded > ($hr->torrent->size * (config('hitrun.buffer') / 100))) {
-                    $exsist = Warning::withTrashed()
-                        ->where('torrent', '=', $hr->torrent->id)
-                        ->where('user_id', '=', $hr->user->id)
-                        ->first();
+                    $exsist = $warnings->where('torrent', '=', $hr->torrent->id)->firstWhere('user_id', $hr->user->id);
```
Roardom, line 71: Add a relation between histories and warnings so you can just do `->with('warning')`

Roardom, line 71: let's fix the `$exsist` typo too while we're at it

app/Console/Commands/SyncPeers.php
```diff
@@ -43,20 +43,18 @@ class SyncPeers extends Command
...
+        LEFT JOIN (
+            SELECT torrent_id,
+            SUM(IF(left = 0 AND active = 1, 1, 0)) AS updated_seeders,
+            SUM(IF(left != 0 AND active = 1, 1, 0)) AS updated_leechers
+            FROM peers
+            GROUP BY torrent_id
+        ) AS seeders_leechers ON torrents.id = seeders_leechers.torrent_id
+        SET seeders = COALESCE(seeders_leechers.updated_seeders, 0),
+        leechers = COALESCE(seeders_leechers.updated_leechers, 0)'
+        );
```
Roardom, line 54: No reason to ditch the orm here. Also, can do the same method as described above about dividing all torrents into 25 and doing 25 queries.

app/Console/Commands/SyncPeers.php
```diff
@@ -43,20 +43,18 @@ class SyncPeers extends Command
      */
     public function handle(): void
     {
-        Torrent::withoutGlobalScope(ApprovedScope::class)
-            ->leftJoinSub(
-                Peer::query()
-                    ->select('torrent_id')
-                    ->addSelect(DB::raw('SUM(peers.left = 0 AND peers.active = 1) AS updated_seeders'))
-                    ->addSelect(DB::raw('SUM(peers.left != 0 AND peers.active = 1) AS updated_leechers'))
```
Roardom, line 51: We can improve this to either `->select(['torrent_id', DB::raw(...), DB::raw(...)])` or `->selectRaw(...)`

app/Console/Commands/AutoStatsClients.php
```diff
@@ -38,14 +38,7 @@ class AutoStatsClients extends Command
...
-                fn ($sub) => $sub
-                    ->select(['agent', 'user_id'])
-                    ->from('peers')
-                    ->groupBy('agent', 'user_id'),
-                'distinct_agent_user'
-            )
+        $clients = Peer::selectRaw('agent, COUNT(DISTINCT user_id) as count')
             ->groupBy('agent')
             ->orderBy('agent')
             ->get()
```
Roardom, line 44: Should be able to use `->pluck('count', 'agent')` instead of mapWithKeys

Roardom (changes requested):

app/Console/Commands/AutoStatsClients.php
HDVinnie: `fromSub` actually is faster then `selectRaw('agent, COUNT(DISTINCT user_id) as count')`.

PR #3409, opened 2024-01-19
(Fix) Fortify Rate Limiters

HDVinnie:


phpstan.neon
Roardom, line 1: Let's try and workaround the ide not figuring out that a return isn't necessary instead of disabling phpstan for this file.

Roardom (changes requested):
Looks good, but let's get phpstan working properly.

PR #3414, opened 2024-01-21
(Add) Email logging

Roardom:
Don't log the email addresses themselves to maintain user privacy.


Roardom:
Unfortunately, livewire still has poor support for template inheritance, I'll have to redo this the way I did the others with an additional Controller file.

PR #3457, opened 2024-02-01
(Add) Most recently downloaded torrents to home page

Roardom:


HDVinnie:
I feel like if this new tab is going to use figures then the other tabs should be updated from tables to figures in this same PR or visa versa. Otherwise this PR adds inconsistency?

Roardom:
> I feel like if this new tab is going to use figures then the other tabs should be updated from tables to figures in this same PR or visa versa. Otherwise this PR adds inconsistency?

Yes, in fact there's an issue with the current implementation. We're currently caching the results... and the results contain user-related data (seeding/leeching history, bookmarks).

PR #3462, opened 2024-02-01
(Fix) Revision of errors in Ukrainian translation and translation of untranslated lines

crisukbot:


Roardom:
Hi, can you please target the 7.x.x branch

lang/uk/user.php
```diff
@@ -221,13 +221,13 @@
...
     'recent-followers'                     => 'Останні послідовники',
-    'registration-date'                    => 'Дата Реєстрації',
+    'registration-date'                    => 'Дата реєстрації',
     'report'                               => 'Доповідь',
     'request-help'                         => 'Керуйте спільним доступом до певної статистики, пов язаної з запитом, та інформацією з групами, яким дозволено доступ до вашого профілю. Ці налаштування перевизначаються, якщо ви не дозволяєте будь-яким групам отримувати доступ до запитаної статистики та інформації, або якщо ви <strong>йдете на приват</strong>',
     'request'                              => 'Запит',
     'requested'                            => 'Запитано',
     'requests'                             => 'Запити',
-    'request-comments'                     => 'Запит Коментарі зроблено',
+    'request-comments'                     => 'Запит коментарів зроблено',
```
Roardom, line 230: I'm pretty sure this translation is incorrect. The string used to be used for the profile statistic of how many comments a user has made on torrent requests. Google translate tells me the new translation means "The request for comments has been made" which isn't related.

lang/uk/subtitle.php
```diff
@@ -17,14 +17,14 @@
...
-                                        <li>Subtitles must be in sync with the video</li>
-                                        <li>ZIP archives are only allowed for SUB+IDX or as bundles of the same language for a TV season pack</li>
-                                        <li>Repeated uploads of junk subs will constitute a violation and subject to a disciplinary action</li>
-                                        <li>Keep the Note of the subtitle short, NO urls/links are allowed</li>
+                                        <li>Допускаються лише точні субтитри (правильна частота кадрів, переклад, орфографія, хронометраж)</li>
+                                        <li>Без гугл-перекладу, машинного перекладу та некоректних субтитрів</li>
+                                        <li>Субтитри повинні бути синхронізовані з відео</li>
+                                        <li>ZIP-архіви дозволені лише для SUB+IDX або як пакети однієї мови для пакету телесезонів</li>
+                                        <li>Повторне завантаження непотрібних "підводних човнів" буде вважатися порушенням і підлягатиме покаранню</li>
```
Roardom, line 26: Is `підводних човнів` the correct term for subtitles? Google translate is telling me it means aquatic submarine vessels instead.

lang/uk/request.php
```diff
@@ -33,13 +33,13 @@
     'claim-anon-choose'   => 'Будь ласка, робіть свій вибір уважно',
     'claim-as-anon'       => 'Чи хотіли б Ви створити заявку анонімно?',
     'claim-now'           => 'Створити заявку на виконання',
-    'claimed'             => 'Заявлено',
-    'claimed-success'     => 'Request Successfully Claimed',
+    'claimed'             => 'Створено',
+    'claimed-success'     => 'Запит успішно створено',
```
Roardom, line 37: Can you double check these are correct? Google translate tells me "The request was created successfully" while the translations are referring to claiming the request. Claiming a request is different than creating a request. Claiming is referring to the act of reserving the request so that no one else is allowed to fill it except for yourself. Claims automatically expire after 1 week.

Roardom (commented):
Thanks for the translations. I noticed a few (potential) mistakes when going over all the strings using Google Translate. And while I know Google Translate isn't perfect, I wanted to make sure that if you're using automated software to do these translations, that you're making sure they're accurate. Of course, it could just be Google Translate on my side not being accurate enough.

lang/uk/user.php
crisukbot: > I'm pretty sure this translation is incorrect. The string used to be used for the profile statistic of how many comments a user has made on torrent requests. Google translate tells me the new translation means "The request for comments has been made" which isn't related.

I checked and yes, I was wrong.. I'm correcting it now.

lang/uk/subtitle.php
crisukbot: > Is `підводних човнів` the correct term for subtitles? Google translate is telling me it means aquatic submarine vessels instead.

yep. This translation of these lines seems to be the best fit.

lang/uk/request.php
crisukbot: > Can you double check these are correct? Google translate tells me "The request was created successfully" while the translations are referring to claiming the request. Claiming a request is different than creating a request. Claiming is referring to the act of reserving the request so that no one else is allowed to fill it except for yourself. Claims automatically expire after 1 week.

Thanks for pointing it out, I've fixed it.

lang/uk/subtitle.php
crisukbot: > > Is `підводних човнів` the correct term for subtitles? Google translate is telling me it means aquatic submarine vessels instead.
> 
> yep. This translation of these lines seems to be the best fit.

 I changed my mind and translated this line differently.

lang/uk/user.php
Roardom: It appears the updated translation is treating "request" as a verb instead of a noun. To give some context, an example of the string being used would be:

```
Torrent comments made: 4
Article comments made: 6
Request comments made: 8
Playlist comments made: 10
```

Can you confirm your translation provides this meaning? When using google translate, it returns "The comment request is made" which makes me think it's referring to an action being performed instead of the count of how many actions were performed.

Roardom:
Thank you!

PR #3485, opened 2024-02-06
(Update) Exclude adult media from random media

Roardom:


HDVinnie (changes requested):
This should be a user setting. Can be false by default. I dont think it should be hardcoded in like this though.

Roardom:
We should see about moving all the user general settings off the users table and into their own table.

PR #3515, opened 2024-02-11
(Update) Migrate & Refactor Chat into Vue 3

clandestine8:
Migrate & Refactor Chat into Vue 3 without changing functionality. This should give us a cleaner base to work from going forward, improved performance from Vue 3 and better structuring, long term support.


database/seeders/UsersTableSeeder.php
```diff
@@ -49,7 +49,7 @@ public function run(): void
                 'username'          => config('unit3d.owner-username'),
                 'email'             => config('unit3d.default-owner-email'),
                 'email_verified_at' => now(),
-                'group_id'          => 10,
+                'group_id'          => 9,
```
Roardom, line 52: Can we keep this as is? I just found a hardcoded check for `group_id = 10` the other day for checking if a user is owner and hope there's no more, but better be safe than sorry.

Gideon75:
Can we now edit chat?

database/seeders/UsersTableSeeder.php
HDVinnie: This was duet to mistake in PR #3502 . A commit has been pushed to 7.x.x to revert that. This can be put back to 10 .

PR #3540, opened 2024-02-19
(Add) Notification filters to work as locical "or" filter

Obi-Wana:
Most checkbox filters within the same row use an "or" logic, for example, the resolution filter on `TorrentSearch..php`.
Therefore, it should also be possible to filter notifications by multiple types.

This MR adds the ability to display notifications which are either of type X or of type Y and so on.


Roardom (approved):

Roardom:
Thank you!

PR #3551, opened 2024-02-22
(Refactor) Normalize bon transactions

Roardom:


Roardom:
We won't worry about bon store transactions for now. Will leave that beast for another day. Meanwhile, we got torrent/post tips and gifts normalized, while requests bon data is stored in other tables.

Issue #3393, opened 2024-01-15
[Bug] Refund code does not seem to work


LostRager:
### UNIT3D Version

7.1.2

### PHP & Platform

8.2.12

### Laravel verion

10.34.1

### Have you done this?

- [X] I am willing to share my stack trace and logs
- [ ] I can suggest a fix as a Pull Request

### Expectation

Refund is not refunding download

### Description

Refund does not refund at all. No logs.

Torrents are marked as refundable.

### Are you a UNIT3D sponsor?

Yes, with my username

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/HDInnovations) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/HDInnovations/UNIT3D-Community-Edition/issues/3393">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/3393/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/3393/pledge.svg">
</picture>
</a>

PR #3556, opened 2024-02-22
(Add) Refundable option to API

Obi-Wana:
Allows to:
- Set refundable on upload
- Query API for refundable uploads
- Query API for non-refundable uploads

Allows to select or deselect refundable. If refundable is not added to GET requests, it will be ignored.

### DIFF
--- a/app/Http/Controllers/API/TorrentController.php
+++ b/app/Http/Controllers/API/TorrentController.php
@@ -163,6 +163,7 @@ class TorrentController extends BaseController
         $torrent->internal = $user->group->is_modo || $user->group->is_internal ? $request->input('internal') : 0;
         $torrent->featured = $user->group->is_modo || $user->group->is_internal ? $request->input('featured') : 0;
         $torrent->doubleup = $user->group->is_modo || $user->group->is_internal ? $request->input('doubleup') : 0;
+        $torrent->refundable = $user->group->is_modo || $user->group->is_internal ? $request->input('refundable') : 0;
         $du_until = $request->input('du_until');
 
         if (($user->group->is_modo || $user->group->is_internal) && isset($du_until)) {
@@ -231,6 +232,7 @@ class TorrentController extends BaseController
             'featured'         => 'required',
             'free'             => 'required|between:0,100',
             'doubleup'         => 'required',
+            'refundable'       => 'required',
             'sticky'           => 'required',
         ]);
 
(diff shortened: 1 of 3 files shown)


HDVinnie (approved):

Roardom:
Do you know if the refund code even works? (#3393)

app/Http/Controllers/API/TorrentController.php
```diff
@@ -434,6 +436,7 @@ public function filter(Request $request): TorrentsResource|\Illuminate\Http\Json
                 ->when($request->filled('adult'), fn ($query) => $query->ofAdult($request->boolean('adult')))
                 ->when($request->filled('free'), fn ($query) => $query->ofFreeleech($request->free))
                 ->when($request->filled('doubleup'), fn ($query) => $query->doubleup())
+                ->when($request->filled('refundable'), fn ($query) => $query->where('refundable', $request->boolean('refundable')))
```
Roardom, line 439: Can you please move this filter to a scope in https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/master/app/Traits/TorrentFilter.php

Roardom (changes requested):

app/Http/Controllers/API/TorrentController.php
Roardom: Looks like we already have a scope in that file actually. You can use it directly.

HDVinnie, line 439: This should use the `scopeRefundable` in `app/Traits/TorrentFilter.php`

HDVinnie (changes requested):

Obi-Wana:
> Do you know if the refund code even works? (#3393)

@Roardom I don't know, I was not aware of any issues with the refundable code itself.

app/Http/Controllers/API/TorrentController.php
Obi-Wana: @Roardom Updated as requested.

Roardom (changes requested):
It should just be `->refundable()` like the other boolean flags. You can find the documentation for local scopes here: https://laravel.com/docs/10.x/eloquent#local-scopes

Roardom:
I should have made it more clear. The torrent search page already uses the `->refundable()` scope, so you can't rename it. If you want to offer this functionality of being able to search when it's false as well, you have to make a new scope instead of renaming the existing one.

Roardom (approved):
Dropped the latest commit and squashed the remaining ones 👍

PR #3590, opened 2024-03-01
(Update) SyncPeers command

HDVinnie:
- reverts query style to how it was while also adding a new query to handle history times_completed. While this is slow as history table can be one of the biggest tables it is wrapped in a transaction to avoid deadlocks and this is needed.

### DIFF
--- a/app/Console/Commands/SyncPeers.php
+++ b/app/Console/Commands/SyncPeers.php
@@ -62,7 +62,7 @@ class SyncPeers extends Command
                     'leechers'   => DB::raw('COALESCE(seeders_leechers.updated_leechers, 0)'),
                     'updated_at' => DB::raw('updated_at')
                 ]);
-        });
+        }, 5);
 
         DB::transaction(function (): void {
             DB::statement("
--- a/app/Console/Commands/SyncPeers.php
+++ b/app/Console/Commands/SyncPeers.php
@@ -13,10 +13,13 @@
 
 namespace App\Console\Commands;
 
+use App\Models\History;
+use App\Models\Peer;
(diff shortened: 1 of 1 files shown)


app/Console/Commands/SyncPeers.php
```diff
@@ -35,37 +37,44 @@ class SyncPeers extends Command
...
+        DB::transaction(function (): void {
+            DB::statement("
+                UPDATE torrents
+                    SET times_completed = (
+                        SELECT COUNT(*)
+                        FROM history
+                        WHERE `completed_at` IS NOT NULL AND torrent_id = torrents.id
+                    )
+            ");
+        });
```
Roardom, line 76:
```suggestion
        }, 5);
```

app/Console/Commands/SyncPeers.php
```diff
@@ -35,37 +37,44 @@ class SyncPeers extends Command
...
+                        ->groupBy('torrent_id'),
+                    'seeders_leechers',
+                    fn ($join) => $join->on('torrents.id', '=', 'seeders_leechers.torrent_id')
+                )
+                ->update([
+                    'seeders'    => DB::raw('COALESCE(seeders_leechers.updated_seeders, 0)'),
+                    'leechers'   => DB::raw('COALESCE(seeders_leechers.updated_leechers, 0)'),
+                    'updated_at' => DB::raw('updated_at')
+                ]);
+        });
```
Roardom, line 65:
```suggestion
        }, 5);
```

Roardom (changes requested):

Roardom (approved):

PR #3621, opened 2024-03-06
(Patch) Livewire URL Attributes

HDVinnie:
#TODO: Update URL attributes once Livewire 3 fixes upstream bug. See: https://github.com/livewire/livewire/discussions/7746


Roardom (approved):
Doesn't look like there's any other possibility for now. Just have to remember to revert this once the bug is fixed.

Issue #3631, opened 2024-03-07
[Bug] #3495 broke underlying user groups


Audionut:
### UNIT3D Version

7.x.x

### PHP & Platform

8.1.2

### Laravel verion

9.2.3

### Have you done this?

- [ ] I am willing to share my stack trace and logs
- [ ] I can suggest a fix as a Pull Request

### Expectation

User groups follow the customization set out for the group, without some other link to some other group.

### Description

User groups follow a top to bottom rule, with every user group having to meet the requirements of the previous user group, regardless of the specific customization of a user group.

It's impossible to have distinct Upload Amount user groups, _and_ distinct Seedsize user groups, for instance.

### Are you a UNIT3D sponsor?

No, don't give priority to this

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/HDInnovations) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/HDInnovations/UNIT3D-Community-Edition/issues/3631">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/3631/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/3631/pledge.svg">
</picture>
</a>

Audionut:
Fixed: https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/3691

PR #3632, opened 2024-03-08
Order groups by descending position order

Audionut:
Should allow having groups with widely different requirements.

With the removal of fixed groups and allowing front end editing, there is an oddity in having groups assigned on a first met basis, meaning that groups in a higher position with widely different requirements would be assigned to some lower position group (with that groups requirements having been met), rather than being assigned strictly based on the requirements set in the group editor.

By having the autogroup task run in descending order, this should mean that groups that fit within seedsize requirements, for instance, are assigned based on their seedsize. Since the task runs in descending order, users not meeting some higher class requirements keep getting pushed down the position ladder until autogroup finds the highest class that the user meets. 

Should fix https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/3631

### DIFF
--- a/app/Console/Commands/AutoGroup.php
+++ b/app/Console/Commands/AutoGroup.php
@@ -51,7 +51,7 @@ class AutoGroup extends Command
         $current = Carbon::now();
         $groups = Group::query()
             ->where('autogroup', '=', 1)
-            ->orderBy('position')
+            ->orderBy('position', 'desc')
             ->get();
 
         $users = User::query()


Roardom:
The current logic keeps promoting a user until it reaches the first group where the user doesn't meet the criteria, where it stops promoting. It's done this way so the site doesn't have to calculate the seedsize/seedtime for users where it doesn't need to be calculated.

Doing it backwards would mean the user would just stay in their current group unless they just so happen to match the first/highest group in the loop, in which they will most likely end up at being the lowest group after the loop finishes since they will most likely match all other groups as well.

PR #3691, opened 2024-03-28
(Refactor) AutoGroup User Assignment

EkoNesLeg:
Encapsulates the logic for assigning a user to a group and setting permissions within a single if condition; the `else` clause that breaks the loop when a user doesn't meet the conditions for a group is removed, so that promotion is no longer a sequential process. As such, users can can jump straight to Seeder/Archivist ranks.


Audionut:
Fixes https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/3631

PR #3694, opened 2024-03-28
(Fix) Ticket attachment upload

Alkl58:
This PR fixes the issue that nothing happens when selecting a file to attach to a ticket.

When looking in the browser console:
grafik

After applying the changes it's working again (for me).

Please check if the bug is reproducible.

Maybe using `upload()` was the main issue, as according to Livewire Docs:
grafik
https://livewire.laravel.com/docs/uploads#storing-uploaded-files


HDVinnie:
@Alkl58 good catch. This was missed when I did the Livewire 3 migration. Thanks for the PR.

PR #3695, opened 2024-03-28
(Fix) Torrent row with null meta

LostRager:
Fixes: 

Attempt to read property "rating" on null (View: /var/www/html/resources/views/components/torrent/row.blade.php)


HDVinnie:
@LostRager thanks for the pull request.

PR #3720, opened 2024-04-04
Update torrent-search.blade.php

BitPyrate:
A Drop Down box seems to make more sense for the year. 1888 being the oldest film posible.

Further possible improvements:
1. You might be able remove any field validation code regarding the year.
2. the drop down list could be made sensitive to the other field so your startyear is not after your endyear. giving you a cropped list to choose from.


HDVinnie:
Unless there's an overabundance of users that would rather select over numeric input, it's much easier to type a number than scroll through hundreds of select options. (also note the first audio recording was in the 1860s, have to add a few more years to the iterator to support the whole range for song recordings).

PR #3727, opened 2024-04-06
(Update) Automatic torrent freeleeches

vincejv:
Add double upload auto tagging and user group criteria to automatic freeleech feature

- Double Upload auto tagging
- By user group criteria

PR #3730, opened 2024-04-07
(Update) Add double upload and user group criteria to automatic torrent freeleeches

vincejv:
Add double upload auto tagging and user group criteria to automatic freeleech feature

- Double Upload auto tagging
- By user group criteria

Original PR: https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/3727

Recreated due to
> Create feature branches - Don't ask us to pull from your master branch.

### DIFF
--- a/app/Helpers/TorrentHelper.php
+++ b/app/Helpers/TorrentHelper.php
@@ -50,6 +50,8 @@ class TorrentHelper
         $torrent->moderated_at = now();
         $torrent->moderated_by = auth()->id();
 
+        $uploader = $torrent->user;
+
         if (!$torrent->free) {
             $autoFreeleechs = AutomaticTorrentFreeleech::query()
                 ->orderBy('position')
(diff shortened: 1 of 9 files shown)


vincejv:
Pushed changes to address CI Failure for static analysis
```
  Line   app/Models/AutomaticTorrentFreeleech.php                                 
 ------ ------------------------------------------------------------------------- 
  62     Method App\Models\AutomaticTorrentFreeleech::group() should return       
         Illuminate\Database\Eloquent\Relations\BelongsTo<App\Models\Resolution,  
         App\Models\AutomaticTorrentFreeleech> but returns                        
         Illuminate\Database\Eloquent\Relations\BelongsTo<App\Models\Group,       
         App\Models\AutomaticTorrentFreeleech>.
```

the other CI error messages seems not to be related to the code?

HDVinnie:
Few things here. 

1) I feel like AutomaticTorrentFreeleech was meant for just freeleech. Adding double upload and other things to it can be misleading code term wise.
2) So maybe this should just be refactored/renamed as a whole to AutomaticTorrentBuff. Controllers, models, tables etc?

@Roardom built this so maybe he will chime in but I deff feel like AutomaticTorrentBuff would be better here and can handle freeleech, double upload and more.

vincejv:
Just to put details on the "...on more" that would be the ff:
- Refundable
- Sticky
- Bump?
- Featured

HDVinnie:
> Just to put details on the "...on more" that would be the ff:
> 
> * Refundable
> * Sticky
> * ~Bump?~
> * Featured

yes just need to make sure there optional. At least atm thats what seems best to me. So its all confined to one system. Model, View, Controller, Migration (table), etc

Roardom:
There are two ways to go about this:

- make the freeleech_percentage column nullable, the is_double_upload column nullable, enforce validation to make sure that at least one of them isn't null, and only update the uploaded torrent's freeleech/double upload if the value isn't null
- have an automatic_double_upload_torrents table that replicates the automatic_freeleech_torrents_table

Otherwise the double upload would overwrite what's given even if you only want to change the freeleech.

HDVinnie:
@vincejv lmk if your up for changing as discussed. If not totally understand and we can visit it ourselves once we finish some other things. Regardless pull requests are always welcome so thank you.

vincejv:
hello! sorry for the late reply, if you have the bandwidth to make further changes and merge this on UNIT3D v8.x.x, I think I'd prefer you do it as I have other things to do on other FOSS projects. 

Thanks for your work! I'll put it on draft for the meantime and close it once you have a PR up.

PR #3732, opened 2024-04-07
Adjust bonus system to incentivize seeding torrents that help the tracker the most

devedse:
### Introduction
I've noticed that the current bonus system encourages a specific way of seeding—for example, older, ~1GB whereas other torrent's award less points. This kind of strategy, or "meta," means most people could chase seeding the same torrents to try to get the most points. In my opinion we should go back to the basics of a tracker where the simple goal should be to have a **wide range of (high quality) torrents alive and available with fast download speeds**.

I'd like to share a new BON algorithm that incentivizes seeding torrents that help accomplish this goal.

### My Suggestions
Here's what I'm proposing to make seeding different kinds of torrents more appealing:

To encourage a broader seeding strategy and to value contributions across all torrent sizes and types, I've devised a bonus algorithm that rewards the scarcity of seeders as well as compensates for the bandwidth and storage resources utilized by seeders. Here's an outline of the proposed changes:

Rewarding Seeder Scarcity
The cornerstone of the new algorithm is to reward torrents based on the number of active seeders, aiming to incentivize seeding where it's needed most. The idea is simple: the fewer the seeders, the higher the bonus. This directly supports our goal of keeping a wide range of torrents alive.

| Seeder Count | Bonus Points |
|--------------|--------------|
| 1 Seeder | 10 Points |
| 2 Seeders | 5 Points |
| 3-5 Seeders | 3 Points |
| 6-9 Seeders | 2 Points |
| 10-19 Seeders | 1.5 Points |
| 20-35 Seeders | 1 Point |
| 36-49 Seeders | 0.75 Points |
| 50+ Seeders | 0.5 Points |

Reward for Torrent Size
Alongside seeder scarcity, the algorithm introduces a size-based reward system. While seeding 1GB torrents remains important and is still rewarded, the algorithm acknowledges the additional bandwidth and storage resources required for larger torrents. Thus, a logarithmic scale multiplier is applied to ensure that larger torrents yield a higher bonus, but without overshadowing the importance of seeding smaller torrents.

| Torrent Size (GB) | Calculation                              | Size Multiplier (approximate) |
|--|--|--|
| 1 | (log(1) + 1) * 0.5 | 0.5 |
| 10 | (log(10) + 1) * 0.5 | 1.65 |
| 20 | (log(20) + 1) * 0.5 | 2 |
| 50 | (log(50) + 1) * 0.5 | 2.45 |
| 100 | (log(100) + 1) * 0.5 | 2.80 |

The formula used for calculating the size multiplier is (log(sizeInGB) + 1) * 0.5, ensuring a balanced increase in rewards that reflects the contribution of seeding larger files without detracting from the value of smaller torrents.

Overall Impact
The proposed algorithm is designed to address the current meta by diversifying the types of torrents that are seen as valuable to seed, thus supporting a healthier, more robust torrent ecosystem. By rewarding both the scarcity of seeders and the size of torrents, we aim to cover a broader spectrum of seeding activities:

Smaller torrents (≥1GB but <5GB) are still important, offering quick downloads and requiring less commitment in terms of bandwidth and storage.
Larger torrents (≥25GB), which typically include high-quality media content, are given additional incentives to reflect the higher resource commitment from seeders, thereby ensuring these files remain readily available and healthy.
Conclusion
This revised BON algorithm is just a proposal based on my observations and understanding of our community's needs. It's crafted with the intention to steer us towards the fundamental goal of a torrent tracker: keeping a diverse range of high-quality torrents alive and ensuring fast download speeds. I believe this approach will encourage a more balanced seeding behavior, rewarding contributions across the entire spectrum of torrent sizes and helping maintain the vitality of our torrent library.

### Developer notes
I've not tested this code so please take a good look at everything before deciding to include this (or not :smile:).

I'm looking forward to hearing the community's thoughts on this proposed change.

### DIFF
--- a/app/Console/Commands/AutoBonAllocation.php
+++ b/app/Console/Commands/AutoBonAllocation.php
@@ -42,211 +42,31 @@ class AutoBonAllocation extends Command
      */
     public function handle(ByteUnits $byteUnits): void
     {
-        $dyingTorrent = DB::table('peers')
-            ->select(DB::raw('count(DISTINCT(peers.torrent_id)) as value'), 'peers.user_id')
-            ->join('torrents', 'torrents.id', 'peers.torrent_id')
-            ->where('torrents.seeders', 1)
-            ->where('torrents.times_completed', '>', 2)
-            ->where('peers.seeder', 1)
-            ->where('peers.active', 1)
-            ->whereRaw('date_sub(peers.created_at,interval 30 minute) < now()')
-            ->groupBy('peers.user_id')
-            ->get()
-            ->toArray();
-
-        $legendaryTorrent = DB::table('peers')
-            ->select(DB::raw('count(DISTINCT(peers.torrent_id)) as value'), 'peers.user_id')
+        $peersWithTorrents = DB::table('peers')
+            ->select('peers.user_id', 
+                     'torrents.id as torrent_id', 
+                     'torrents.size', 
+                     'torrents.seeders as seeder_count')
             ->join('torrents', 'torrents.id', 'peers.torrent_id')
             ->where('peers.seeder', 1)
             ->where('peers.active', 1)
-            ->whereRaw('torrents.created_at < date_sub(now(), interval 12 month)')
-            ->whereRaw('date_sub(peers.created_at,interval 30 minute) < now()')
-            ->groupBy('peers.user_id')
-            ->get()
-            ->toArray();
+            ->whereRaw('date_sub(peers.created_at, interval 30 minute) < now()')
+            ->get();
 
-        $oldTorrent = DB::table('peers')
-            ->select(DB::raw('count(DISTINCT(peers.torrent_id)) as value'), 'peers.user_id')
-            ->join('torrents', 'torrents.id', 'peers.torrent_id')
-            ->where('peers.seeder', 1)
-            ->where('peers.active', 1)
-            ->whereRaw('torrents.created_at < date_sub(now(), Interval 6 month)')
-            ->whereRaw('torrents.created_at > date_sub(now(), interval 12 month)')
-            ->whereRaw('date_sub(peers.created_at,interval 30 minute) < now()')
-            ->groupBy('peers.user_id')
-            ->get()
-            ->toArray();
+        $bonAllocations = [];
 
-        $hugeTorrent = DB::table('peers')
-            ->select(DB::raw('count(DISTINCT(peers.torrent_id)) as value'), 'peers.user_id')
-            ->join('torrents', 'torrents.id', 'peers.torrent_id')
-            ->where('peers.seeder', 1)
-            ->where('peers.active', 1)
-            ->where('torrents.size', '>=', $byteUnits->bytesFromUnit('100GiB'))
-            ->whereRaw('date_sub(peers.created_at,interval 30 minute) < now()')
-            ->groupBy('peers.user_id')
-            ->get()
-            ->toArray();
-
-        $largeTorrent = DB::table('peers')
-            ->select(DB::raw('count(DISTINCT(peers.torrent_id)) as value'), 'peers.user_id')
-            ->join('torrents', 'torrents.id', 'peers.torrent_id')
-            ->where('peers.seeder', 1)
-            ->where('peers.active', 1)
-            ->where('torrents.size', '>=', $byteUnits->bytesFromUnit('25GiB'))
-            ->where('torrents.size', '<', $byteUnits->bytesFromUnit('100GiB'))
-            ->whereRaw('date_sub(peers.created_at,interval 30 minute) < now()')
-            ->groupBy('peers.user_id')
-            ->get()
-            ->toArray();
-
-        $regularTorrent = DB::table('peers')
-            ->select(DB::raw('count(DISTINCT(peers.torrent_id)) as value'), 'peers.user_id')
-            ->join('torrents', 'torrents.id', 'peers.torrent_id')
-            ->where('peers.seeder', 1)
-            ->where('peers.active', 1)
-            ->where('torrents.size', '>=', $byteUnits->bytesFromUnit('1GiB'))
-            ->where('torrents.size', '<', $byteUnits->bytesFromUnit('25GiB'))
-            ->whereRaw('date_sub(peers.created_at,interval 30 minute) < now()')
-            ->groupBy('peers.user_id')
-            ->get()
-            ->toArray();
-
-        $participaintSeeder = DB::table('history')
-            ->select(DB::raw('count(*) as value'), 'history.user_id')
-            ->where('history.active', 1)
-            ->where('history.seedtime', '>=', 2_592_000)
-            ->where('history.seedtime', '<', 2_592_000 * 2)
-            ->groupBy('history.user_id')
-            ->get()
-            ->toArray();
-
-        $teamplayerSeeder = DB::table('history')
-            ->select(DB::raw('count(*) as value'), 'history.user_id')
-            ->where('history.active', 1)
-            ->where('history.seedtime', '>=', 2_592_000 * 2)
-            ->where('history.seedtime', '<', 2_592_000 * 3)
-            ->groupBy('history.user_id')
-            ->get()
-            ->toArray();
-
-        $commitedSeeder = DB::table('history')
-            ->select(DB::raw('count(*) as value'), 'history.user_id')
-            ->where('history.active', 1)
-            ->where('history.seedtime', '>=', 2_592_000 * 3)
-            ->where('history.seedtime', '<', 2_592_000 * 6)
-            ->groupBy('history.user_id')
-            ->get()
-            ->toArray();
-
-        $mvpSeeder = DB::table('history')
-            ->select(DB::raw('count(*) as value'), 'history.user_id')
-            ->where('history.active', 1)
-            ->where('history.seedtime', '>=', 2_592_000 * 6)
-            ->where('history.seedtime', '<', 2_592_000 * 12)
-            ->groupBy('history.user_id')
-            ->get()
-            ->toArray();
-
-        $legendarySeeder = DB::table('history')
-            ->select(DB::raw('count(*) as value'), 'history.user_id')
-            ->where('history.active', 1)
-            ->where('history.seedtime', '>=', 2_592_000 * 12)
-            ->groupBy('history.user_id')
-            ->get()
-            ->toArray();
-
-        //Move data from SQL to array
-
-        $array = [];
-
-        foreach ($dyingTorrent as $value) {
-            if (\array_key_exists($value->user_id, $array)) {
-                $array[$value->user_id] += $value->value * 2;
+        foreach ($peersWithTorrents as $peerSeedingTorrent) {
+            $seederBonus = calculateSeederBonus($peerSeedingTorrent->seeder_count);
+            $sizeInGB = ceil($byteUnits->toGigabytes($peerSeedingTorrent->size));
+            $totalBON = $seederBonus * $sizeInGB;
+        
+            // Accumulate bonuses per user
+            if (isset($bonAllocations[$peerSeedingTorrent->user_id])) {
+                $bonAllocations[$peerSeedingTorrent->user_id] += $totalBON;
             } else {
-                $array[$value->user_id] = $value->value * 2;
+                $bonAllocations[$peerSeedingTorrent->user_id] = $totalBON;
             }
-        }
-
-        foreach ($legendaryTorrent as $value) {
-            if (\array_key_exists($value->user_id, $array)) {
-                $array[$value->user_id] += $value->value * 1.5;
-            } else {
-                $array[$value->user_id] = $value->value * 1.5;
-            }
-        }
-
-        foreach ($oldTorrent as $value) {
-            if (\array_key_exists($value->user_id, $array)) {
-                $array[$value->user_id] += $value->value * 1;
-            } else {
-                $array[$value->user_id] = $value->value * 1;
-            }
-        }
-
-        foreach ($hugeTorrent as $value) {
-            if (\array_key_exists($value->user_id, $array)) {
-                $array[$value->user_id] += $value->value * 0.75;
-            } else {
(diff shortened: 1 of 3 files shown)


HDVinnie:
@Audionut @MiM-MiM @Roardom

Roardom:
A few notes:

- How long does it take to run this code on a site with a few million active peers and a few hundred thousand torrents?
- seedtime / type_id / resolution_id / internal / personal_release and maybe other torrent flags should be relevant, and customizable (different multipliers) per site.
- See prior work done on this: https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/2411. Its feature set has been discussed by multiple sites and everyone has agreed with it so far, but it needs some work to make sure the code takes ideally less than 30 seconds on large sites.

Audionut:
Replacing 1 hardcoded solution with another hardcoded solution is not ideal.

I'm also not a fan of the numbers. Anything with 6+ seeders is effectively safe. I personally would consider that the baseline (1 point), with an upper limit of 5 points.

Roughly, using the points distribution outlined, my BON rate would increase by about 10%, without any seedtime bonus.
IMO, seedtime is an important factor, encourages long term seeding which is another important aspect to longevity of data, and must be rewarded.

I'd rather see the seeder count distribution reduced too:
5
3
2
1
0.75
0.5

The above, combined with the torrent size distribution outlined here, and the current seedtime bonus would see my BON rate increase by around 11%.

A 10% increase against the current earnings is probably too much, since ideally long running sites should maintain an equal points distribution, but probably only really reflects my seeding habits, rather than the average of the entire user base. And it's better to err on the site of more points, rather than less points IMO. Stores can be edited to reflect increased earning rates.

As mentioned by Roardom, also needs added flags for particular torrents.

devedse:
> A few notes:
> 
> 1. How long does it take to run this code on a site with a few million active peers and a few hundred thousand torrents?
> 1. seedtime / type_id / resolution_id / internal / personal_release and maybe other torrent flags should be relevant, and customizable (different multipliers) per site.
> 1. See prior work done on this: [(Update) Make bonus point allocation more customizable #2411](https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/2411). Its feature set has been discussed by multiple sites and everyone has agreed with it so far, but it needs some work to make sure the code takes ideally less than 30 seconds on large sites.

1. Currently I'm not sure, based on the formula's I'm doing I expect I could run the full query in SQL though. E.g. by doing something like this:
```
INSERT INTO Users (user_id, seedbonus)
SELECT 
    p.user_id,
    SUM(
        -- Seeder bonus calculation
        CASE
            WHEN t.seeder_count = 1 THEN 10
            WHEN t.seeder_count = 2 THEN 5
            WHEN t.seeder_count BETWEEN 3 AND 5 THEN 3
            WHEN t.seeder_count BETWEEN 6 AND 9 THEN 2
            WHEN t.seeder_count BETWEEN 10 AND 19 THEN 1.5
            WHEN t.seeder_count BETWEEN 20 AND 35 THEN 1
            WHEN t.seeder_count BETWEEN 36 AND 49 THEN 0.75
            ELSE 0.5
        END
        *
        -- Size calculation and scaling
        (LOG(GREATEST(CEIL(t.size / 1073741824), 1)) + 1) * 0.5
    ) AS bonus_points
FROM 
    peers AS p
JOIN 
    torrents AS t ON t.id = p.torrent_id
WHERE 
    p.seeder = 1
    AND p.active = 1
    AND p.created_at > NOW() - INTERVAL 30 MINUTE
GROUP BY 
    p.user_id
ON DUPLICATE KEY UPDATE
    seedbonus = seedbonus + VALUES(seedbonus);
```
2. See point 3
3. I agree customizability would be nice. It seems the PR was closed though. Some thoughts:
I'm normally a C# developer and what you can do there is actually write queries using C# (e.g. dynamically add conditions and all) and still have them executed on the database. If that would be possible in PHP as well I think the issues of running 30 seconds could be avoided. And by the way, this is also a PR that wasn't merged for 2 years so I think it leaves some room for other ideas (else it probably would've been done :) )

devedse:
> Replacing 1 hardcoded solution with another hardcoded solution is not ideal.

Agree, let's see if we can come up with something that's feasible. Or if simply taking 1 step into a direction would already help (e.g. being able to select between algorithm 1 and 2 for the BON allocation)

> I'm also not a fan of the numbers. Anything with 6+ seeders is effectively safe. I personally would consider that the baseline (1 point), with an upper limit of 5 points.

Sure we can modify them. (or make them configurable in a bit of an intermediate solution before becoming fully dynamic)

> Roughly, using the points distribution outlined, my BON rate would increase by about 10%, _**without any seedtime bonus**_. IMO, seedtime is an important factor, encourages long term seeding which is another important aspect to longevity of data, and must be rewarded.

I personally disagree with you here. I'd much rather have someone delete a torrent and pickup one that's not seeded well then to have that specific person effectively be stuck with seeding the same torrent even though there's already 20 other people doing the same. I think this specific criteria insentivizes meta play rather then what we actually need, and that is focus on torrents not being seeded. (Again, just my opinion)

> I'd rather see the seeder count distribution reduced too: 5 3 2 1 0.75 0.5

Sure we can change that

> The above, combined with the torrent size distribution outlined here, and the current seedtime bonus would see my BON rate increase by around 11%.
> 
> A 10% increase against the current earnings is probably too much, since ideally long running sites should maintain an equal points distribution, but probably only really reflects my seeding habits, rather than the average of the entire user base. And it's better to err on the site of more points, rather than less points IMO. Stores can be edited to reflect increased earning rates.

I personally think having a BON increase wouldn't matter too much. Assuming I'd want to use let's say 25gb per day in downloads. I'd currently need around a 1000 40gb torrents being seeded. Which feels a bit stacked against me. I think the website would be healthy if all torrents are seeded with 3-5 people. If we can manage I think everyone should get enough BON to do whatever they want since our goal is effectively solved.

> As mentioned by Roardom, also needs added flags for particular torrents.

I can't think of some examples on why you'd need these in the BON calculation, but maybe you can provide me some example earnings that would make sense to add to my idea.

Audionut:
> Or if simply taking 1 step into a direction would already help

IMO, it doesn't. We don't want to be changing BON earnings on an on-going basis. Any refactor of the BON system should satisfy all requirements going forward. + for https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/2411

> I personally disagree with you here. I'd much rather have someone delete a torrent and pickup one that's not seeded well then to have that specific person effectively be stuck with seeding the same torrent even though there's already 20 other people doing the same. I think this specific criteria insentivizes meta play rather then what we actually need, and that is focus on torrents not being seeded. (Again, just my opinion)

I think you're conflating 2 issues here. Granting a seedtime bonus does not force anyone to do anything. It simply incentivizes long term seeding. Look at recruitment threads of many trackers these days and you will see..."_Seedboxes are not required (there's already plenty) to keep a good ratio at PTP. We prefer long term archivist seeders_", because long term seeding is important to the community, more importantly than a users ability to generate BON.

HnR's effectively force users to seed for x amount of time, in an effort to have people continue to seed. If sites thought they could get away with it (without alienating their userbase), they would force HnR rules for 12+ months.

Ideally, people wouldn't download stuff for the sole purpose of ratio or BON, but here we are.....

> I can't think of some examples on why you'd need these in the BON calculation,

Other that what @Roardom already listed, there's also, :is_raw. There's 2 main points too https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/2411

1.  It fully satisfies the "Community" aspect of "UNIT3D-Community-Edition. It gives sites the control over BON earnings, _### however they want_, rather than being forced to use what 1, or 2 or 5 devs consider is appropriate.
2. It gives control over many parameters.

Anything less is meh. I gave an opinion on numbers above in case others disagree.

devedse:
> I think you're conflating 2 issues here. Granting a seedtime bonus does not force anyone to do anything. It simply incentivizes long term seeding. Look at recruitment threads of many trackers these days and you will see..."Seedboxes are not required (there's already plenty) to keep a good ratio at XXX. We prefer long term archivist seeders", because long term seeding is important to the community, more importantly than a users ability to generate BON.
> 
> HnR's effectively force users to seed for x amount of time, in an effort to have people continue to seed. If sites thought they could get away with it (without alienating their userbase), they would force HnR rules for 12+ months.

But would it matter who seeds a torrent? If you increase the ratio's for torrents with a low seeder amount you'd reward users for picking up torrents that are dying. I think this would accomplish the same goal as long term seeding bonus without the (maybe) unwanted side effect of users just choosing some torrents and sticking with those forever, even though they are already well seeded.

I agree with all other points, I'm just curious if there's a plan forward to implement this.

Audionut:
> But would it matter who seeds a torrent? If you increase the ratio's for torrents with a low seeder amount you'd reward users for picking up torrents that are dying. I think this would accomplish the same goal as long term seeding bonus without the (maybe) unwanted side effect of users just choosing some torrents and sticking with those forever, even though they are already well seeded.

You seem to be still conflating that long term seeds = seeds with many peers.

You want to reward/incentivize peers for seeds with low seeder counts. That is a given we both agree on.

You don't want to _obsessively_ incentivize single seed peers. This leads to collusion/cheating, whatever you want to call it. Ideally a seed should have peers+n.

Without a seedtime bonus, there is nothing stopping a single seed from dropping torrent X and picking up torrent Y, leaving torrent X dead. A seedtime bonus gives extra that is only rewarded with patience.

Peer counts rise and fall, and rise and fall. What ultimately keeps seeds alive is long term seeders. Without long term seeders, there is an increased risk of seeds=0 during the falls.

There's more involvement to torrenting than BON alone. Long term seeds, regardless of peer counts, keep torrents alive.

I don't like relying on appeal to authority, but to add to the argument.....
> Look at recruitment threads of many trackers these days and you will see..."...We prefer long term archivist seeders".


> Assuming I'd want to use let's say 25gb per day in downloads. I'd currently need around a 1000 40gb torrents being seeded. Which feels a bit stacked against me.

On one particular site, I have ~600 torrents earning ~1200 points from long term seeding bonus. 1200 * 24 = 28,800/day.
The same site allows me to buy 25 GiB Upload for 20,000 points.

Why so many points......Because that is reliable. Regardless of the torrent size, they could all be 1 Mb torrents if I was a BON wanker, this demonstrates reliability. Long term seeders can be relied upon to remain as a seeder when peer counts drop. Long term seeders can be relied upon to contribute to bandwidth, for the length of time the seed is active.

The assumption, and I'm going to paraphrase/emphasize here, "I want enough points to download whatever I want whilst only seeding 1 torrent, and I'll only seed that 1 torrent when I'm the only peer", is arrogant.

That 25 GiB of data, with the associated bandwidth to deliver it, was contributed by others. A community is built by others also contributing data and bandwidth, not by stacking the cards so in favor of a lack of effort, that no one puts in effort. Another reason why we have long term seeding bonuses, is rewarding reliable users to seed torrents forever, who otherwise couldn't survive with slow bandwidths. "_If you continue to seed this torrent, we will give you free bandwidth (being able to download from others) without having to contribute that same bandwidth back to the community._"

Low peer count bonuses are effectively, "_OH shit, this torrent might die, please accept this **extra reward** to keep this alive._"

devedse:
I think we agree largely, I might have made the simplification that simply throwing more BON to unseeded torrents might get people interested in them, but maybe giving bonusses to keep long term seeders simply doesn't make people drop them.

Anyway, with all this discussion I'm curious if there's still a plan on improving / rewriting the BON engine?

Audionut:
> Anyway, with all this discussion I'm curious if there's still a plan on improving / rewriting the BON engine?

> * See prior work done on this: https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/2411  Its feature set has been discussed by multiple sites and everyone has agreed with it so far, but it needs some work to make sure the code takes ideally less than 30 seconds on large sites.

Make that, or something extremely similar, take less than 30 seconds. Other than people's priorities changing, for instance, site security trumps stupid BON, and people's free time changing, the fact that it hasn't been pushed in ~2 years only demonstrates it's not as 'easy fix'.

devedse:
It was honestly not my intention to question you guys, it was an honest question if the BON system was still on your backlog. I'm sorry if it made you feel that way 😢 .

Keep me up to date incase something changes.

HDVinnie:
> It was honestly not my intention to question you guys, it was an honest question if the BON system was still on your backlog. I'm sorry if it made you feel that way 😢 .
> 
> 
> 
> Keep me up to date incase something changes.

You didn't upset anyone. Pull requests are always welcome. Your idea is good but as linked in old pull request it would be much better to make bon system fully configurable so different sites can use different rates they want. It's def a must needed feature as current hardcode is not practical. There just so much to do and we have been working on many other things. A project like this is never ending.

Roardom:
Feel free to use my old code in that PR. I had a dozen draft PRs open that were all closed when v8 was released. All of those PRs were unfinished, but desired features. I generally work on 20+ branches at a time, and if I can't finish a feature in one night, I push a draft PR so it's not forgotten, work on something else the next night and eventually :tm: come back to it from time to time until it's finished.

Also, Laravel's ORM offers what you suggest if I understand correctly. The official documentation is here: https://laravel.com/docs/11.x/eloquent, but the full reference docs can be found at https://laravel.com/api/11.x/Illuminate/Database/Eloquent/Builder.html, https://laravel.com/api/11.x/Illuminate/Database/Query/Builder.html, and maybe a few other classes.

BitPyrate:
I personally not a fan of the bon system. But i ways earned my upload raw dogging and keeping a good ratio. Bin sounded extra steps and complicated gamey. We honestly were about to scrap the bon system entirely but user feed back stopped us. We now embrace it and have learned to love it. 

I am not specifically interested in this change listed in this pull request. But love thinking about the game theory.
Incentivize . 1 seed torrents will get gamed.   But not giving dying the incentive  but rather giving the incentive where seed is 2-5 is where it might get more interesting.  A person will seed solo without any bon incentive knowing that it becomes an incentive for person 2 to join.   Other people will be on the look out for 1 seed to upgrade it to something that gives a bonus. Harder to game as it premotes group effort.  And beacuse 1seed is good bait there is little incentive to try and game using a second account.  It is fueled by greed of fellow humans

PR #3738, opened 2024-04-08
(Fix) Helpdesk Route

HDVinnie:


Roardom (approved):
Few questions:

- Did this route error when accessed or did it magically bypass the middleware
- If it magically bypassed, what can we do to prevent this from happening again?
  - Class strings
  - Write a phpstan dynamic return type extension
  - Other?

HDVinnie:
> Few questions:
> 
> * Did this route error when accessed or did it magically bypass the middleware
> * If it magically bypassed, what can we do to prevent this from happening again?
>   
>   * Class strings
>   * Write a phpstan dynamic return type extension
>   * Other?

It threw an error that is_modo didn't exist. 

```
Target class [is_modo] does not exist. at /var/www/html/vendor/laravel/framework/src/Illuminate/Container/Container.php:914)
```

This is because it doesn't. The alias is just `modo`. https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/3eb9db353099e79998f18067daf2f9b323f971b4/app/Http/Kernel.php#L97

Using an alias is fine or could use class. See https://laravel.com/docs/10.x/middleware#assigning-middleware-to-routes

None the less it's fixed now and there was also check in controller so wasn't a security issue. https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/3eb9db353099e79998f18067daf2f9b323f971b4/app/Http/Controllers/TicketController.php#L90

PR #3739, opened 2024-04-08
(Update) AutoCacheRandomMediaIds Command

maxwbot:
Is the new command (Depreciate command);


Roardom:
Can you explain your reasoning?

This change makes it less obvious of what's going on as it uses the "magic" method, that ends up doing the same thing:

- https://github.com/laravel/framework/blob/81d8154874400f2625553a606c2d801d0f014a9a/src/Illuminate/Redis/Connections/PhpRedisConnection.php#L562
- https://github.com/laravel/framework/blob/81d8154874400f2625553a606c2d801d0f014a9a/src/Illuminate/Redis/Connections/Connection.php#L223

Also, all other places in the codebase that use the Redis facade use this convention already of calling `->command()` directly. This benefits the instances in critical paths as it prevents extra code from being ran.

PR #3741, opened 2024-04-08
(Update) GroupRequest

maxwbot:
Resolved problem where not selecting the checkbox resulted in an empty or unfilled field error.


Roardom (changes requested):
Please follow the code style guidelines. It looks like you replaced all the spaces with tabs.

Also, please use laravel's validation rules. You can find the documentation here: https://laravel.com/docs/validation. Ctrl+f for `conditional`.

Issue #3650, opened 2024-03-13
[Request] Option to select multiple torrents from main torrent page


timmy171297:
add option to select multiple torrents directly from the torrents page as staff/admin

eg. just add an option similar to when you press similar torrents to mass delete. we need this for all torrents. So you can select all from main page and mark as free leech, x2, delete them or move them to a different category and more.... currently you can only do this 1 by 1.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/HDInnovations) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/HDInnovations/UNIT3D-Community-Edition/issues/3650">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/3650/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/3650/pledge.svg">
</picture>
</a>

HDVinnie:
This request is complete and under testing/cleanup and will be part of next release v8.0.3

PR #3744, opened 2024-04-09
(Add) Request #3650

HDVinnie:


Roardom (changes requested):
Just a quick review.

The suggestions on the freeleech function apply to the other buffs as well.

PR #3751, opened 2024-04-13
(Update) Autogroup - require 2fa to enable invite

Audionut:
Try to fix user profile showing invite enabled when 2fa is not enabled.
image


HDVinnie (approved):

Roardom (changes requested):
I don't like this.

There's already multiple issues with can_download being affected by multiple things (both warned and low ratio save their state in the same field) which causes bugs where users can still download torrents during edge cases even if they have more than 3 warnings.

Currently, can_download is the only majorly bugged permission. To fix permissions, can_download has to be split into multiple permissions with all relevant permissions being checked upon the desired action.

Having can_invite also merges 2 explicitly different conditions that should be saved individually. The conditions can both be checked upon the relevant action, but should not be saved together. Otherwise, it just cases bugs.

For example, ctrl+f for the 2 other locations in the codebase that this variable is set to true and realize you've only changed one of them. Not only that, but some of the locations might not be relevant to 2fa.

2fa is already required to invite, but it checks if 2fa is enabled directly instead of saving it on the user table. So, copy the logic and check if 2fa is enabled directly when displaying a user's permissions on their profile page. Or, if you want to be extra fancy and make sure there's no code duplication between showing the profile permissions and allowing an invite, a computed attribute on the user model would work as well to share the code between both locations.

Audionut:
I'm not seeing anywhere else in the code where `can_invite = true`, when searching for `can_invite`. There are a couple of instances where `can_invite=1`.

Certainly, it seems appropriate to set `can_invite=0` in the unbancontroller, letting autogroup sort out the permissions.

The existing code from the invitecontroller is exactly to logic I copied, and if invite ability is being controlled elsewhere, other than via autogroup, why....... Why on earth would you have some setting being set somewhere, and then attempting to be set every 24 hours by some other thing.

Audionut:
Does this work better instead?
https://github.com/Audionut/UNIT3D-Community-Edition/commit/8a122925dfac7c4dbdbe1ec683e53a476513faa8

Roardom:
> I'm not seeing anywhere else in the code where can_invite = true, when searching for can_invite. There are a couple of instances where can_invite=1.

mysql will automatically save a `1` do the database when you use either `1` or `true`. They can both be considered the same in this case. Autogroup's can_invite state, and the "is 2fa enabled" can_invite state should not be merged when saving, but they should both be compared when displaying the user's currently enabled/disabled permissions, or when validating the invite being sent.

> Why on earth would you have some setting being set somewhere, and then attempting to be set every 24 hours by some other thing

This is why the permissions are broken. Multiple things are changing them, varying from staff manually changing it, to autogroup, to when a user is re-enabled. Ideally, every condition should be saved individually and then all applicable saved conditions/states should be checked when the combined "can this user invite" condition is needed. In this case, 2fa and autogroup's can_invite, and the user's can_invite are all saved individually which means we don't have an issue with it currently. All that's needed is an extra check to determine if the green checkmark or red X is shown

Roardom:
> Does this work better instead? [Audionut@8a12292](https://github.com/Audionut/UNIT3D-Community-Edition/commit/8a122925dfac7c4dbdbe1ec683e53a476513faa8)

```php
                        @if ($user->can_invite == 1 && user->two_factor_confirmed_at !== null)
```

Audionut:
I thought re-basing might have flattened the commits somewhat. I can push a new, cleaner PR if that suits better Vinnie.

Roardom:
> I thought re-basing might have flattened the commits somewhat. I can push a new, cleaner PR if that suits better Vinnie.

You can do `git rebase HEAD~5 -i` to interactively rebase the last 5 commits, then edit the file that pops up in your terminal. Edit all the `pick`s with `fixup`s except for the first commit which should remain as `pick`. Then depending on if your default text editor (assuming vi is default), press escape, then `:x`and enter to save and exit (if it's nano, then just press ctrl+x, and follow the instructions for saving). This will combine all the commits into only the oldest commit. Alternatively, you can also do the same with `squash` instead of `fixup` to combine all the commits into the newest commit.

Messy git history bugs me a bit (not a blocker to the PR), but I wish someone was around to teach me it in the beginning.

Audionut:
Made a mess of this, lets start fresh.

PR #3753, opened 2024-04-13
(Add) Missing media search

Obi-Wana:
Allow search for media title (as TMDB refers to it) and/or release year.

### DIFF
--- a/resources/views/livewire/missing-media-search.blade.php
+++ b/resources/views/livewire/missing-media-search.blade.php
@@ -1,10 +1,45 @@
 <section class="panelV2">
-    <h1 class="panel__heading">Missing Media</h1>
+    <header class="panel__header">
+        <h2 class="panel__heading">Missing Media</h2>
+        <div class="panel__actions">
+            <div class="panel__action">
+                <div class="form__group">
+                    <input
+                        id="name"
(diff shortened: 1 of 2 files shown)


HDVinnie (approved):

resources/views/livewire/missing-media-search.blade.php
```diff
@@ -1,10 +1,42 @@
...
+                        {{ __('torrent.title') }}
+                    </label>
+                </div>
+            </div>
+            <div class="panel__action">
+                <div class="form__group">
+                    <input
+                        id="year"
+                        class="form__text"
+                        type="number[4]"
```
Roardom, line 24: Typo here. Also, try and use text types instead of numeric to avoid the value incrementing/decrementing when scrolled over.
```
type="text"
class="form__text"
inputmode="numeric"
pattern="[0-9]*"
```

Roardom (changes requested):

resources/views/livewire/missing-media-search.blade.php
Obi-Wana: Updated.

Roardom (approved):

PR #3755, opened 2024-04-14
(Fix) Group Validation

HDVinnie:


app/Http/Requests/Staff/StoreGroupRequest.php
```diff
@@ -112,29 +113,39 @@ public function rules(): array
                 'boolean',
             ],
             'min_uploaded' => [
-                'sometimes',
-                'integer',
-                'min:0',
+                Rule::when((bool) $request->integer('autogroup'), [
```
Roardom, line 116: Directly use the boolean helper instead of casting to int first.

```php
$request->boolean('autogroup')
```

Also, prohibit the field if it's not autogroup

Roardom (changes requested):

PR #3759, opened 2024-04-17
(Refactor) Change input type for better 2FA inference

EkoNesLeg:
This simple change improves passwd manager detection for 2FA fields, ensuring auto clipboard copying or user prompt popups. It also improves mobile usage by calling the numeric keypad for the tel-set 2FA input field, rather than a full keyboard.
Tested on ProtonPass; with input type set to `text`, no clipboard copying or popup occurs, requiring manual copy-pasting. But with type `tel`, 2FA autofill features are fully activated.

### DIFF
--- a/resources/views/auth/two-factor-challenge.blade.php
+++ b/resources/views/auth/two-factor-challenge.blade.php
@@ -81,7 +81,7 @@
                             inputmode="numeric"
                             name="code"
                             x-bind:required="!recovery"
-                            type="text"
+                            type="tel"
                             value="{{ old('code') }}"
                             x-ref="code"
                         />
@@ -96,7 +96,7 @@
                             autocomplete="one-time-code"
                             name="recovery_code"
                             x-bind:required="recovery"
-                            type="text"
+                            type="tel"
                             x-ref="recovery_code"
                         />
                     </p>


Roardom (commented):
I'm curious how you managed to find this information and if it's documented anywhere, or if it's a protonpass bug? I don't see how it makes sense for protonpass to whitelist this field but not the other attributes already included on the form fields. Have you tried other types? Perhaps they're instead blacklisting `type="text"`

EkoNesLeg:
`tel` is standard practice for 2FA. It can be refined further with constraints such as autocorrect=off and spellcheck=off. 
I do need to edit the recovery code one though, since that is alphanumeric, and it doesn't need this capability either.

EkoNesLeg:
Regarding documentation, there isn't much, so initially, I modeled it after Proton's implemenatiaon. But given your claims about Protonpass having bugs, I checked alternatives:

Google's:
```html
<input type="tel" pattern="[0-9 ]*" class="input-class" autocomplete="off" spellcheck="false" tabindex="0" aria-label="Enter code" name="totpPin" autocapitalize="off" id="totpPin" autocorrect="off" dir="ltr">
```

Microsoft's:
```html
<input data-bind="attr: {'maxlength': 'otcLength', 'aria-labelledby': 'idDiv_SAOTCC_Title', 'aria-describedby': 'idDiv_SAOTCC_Description idSpan_SAOTCC_Error_OTC'}, css: {'has-error': 'error'}, textInput: 'otcInputTextbox.value', ariaLabel: 'CT_SAOTCC_STR_OTC_TBHint', hasFocusEx: 'otcInputTextbox.focused', placeholder: '$placeholderText'">
```

Indeed, it would be wise to also add the autocaps and autofill constraints as suggested earlier.

EkoNesLeg (commented):
Remove `tel` from the recovery-code form as codes are alphanumeric and passwd manager compatibility is not relevant.

Added additional contraints to the input elements to prevent auto capping, autocorrect, spellcheck. Also added aria element for screenreader compatibility.

Roardom (changes requested):
Good idea of inspecting the html on those large websites 👍. I can see they use `type="tel"`. I'm surprised this isn't more widely documented.

Approved except can you remove some of the duplicate values, I think the original `autocomplete` value of `one-time-code` is more appropriate as well. The aria stuff also isn't needed because of the label that references the input's id and the first rule of aria is to not use aria.

Roardom (approved):

PR #3778, opened 2024-04-25
(Update) Rate limit IP's

Audionut:
Limit to 1 per minute.

Slowing down legitimate users seems appropriate also, as they may take their time instead of spamming the wrong values. Ideally the rate limiting would further reduce the limits per minute with continued failed attempts.

### DIFF
--- a/app/Providers/FortifyServiceProvider.php
+++ b/app/Providers/FortifyServiceProvider.php
@@ -142,15 +142,15 @@ class FortifyServiceProvider extends ServiceProvider
      */
     public function boot(): void
     {
-        RateLimiter::for('login', fn (Request $request) => Limit::perMinute(5)->by('fortify-login'.$request->ip()));
-        RateLimiter::for('fortify-login-get', fn (Request $request) => Limit::perMinute(5)->by('fortify-login'.$request->ip()));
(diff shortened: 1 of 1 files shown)


Roardom:
To login you need 4 requests:

- To view the login page
- To submit your user and pass
- To view the 2fa page
- To send the 2fa code

It will thus take 4 minutes to sign in, and you need to wait 1 minute between each of these actions.

Roardom:
Actually five, if you have unit3d.dev bookmarked and not the permanent redirect to unit3d.dev/login.

There's also the issue of if you mistype your password wrong once, you need to wait a minute for the rate limit to reset

Roardom (commented):
You'll have to key each of these by a unique name inside the `->by()` method otherwise the rate limiters get incremented multiple times for each of these actions. Also, you have to balance this and the significant amount of support issues it generates, especially when a tracker switches frameworks and requires users to reset their password.

Audionut:
I expected each individual request to be it's own limit, not tied overall to each other request. I'll see if I can modify as per your latest comment.

> There's also the issue of if you mistype your password wrong once, you need to wait a minute for the rate limit to reset

I don't see that as an issue at all. Password managers are best practice, and it's asking for only < 60 seconds of patience.

PR #3782, opened 2024-04-26
(Fix) Bug #3752

HDVinnie:


resources/views/livewire/user-notes.blade.php
```diff
@@ -92,6 +92,7 @@ class="form__textarea"
                                                     name="message"
                                                     placeholder=" "
                                                     wire:model="messages.{{ $note->id }}"
+                                                    value="{{ $note->message }}"
```
Roardom, line 95: I recall there being bugs when I did this initially when editing user notes. Your mileage may vary. I can't remember what issue I ran into exactly but I believe there were conflicts with the wire:model. Just an fyi.

HDVinnie: Seems fine to me. Tested it locally and has been in prod for over a week. If something else pops up a new issue can be opened.

PR #3787, opened 2024-04-28
(Release) v8.1.1

HDVinnie:
- quick patch release


Neodkbits:
is irc announce fixed so it doesn get the fwrite error ?

PR #3823, opened 2024-05-14
(Add) Validate max textarea length

Obi-Wana:
`Torrent.store` does not validate the length of `description` and `media info`even though they both have a DB limit of 65535 chars (type `text`).

This adds a max length of 65535 characters to the mentioned textareas. It prevents the error 500 when exceeding the DB character limit.

Other views using the same live wire `bbcode-input` (e.g. /users/USER/edit) are not affected and do not have character limit applied.

### DIFF
--- a/app/Http/Requests/StoreTorrentRequest.php
+++ b/app/Http/Requests/StoreTorrentRequest.php
@@ -115,7 +115,7 @@ class StoreTorrentRequest extends FormRequest
             'bdinfo' => [
                 'nullable',
                 'sometimes',
-                'max:4294967295',
+                'max:2097152',
             ],
             'category_id' => [
                 'required',
--- a/app/Http/Requests/StoreTorrentRequest.php
+++ b/app/Http/Requests/StoreTorrentRequest.php
@@ -105,17 +105,17 @@ class StoreTorrentRequest extends FormRequest
             ],
             'description' => [
                 'required',
-                'max:2097152'
+                'max:65535'
             ],
             'mediainfo' => [
                 'nullable',
                 'sometimes',
-                'max:2097152',
+                'max:65535',
             ],
             'bdinfo' => [
                 'nullable',
                 'sometimes',
-                'max:2097152',
+                'max:4294967295',
             ],
             'category_id' => [
                 'required',


Roardom (commented):
It's easy to paste in more than 65k characters and not realize that the end gets truncated when pasted into the browser so it's not the best idea. I also added backend validation here already: https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/3342

However, I notice now that UNIT3D's schema uses the `text` type for those columns, while when I submitted that PR I was using a database that happened to have used `longtext` for those fields instead. So, the validation isn't currently correct. However, I now wonder if those columns in UNIT3D should be adjusted to longtext (or even mediumtext). The database I was working on has around ~100 torrents with mediainfos longer than 65 k with the max being 1.2 m, and also about ~100 torrents with descriptions longer than 65k with the max being 900k (which is probably why I chose 2M as the limit in the previous validation), so it's quite possible that users may use slightly more data than 65 k and would not notice that their data is truncated when they paste it in.

Obi-Wana:
> It's easy to paste in more than 65k characters and not realize that the end gets truncated when pasted into the browser so it's not the best idea. I also added backend validation here already: #3342
> 
> However, I notice now that UNIT3D's schema uses the `text` type for those columns, while when I submitted that PR I was using a database that happened to have used `longtext` for those fields instead. So, the validation isn't currently correct. However, I now wonder if those columns in UNIT3D should be adjusted to longtext (or even mediumtext). The database I was working on has around ~100 torrents with mediainfos longer than 65 k with the max being 1.2 m, and also about ~100 torrents with descriptions longer than 65k with the max being 900k (which is probably why I chose 2M as the limit in the previous validation), so it's quite possible that users may use slightly more data than 65 k and would not notice that their data is truncated when they paste it in.

The truncating without notice sucks, for sure. My pain point is that there will be an error 500 once one exceeds the database limit of `text` instead of a warning/informational text. This issue will also persist nevertheless of the column type being `text`, `long`, or whatsoever.

Roardom:
Yes, the error 500 occurs because the current validation is validating a max length of 2M while the db errors if the length is greater than 65k. They should be in sync.

Obi-Wana:
> Yes, the error 500 occurs because the current validation is validating a max length of 2M while the db errors if the length is greater than 65k. They should be in sync.

So updating the existing validation would be an acceptable solution for now? If yes, I'll update the PR.

Roardom:
> > Yes, the error 500 occurs because the current validation is validating a max length of 2M while the db errors if the length is greater than 65k. They should be in sync.
> 
> So updating the existing validation would be an acceptable solution for now? If yes, I'll update the PR.

Indeed. Ideally all text form inputs should have a max limit on them in sync with their max db type, but many are missing and cause 500s. Would definitely appreciate it if in a future PR, you could go through the app/Http/Requests folder and add max length validation to everything, if you have the time. Since I doubt torrent submission is the only place affected.

app/Http/Requests/StoreTorrentRequest.php
```diff
@@ -105,17 +105,17 @@ function (string $attribute, mixed $value, Closure $fail): void {
...
             'mediainfo' => [
                 'nullable',
                 'sometimes',
-                'max:2097152',
+                'max:65535',
             ],
             'bdinfo' => [
                 'nullable',
                 'sometimes',
-                'max:2097152',
```
Roardom, line 118: Let's keep the bdinfo one as is so users don't try and paste in 4GB of text. I should have specified earlier but the validation should be less than or equal to the database column, but the limit should still be reasonable.

Roardom (changes requested):

app/Http/Requests/StoreTorrentRequest.php
Obi-Wana: Reverted in 548a7e4d0bfbdf21150557d703cff2606c09c995

Roardom (approved):

PR #3838, opened 2024-05-21
(Update) Improve key-value styles

Roardom:
If a short key and a long value are followed by a long key and short value, this PR allows them to both expand within their own row and overlap the previous column gap. It also allows keys with multiple values.


Roardom:
Before:

image


After:

image

PR #3854, opened 2024-05-22
(Update) comparison inside code block

LostRager:
Allows [comparison] to be placed inside [code] without being rendered.


Roardom (changes requested):
UNIT3D's bbcode code block is not the same as other platforms and does not prevent rendering of inner bbcode.

There are many tens of thousands of uploaded torrents with descriptions, forum posts, comments, etc. that already rely on the rendering of bbcode inside code blocks. It's basically impossible to compatibly migrate such content to a new system where code blocks don't render its contents, which would be necessary if such a change such as this were to occur. 

Also, comparison tags shouldn't be special. If the intention of code blocks were to be changed to allow the raw content within without rendering, then it should be all content that doesn't render and not just comparison tags.

PR #3859, opened 2024-05-24
(Fix) Broken bookmarks button

Roardom:
This must have been broken for a long time. This also fixes the issue when ticking checkboxes too quickly the state being reset. It required both making sure `small-bookmark-button.blade.php` was surrounded by an element completely, as well as adding wire:key to the inner row loop. Adding wire:key to the component itself doesn't work because the attributes aren't passed through...

### DIFF
--- a/resources/views/components/torrent/row.blade.php
+++ b/resources/views/components/torrent/row.blade.php
@@ -19,6 +19,7 @@
     data-category-id="{{ $torrent->category_id }}"
     data-type-id="{{ $torrent->type_id }}"
     data-resolution-id="{{ $torrent->resolution_id }}"
+    wire:key="torrent-search-row-{{ $torrent->id }}"
 >
     @if (auth()->user()->show_poster == 1)
         <td class="torrent-search--list__poster">
@@ -141,7 +142,7 @@
                 </a>
             @endif
 
-            {{-- @livewire('small-bookmark-button', ['torrent' => $torrent, 'isBookmarked' => $torrent->bookmarks_exists, 'user' => auth()->user()], key('torrent-'.$torrent->id)) --}}
+            @livewire('small-bookmark-button', ['torrent' => $torrent, 'isBookmarked' => $torrent->bookmarks_exists, 'user' => auth()->user()], key('bookmark-torrent-'.$torrent->id))
 
             @if (config('torrent.download_check_page'))
                 <a
--- a/resources/views/livewire/small-bookmark-button.blade.php
+++ b/resources/views/livewire/small-bookmark-button.blade.php
@@ -1,17 +1,17 @@
-@if ($this->isBookmarked)
-    <button
+<button
+    @if ($this->isBookmarked)
         wire:click="destroy({{ $torrent->id }})"
         class="form__standard-icon-button"
         title="Unbookmark"
-    >
-        <i class="{{ config('other.font-awesome') }} fa-bookmark-slash"></i>
-    </button>
-@else
-    <button
+    @else
         wire:click="store({{ $torrent->id }})"
         class="form__standard-icon-button"
         title="Bookmark"
-    >
+    @endif
+>
+    @if ($this->isBookmarked)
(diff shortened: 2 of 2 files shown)


Roardom:
Nevermind, the bug is triggered by selecting a couple of filters, and then bookmarks look like this...

image

This doesn't look like an easy fix...

Roardom:
Figured it out.

LostRager:
Does this mess with selecting torrents to be deleted in the similar view?

Roardom:
> Does this mess with selecting torrents to be deleted in the similar view?

I selected a few torrents in similar view rapidly and got no errors in the console. Is there something specific you want me to check for?

LostRager:
> > Does this mess with selecting torrents to be deleted in the similar view?
> 
> I selected a few torrents in similar view rapidly and got no errors in the console. Is there something specific you want me to check for?

It broke for me, but i have an altered similar page so its probably just my fault then. 
When i selected a torrent, all torrents disappeared and got the error: `Uncaught Snapshot missing on Livewire component with id: xxxx`

No worries then. all good :)

Roardom:
That's exactly the error the bookmark button was causing. Go through every loop in your blade and add a `wire:key="unique-string"` to the wrapping element inside the loop, as well as adding `:key="unique-string"` to any livewire component. Also make sure any livewire components are wrapped in a single element. I can't trigger that error on this branch however. You can also view the html in the dev console and it will be completely messed up caused from livewire doing bad diffing. Hopefully that can help you narrow it down.

LostRager:
Seems like there are still issues with this. 
image
from blu.

HDVinnie:
> Seems like there are still issues with this. image from blu.
> 
> I have the same issue on my install.
> 
> Happends when searching for torrents.

BLU small bookmark seems fine for me. It was just updated to latest v8 branch and there hasn't been a new release so fresh install wouldn't have it.

LostRager:
> > Seems like there are still issues with this. image from blu.
> > I have the same issue on my install.
> > Happends when searching for torrents.
> 
> BLU small bookmark seems fine for me. It was just updated to latest v8 branch and there hasn't been a new release so fresh install wouldn't have it.

Sent a dm on discord with a video.

HDVinnie:
@Roardom @LostRager is correct. Still broken. Seems to happen when searching by name

Roardom:
:disappointed: 

I'll try and figure this out

Roardom:
@LostRager After a ridiculously long time spent debugging, I've discovered that removing `x-cloak` from this line removes the `component not found` error from the console (at least in my local setup).

The only way I was able to trigger that error was by typing ~5-10 characters in the name field, at a frequency slower than the 250ms debounce but fast enough that the previous search hadn't returned results yet.

Do you mind trying that out and see if that works for you? I found triggering the `snapshot` error much more difficult so I was focusing on fixing the `component not found` error hoping that fixing one would fix the other.

LostRager:
> @LostRager After a ridiculously long time spent debugging, I've discovered that removing `x-cloak` from this line (
> 
> https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/fc18ca319798228e3dc469842bdd6414f66ea01c/resources/views/livewire/torrent-search.blade.php#L30
> 
> ) removes the `component not found` error from the console (at least in my local setup).
> The only way I was able to trigger that error was by typing ~5-10 characters in the name field, at a frequency slower than the 250ms debounce but fast enough that the previous search hadn't returned results yet.
> 
> Do you mind trying that out and see if that works for you? I found triggering the `snapshot` error much more difficult so I was focusing on fixing the `component not found` error hoping that fixing one would fix the other.

Seems like the issue is fixed with that change.

Roardom:
Glad to hear!

Also should note: it appears that it will still break if you have view cache enabled. We need to both remove x-cloak as well as clear view cache to remove these console errors.

Roardom:
Found this: https://github.com/livewire/livewire/discussions/5658

Which seems to align with what I'm seeing.

PR #3884, opened 2024-05-28
(fix) allow port 1080 for SOCKS5 connections

bersLucas:
Ports in the blacklist or ports below 1024 are blocked, but if a user is using a SOCKS5, the connection will go through port 1080 and they will see the "Illegal port" error message.

This PR allows 1080 specifically

### DIFF
--- a/app/Http/Controllers/AnnounceController.php
+++ b/app/Http/Controllers/AnnounceController.php
@@ -65,6 +65,12 @@ final class AnnounceController extends Controller
         6699,
     ];
 
+    // Port Whitelist
+    private const WHITE_PORTS = [
+        // Port used by SOCKS5 connections
+        1080
+    ];
+
     private const HEADERS = [
         'Content-Type'  => 'text/plain; charset=utf-8',
         'Cache-Control' => 'private, no-cache, no-store, must-revalidate, max-age=0',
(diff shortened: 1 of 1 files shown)


Roardom (changes requested):
`$queries['port'] !== 1080` is always false because `$queries['port']` is a string.

Also, `$queries['port'] < 1024 && $queries['port'] != 1080` and `$queries['port'] < 1024` are logically equivalent.

bersLucas:
> `$queries['port'] !== 1080` is always true because `$queries['port']` is a string.
> 
> Also, `$queries['port'] < 1024 && $queries['port'] != 1080` and `$queries['port'] < 1024` are logically equivalent.

But 1080 is greater than 1024, should a whitelist be used instead?

bersLucas:
Added changed. My apologies as I am a little new to PHP

HDVinnie:
> > `$queries['port'] !== 1080` is always true because `$queries['port']` is a string.
> 
> > 
> 
> > Also, `$queries['port'] < 1024 && $queries['port'] != 1080` and `$queries['port'] < 1024` are logically equivalent.
> 
> 
> 
> But 1080 is greater than 1024, should a whitelist be used instead?

The logic is a less than sign not greater than.

bersLucas:
Screenshot 2024-05-29 154738

That's interesting. My message is "illegal port 1". Could this mean anything?

bersLucas:
Related: https://github.com/qbittorrent/qBittorrent/issues/18472

HDVinnie:
Yes it says right in message your client is using port 1. This is a client bug. Not UNIT3D.

bersLucas:
Since Libtorrent is intentionally sending 1, can this ever be supported in Unit3d or is this a lost cause?

HDVinnie:
I'll discuss it over with the team and let you know. As for now this pull request is closed as the logic didn't make sense and it was against master branch instead of v8.x.x as described in contributing guide.

Roardom:
> [...]
> 
> That's interesting. My message is "illegal port 1". Could this mean anything?

If you want to allow libtorrent-based socks5, you have to do `!== '1'`, not `!== '1080'`.

UNIT3D is designed to incentivize a strong seedbase, such as:

- Offering bonus points for seeders
- Offering permanent freeleech if you have seeded 10TiB for an average of 60 days
- Displays to other users the status of your connectivity in the peerlist

It is not the goal of UNIT3D to encourage misuse of p2p protocols. BitTorrent is designed for peer to peer usage. This means that all users should be connectable. Users refusing to be connectable is against the very intention of peer to peer communication. There are many alternative non-peer-to-peer methods of data transfer that you could use instead if you prefer to not be connectable: web server, ftp server, etc.

wgomg:
I don't mean to necrobump but has there been further discussion on this?

Roardom:
> I don't mean to necrobump but has there been further discussion on this?

The stance has remained firm. Since this PR, the tracker now has the option to not provide peer lists to unconnectable users, as well as not provide bonus points for unconnectable peers.

PR #3908, opened 2024-06-05
(Update) Use unsigned 32-bit integer for migration create

Roardom:
For 99% of cases it's not necessary to support more than 4 billion rows.


Roardom:
Nevermind... I already made this change . No clue what I was thinking: https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/3587

PR #3910, opened 2024-06-05
(Fix) Language stats

HDVinnie:
- locale is now stored in user_settings table.


Roardom (changes requested):
if `$name` is `config('app.locale')`, add in the number of users that don't have a `UserSetting`

PR #3915, opened 2024-06-08
Add user seeding status to api

Audionut:
Does this work?


Roardom:
No it doesn't work at all. You'll also take down the site if you add any sort of loading a user's seed count on an api route that users negligently spam every 10 seconds to race

PR #3950, opened 2024-06-26
(Fix) IRC Bot PRIVMSG prefix missing

costaht:
According to this article https://matt.ucc.asn.au/mirror/irctelnet.html and some examples from the RFC 1459 (https://www.rfc-editor.org/rfc/rfc1459#section-4.4.1), when using PRIVMSG to send a long text, the message has to be prefixed by a colon otherwise the message will be cropped.
Below is a screenshot of the before and after adding the prefix to the messages:

image


Roardom:
This should probably have been put inside the `say()` method itself instead of where it was called: https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/68ef67e0fd44e430c1130ef7520c4a7479c0ba53/app/Bots/IRCAnnounceBot.php#L140. There's still a few other places where the IRCAnnounceBot is used

Roardom:
Do you happen to know more about this specification requirement? I was searching through a few irc specifications and haven't yet found reference to the colon being officially required. I would like to know more about it.

Roardom:
Aha, it's in the parameter section. Perhaps it's better to add it to the privmsg() method directly instead:

https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/68ef67e0fd44e430c1130ef7520c4a7479c0ba53/app/Bots/IRCAnnounceBot.php#L222

I would appreciate it if you can open another PR for that.

Issue #3646, opened 2024-03-11
[Request] Collapsible list of voters in request


utpto:
Some requests has many voters, sometimes more than 50.
It would by great to have collapsible list of voters in request, collapsed by default.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/HDInnovations) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/HDInnovations/UNIT3D-Community-Edition/issues/3646">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/3646/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/3646/pledge.svg">
</picture>
</a>

PR #3954, opened 2024-07-01
(Add) Request #3646

HDVinnie:


Roardom (changes requested):
Instead of grouping, maybe just only add the x-show attribute to the row if $loop is greater than 10. I think that would solve the problem better. It's useful to know when a user has requested multiple times and if they're spamming it daily or a few times a year

PR #3959, opened 2024-07-02
(Fix) `bounty->request` relationship

Roardom:
The foreign key isn't the standard name, so we have to customize it, otherwise our relations return `null`.


HDVinnie:
Shouldn't we just rename the column in DB table.

Roardom:
> Shouldn't we just rename the column in DB table.

Yes. But at that point might as well fix all the wrong names and poor relations in the torrent requests / bounty / claims structure.

PR #3964, opened 2024-07-04
(Update) Use logical `OR` when searching tmdb/imdb/tvdb/malid

Roardom:


Roardom:
Note: Newznab requires the id search to be `OR` instead of `AND`[^1], but some non-*znab scripts currently rely on it being `AND`. 

Possible solutions:

- Add another parameter to specify if the logical condition should or `OR` or `AND`, defaulting to `AND` if neither was specified.
- Add a dedicated torznab-compatible API.

I plan on going with the second option, and incorporating #3061 into it.

[^1]: https://torznab.github.io/spec-1.3-draft/revisions/1.1-Newznab-supportedParams.html

PR #3824, opened 2024-05-16
(Add) Active and inactive uploader staff view

Obi-Wana:
Adds a view for staff to see uploaders and their stats, such as total uploads and uploads within the last 60 days.
This will reveal uploaders that might need demotion.

The action column is currently empty. Not sure if any action should be added. Maybe a "demote" or "edit" button?

PR #3972, opened 2024-07-08
(Update) Uploader Stats

HDVinnie:
- recent PR #3824
- route name and prefix should be plural
- a site may have many is_uploader groups so lets use whereIntegerInRaw and orderBY group_id

### DIFF
--- a/app/Http/Controllers/Staff/UploaderController.php
+++ b/app/Http/Controllers/Staff/UploaderController.php
@@ -27,12 +27,10 @@ class UploaderController extends Controller
      */
     public function index(): \Illuminate\Contracts\View\Factory|\Illuminate\View\View
     {
-        $uploaderGroups = cache()->rememberForever('uploader_groups', fn () => Group::where('is_uploader', '=', true)->pluck('id'));
-
         return view('Staff.uploader.index', [
             'uploaders' => User::with(['group'])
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Staff/UploaderController.php
```diff
@@ -27,12 +27,12 @@ class UploaderController extends Controller
...
     public function index(): \Illuminate\Contracts\View\Factory|\Illuminate\View\View
     {
-        $uploaderGroup = cache()->rememberForever('uploader_group', fn () => Group::where('is_uploader', '=', true)->pluck('id'));
+        $uploaderGroups = cache()->rememberForever('uploader_groups', fn () => Group::where('is_uploader', '=', true)->pluck('id'));
 
         return view('Staff.uploader.index', [
             'uploaders' => User::with(['group'])
                 ->withCount('torrents as total_uploads')
-                ->where('group_id', '=', $uploaderGroup)
+                ->whereIntegerInRaw('group_id', [$uploaderGroups])
```
Roardom, line 35: Just use a subquery:

```php
->whereIn('group_id', Group::select('id')->where('is_uploader', '=', true))
```

Roardom (changes requested):

PR #3973, opened 2024-07-08
(Update) Prefer foreach array iterator over  variable

Roardom:
Only do the easy cases for now. Goal is to eventually use a custom blade directive for  to avoid Laravel's default  as it approximately halves blade execution time for pages of 25 items per loop.


HDVinnie (changes requested):
Conflict

PR #3975, opened 2024-07-09
(Add) Phpstan bladestan static analysis

Roardom:
Allows static analysis of blade templates.

I had to add semi-colons to the end of `@php` directives, and swap `@json` for `{{ @JS::from }}` otherwise bladestan would crash.


AJenbo:
> I had to add semi-colons to the end of `@php` directives, and swap `@json` for `{{ @JS::from }}` otherwise bladestan would crash.

These issues have been addressed in the latest version

PR #3983, opened 2024-07-16
(Update) Use composite primary key for peers

Roardom:
The same as was done in #2446 and reverted in ace8dcb but this time using workarounds for Eloquent when it tries to use the non-existing `id` column.


Roardom:
This change increased insertion rate performance from 15.1k records/s  to 26.4k records/s.

PR #4001, opened 2024-07-22
(Fix) Small bookmark button again again

Roardom:
Goodbye livewire. Livewire's diffing algorithm kept messing up this livewire component.


LostRager:
Seems like there are issues with this, if you bookmark something and then switch to page 2 it sends a delete bookmark request for every torrent on the page. It happends everytime you switch pages aswell, causing you to get ratelimited on api calls.

PR #4007, opened 2024-07-23
(Update) Use composite primary key for history

Roardom:


Roardom:
This change increased insertion rate performance from 13.1k records/s to 19.1k records/s.

PR #4017, opened 2024-08-03
(Add) Responsive styles for comments on mobile screens

EkoNesLeg:
Adds media queries to adjust the margins and padding for the comment replies so that they're properly displayed on screens with a max-width of 576px.


HDVinnie:
Forums and comments will need mobile overhaul but this works for now. Thanks.

PR #4044, opened 2024-08-12
(Fix) Various invite tree fixes

Roardom:
Average seedtime is now average and not total. Swap tree stats of seedsize vs seedtime. Use correct user for last action. Filter out infinite ratios.

### DIFF
--- a/app/Http/Controllers/User/InviteTreeController.php
+++ b/app/Http/Controllers/User/InviteTreeController.php
@@ -94,6 +94,7 @@ class InviteTreeController extends Controller
                 'receiver' => fn ($query) => $query
                     ->withTrashed()
                     ->with('group')
+                    ->withAvg('history', 'seedtime')
                     ->withSum('history', 'seedtime')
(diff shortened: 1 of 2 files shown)


Audionut:
Ancestors should not be viewable to user accounts IMO.

Roardom:
> Ancestors should not be viewable to user accounts IMO.

I consulted multiple staffers and they all agreed in having transparency.

Audionut:
Transparency for what, exactly?

Roardom:
> Transparency for what, exactly?

To see the full dependency chain of who's trusting you to be a good user.

Also a reminder that other tracker software such as the one you used also showed everyone's inviter on their profile page for anyone to view.

Audionut:
They don't show up in the tree. Why on earth does anyone need to see who is depending on them being a good user.

Audionut:
MiM also suggests that the amount of levels down that can be seen should be tied to a perk.

PR #4078, opened 2024-08-23
(Fix) New users not receiving default stylesheet defined in config

Roardom:


HDVinnie:
@Roardom I think when user_settings was refactored to its own table we missed some. Instead of these null coalescing operators should we just set user settings in action. https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/5125cd33d6c05027f5c2ac16cbeba192dc44ad39/app/Actions/Fortify/CreateNewUser.php#L86-L87

Roardom:
I was trying to implement it in a way where a sysop could change the default style to e.g. Halloween during October and can reset it back afterwards, unless the user chose their own stylesheet which would then override the default. Creating the settings object on user creation wouldn't allow that.

However, I should probably add the same thing to locale and remove those attributes from the user create method. There's a rule in larastan to detect when properties are passed to the create method that aren't actually properties. Might be worth adding it.

PR #4079, opened 2024-08-24
(Fix) Warning actions on wrong warning

Obi-Wana:
Fixes the issue where deactivating/deleting/restoring/reactivating a warning performed the action on a random warning instead of the once selected.


Roardom (changes requested):
Already reviewed this elsewhere, but seeing I have this notification still, basically you should be able to remove the `data-warning-id` from every button, and use `this.$root` instead of `this.$el`

Roardom (approved):

PR #4099, opened 2024-08-27
(Update) imgbb.js

frenchcutgreenbean:
Change imgbb upload helper from thumbnails to fullsize images, set width to 350, and single space separate.


HDVinnie:
@frenchcutgreenbean thank you

PR #4103, opened 2024-08-27
(Fix) Prevent Password Manager and Search History Interference

EkoNesLeg:
Added `type="search"` to properly classify the input field and `autocomplete="off"` to prevent browsers from saving search history and reduce interference from password managers.


Roardom (changes requested):
Need to do a ctrl+f across the resources/view/livewire directory for `<input` and add these to all the searches, not just this one.

Roardom (approved):

PR #4113, opened 2024-08-29
(Add) Torrent folder to API

Audionut:
Only when present.


Roardom (changes requested):
Discussed this internally and we'd rather the folder always be included in the api response, even if it is null or empty.

Roardom (changes requested):
Thanks. Can you also squash both commits and rename them to match the contributor guidelines.

Roardom (approved):

PR #4131, opened 2024-09-02
(Update) Don't query database in filter api

Roardom:
Better database performance using raw pagination since it doesn't have to run the data through eloquent resources (bad when there's thousands of TorrentFiles) or query the database to count the count of torrents inside the database for the total item count.


Roardom:
This PR dropped latency from an average of 95ms (95th 339ms) to an average of 14ms (95th 26ms)

PR #4144, opened 2024-09-05
(Fix) Stats show total seedtime, not avg

Obi-Wana:
The stats calculate the total seed time, not the average. Therefor, naming should be for total time.


Roardom:
Why not both? Send a PR for another tab :)

PR #4156, opened 2024-09-12
(Add) Allow staff to set individual history entries to immune

Obi-Wana:
This is meant for some edge cases where a staff would want to protect a user from receiving a HnR warning on a specific history entry.


Roardom (changes requested):
Can you instead have an `update()` method, that accepts an UpdateHistoryRequest Form Request with a sole validated boolean `immune` field? Then on the front-end, the wording and value can be changed depending on the state it's in.

Also, after doing so, can you use `$request->user()` instead of `auth()->user()`.

Additionally, I just noticed you added this to the user-torrents page, which is already very cramped. Can you maybe make the icon clickable instead? Or another way for the page to still be manageable at 1080p.

Roardom (approved):
Thank you!

PR #4164, opened 2024-09-15
 (Update) Don't query database in filter api take 2

Roardom:
Take 2 of #4131

### DIFF
--- a/app/Http/Controllers/API/TorrentController.php
+++ b/app/Http/Controllers/API/TorrentController.php
@@ -645,8 +645,6 @@ class TorrentController extends BaseController
                             'type_id'          => $hit['type']['id'] ?? null,
                             'resolution_id'    => $hit['resolution']['id'] ?? null,
                             'created_at'       => date('Y-m-d\TH:i:s.Z\Z', $hit['created_at']),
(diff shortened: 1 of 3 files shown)


HDVinnie (approved):
I have reviewed this twice now. Still worried lol. Can you review it again yourself as well @Roardom

Roardom:
The only sensitive info meilisearch stores is username for anon torrents, which is properly dealt with by return `Anonymous` for the uploader username if the `anon` flag is checked.

The only data being added to the return collection that isn't from meilisearch is the tmdb image url, the created_at date, and the route to `torrents.show`, which contains a numeric `id` parameter.

- The tmdb image url function does not output any sensitive data.
- The created_at date does not output any sensitive data
- The `torrents.show` route does not output any sensitive data - the id is not considered senstive

The above only concerns the cached data. After the cached data is fetched from cache, it alters the data before sending as response (but does not alter the data in the cache). It adds sensitive info to the response, such as the rsskey, which itself provides access to the passkey via the magnet link or by downloading the torrent. This sensitive data is not cached. The sensitive data is sourced from the auth helper, and not through an existing (possibly incorrect) user object.

We were originally concerned about apikeys leaking through the pagination urls, but that doesn't seem to be the case since other users' apikeys are never cached.

PR #2251, opened 2022-05-31
(Refactor) Use Laravel Carbon

HDVinnie:
This pull request includes changes from your build using the "Shift Workbench".

**Before merging**, you need to:

- Checkout the `shift-build-3080` branch
- Review **all** comments for additional changes
- Thoroughly test your application ([no tests?](https://laravelshift.com/laravel-test-generator), [no CI?](https://laravelshift.com/ci-generator))

Don't hesitate to send your feedback to shift@laravelshift.com or share your :heart: for Shift on [Twitter](https://twitter.com/laravelshift).

PR #4165, opened 2024-09-20
(Add) support for annual donation goals and one-off drives

EkoNesLeg:


config/donation.php
```diff
@@ -22,8 +22,11 @@
     | Configure site to use Donation System
     |
     */
-    'is_enabled'   => true,
+    'is_enabled'   => false,
     'monthly_goal' => 100,
+    'yearly_goal'  => 1200,
+    'goal_type'    => 'yearly', // monthly or yearly
```
Obi-Wana, line 28: Instead of having both values defined and choosing between them, how about adding a general goal variable that is either used for monthly or yearly goal (defined by `goal_type`)?

EkoNesLeg: Great idea

resources/views/partials/top_nav.blade.php
```diff
@@ -1,3 +1,8 @@
+@php
+    use App\Models\Donation;
+    use Carbon\Carbon;
+@endphp
+
```
Roardom, line 5: Use `@use()` instead of `@php`.

Also use `Illuminate\Support\Carbon' instead of `Carbon\Carbon`: https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/2251

resources/views/partials/top_nav.blade.php
```diff
@@ -221,15 +224,21 @@
...
+                    $startDate = config('donation.start_date');
+
+                    $sum = Donation::join('donation_packages', 'donations.package_id', '=', 'donation_packages.id')
+                        ->where('donations.status', Donation::APPROVED)
+                        ->when(
+                            config('donation.goal_type') === 'yearly',
+                            fn ($query) => $query->whereBetween('donations.created_at', [Carbon::parse($startDate), Carbon::parse($startDate)->addYear()]),
+                            fn ($query) => $query->whereBetween('donations.created_at', [now()->startOfMonth(), now()->endOfMonth()])
+                        )
```
Roardom, line 237: Monthly should also use a start date. Some math using carbon functions can be done here to use the day of the month provided in the config, and find the dates with the same day of the month before and after `now()`. It would be useful to break the arrays in the whereBetween across multiple lines for better readability

resources/views/partials/top_nav.blade.php
```diff
@@ -221,15 +224,21 @@
...
-                        });
-                    $percentage = $sum ? min(100, number_format(($sum / config('donation.monthly_goal')) * 100)) : 0;
+                    $goal = config('donation.goal');
+
+                    $startDate = config('donation.start_date');
+
+                    $sum = Donation::join('donation_packages', 'donations.package_id', '=', 'donation_packages.id')
+                        ->where('donations.status', Donation::APPROVED)
+                        ->when(
+                            config('donation.goal_type') === 'yearly',
```
Roardom, line 234: Yearly can probably be renamed to "one-time" or "drive", as it's not actually yearly.

Roardom (changes requested):

HDVinnie:
Stale PR

PR #4169, opened 2024-09-22
(Add) Client torrent size sum to profile page

Roardom:
This query's execution time is increased from 10 ms to ~550 ms for power users with a few thousand torrents, or ~30 ms for average users with a few dozen torrents. Is the increased page load time worth its usefulness?


HDVinnie (commented):
Not worth it imo

Roardom:
Can we test this temporarily in an environment with a tuned database and see how fast the queries load? I've got a feeling there's more to the story than my numbers tell. Considering the usefulness of this PR, I'd like to double check.

HDVinnie:
> Can we test this temporarily in an environment with a tuned database and see how fast the queries load? I've got a feeling there's more to the story than my numbers tell. Considering the usefulness of this PR, I'd like to double check.

Will test in a tuned production environment this weekend and report back.

PR #4170, opened 2024-09-22
(Update) Allow reopening tickets

Roardom:


HDVinnie (approved):
In future we should probably move close and reopen to separate controller with update and destroy or alike for better CRUD.

Roardom:
> In future we should probably move close and reopen to separate controller with update and destroy or alike for better CRUD.

Agreed. We can have an `OpenTicketController` with store and destroy methods.

PR #4178, opened 2024-09-25
(Update) api folder searching

Audionut:


Audionut:
> Chatgpt?

It probably would have caught those things I overlooked. Not a bad idea actually, I should dump the entire file into i and see what it spits out. K said you liked reviewing chatgtp code.

Roardom:
> K said you liked reviewing chatgtp code.

The entire PR consists of

- setting a non-existent property to a variable that isn't defined anywhere,
- setting another non-existent property,
- adding a validation rule that would've prevented it from working,

How could I expect that this wasn't chatgpt?

HDVinnie:
Stale PR

PR #4209, opened 2024-10-08
(Refactor) Quicksearch

HDVinnie:
- refactored from livewire to alpinejs
- this is a crudish draft but is fullly functional
- script should be moved to own alpine component in js dir
- person command and filter is simply a test and works
- fixes long time safari issue when click result just closes search and doesnt direct to URL by following:

@click.prevent Directive: Added @click.prevent="navigateTo(result.url)" to the anchor tag to handle the click event and navigate to the URL.

navigateTo Method: Added a navigateTo method in the quickSearch function to change the window location to the clicked URL.

Event Listener: Updated the init method to add a click event listener that stops propagation and navigates to the URL.


app/Console/Commands/IndexPeopleToMeilisearchCommand.php
```diff
@@ -0,0 +1,35 @@
...
+    protected $signature = 'index:people-to-meilisearch';
+
+    protected $description = 'Command description';
+
+    public function handle(): void
+    {
+        $client = new Client(config('scout.meilisearch.host'), config('scout.meilisearch.key'));
+        $index = $client->index('people');
+
+        $people = Person::all(['id', 'name', 'birthday', 'still']);
```
Roardom, line 22: It will be much quicker if you use json syntax in mysql, that's the main bottleneck when importing data.

Since you're not using scout, try https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/3061

This should also be ran daily.

app/Console/Commands/IndexPeopleToMeilisearchCommand.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Console\Commands;
```
Roardom, line 5: Missing docblock.

Roardom (changes requested):

app/Http/Controllers/API/QuickSearchController.php
```diff
@@ -0,0 +1,108 @@
...
+                'year'  => $result['birthday'],
+                'image' => $result['still'] ? tmdb_image('poster_small', $result['still']) : 'https://via.placeholder.com/90x135',
+                'url'   => route('mediahub.persons.show', ['id' => $result['id']]),
+                'type'  => 'Person',
+                'score' => $result['_rankingInfo']['score'] ?? 0,
+            ];
+        }
+
+        // Sort results by score
+        usort($results, fn ($a, $b) => $b['score'] <=> $a['score']);
```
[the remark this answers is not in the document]
HDVinnie: This means we cannot use scout and instead Meilisearch client directly which im more than happy to do.

app/Console/Commands/IndexPeopleToMeilisearchCommand.php
Roardom: Also, you could try experimenting with adding a covering index for these columns. That way when you do a full table scan, it can do a full scan over just the index instead of the table.

HDVinnie: To be done in separate PR once get baseline in prod on current.

PR #4218, opened 2024-10-10
Change mediainfo to longtext

Obi-Wana:
Some special cases (e.g., many subtitles) cause the mediainfo to become larger than `text` can store. The next bigger type should fix this.


HDVinnie (approved):

HDVinnie:
Thanks.

PR #4221, opened 2024-10-12
(Fix) Don't compress webp losslessly through image proxy

Roardom:
The `&ll` query parameter prevents webp from being compressed: https://wsrv.nl/docs/format.html#lossless-compression. Adding `&ll` to non-webp images still seems to work fine from testing.


Roardom:
Must have been asleep when I wrote this patch. Should have been `don't compress webp lossily through image proxy`.

PR #4265, opened 2024-10-21
(Update) Playlist searching

HDVinnie:
- use wildcards


Roardom (changes requested):
Replacing space with a wildcard as well would be useful.

Roardom (changes requested):
Could you squash both commits to a single commit and name it something related to the code that was changed?

E.g. `update: use wildcard search when searching playlists` or "update: use mysql \`LIKE\` instead of \`=\` when searching playlists" or `update: swap exact search for fuzzy search when searching playlists`

PR #4270, opened 2024-10-22
(Fix) Validation for torrent request updates

thepolishplumber:
Hi,

Updated the validation rules in UpdateTorrentRequestRequest so they match the ones in StoreTorrentRequestRequest. The change came about because updating requests with no_meta category would break on validation. Weren't sure if any test needed updating.


HDVinnie (changes requested):
You need to update the return for new conditional rule so CI passes for larastan

thepolishplumber:
Done.

PR #4273, opened 2024-10-23
(Fix) Invalid value type for page in Torrent search

HDVinnie:
- This ensures that the page parameter is correctly cast to an integer before being passed to the Meilisearch API.


Roardom (commented):
This might need to be casted to `1` and not `0`. Have you tested it yet? (I haven't)

PR #4291, opened 2024-10-30
(Add) Internal notes for staff invitations

### LINKED ISSUE #4045: [Request] Apply note to invite
Would be highly beneficial for staff members to be able to send an invite, and at the same time apply a note to the future account that is created via that invite. Case in point, Invite recruitment. 

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/HDInnovations) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/HDInnovations/UNIT3D-Community-Edition/issues/4045">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/4045/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

IonBazan:
Fixes #4045

Adds an `internal_note` for invitations which will be used to create an internal note when user registers.


app/Actions/Fortify/CreateNewUser.php
```diff
@@ -93,10 +93,18 @@ public function create(array $input): RedirectResponse | User
         $user->emailUpdates()->create();
 
         if (config('other.invite-only') === true) {
-            Invite::where('code', '=', $input['code'])->update([
+            $invite = Invite::where('code', '=', $input['code'])->withoutTrashed()->first();
```
Roardom, line 96: Shouldn't need to do `->withoutTrashed()` here. It's only included above because soft deletion isn't considered for `Rule::exists()` validation.

resources/views/user/invite/create.blade.php
```diff
@@ -69,6 +69,18 @@ class="form__textarea"
                             {{ __('common.message') }}
                         </label>
                     </p>
+                    @if ($user->group->is_modo)
+                    <p class="form__group">
+                        <textarea
+                            id="internal_note"
+                            class="form__textarea"
+                            name="internal_note"
```
Roardom: You might need to add `placeholder=" "` here for the autoexpand to work in browsers that don't yet support `field-sizing: content` as well as to get the label to float.

Roardom (changes requested):
Thanks!

app/Actions/Fortify/CreateNewUser.php
IonBazan: Thanks! I wasn't sure why is it used above so made it explicit just in case. Fixed in my next commit.

IonBazan:
@Roardom thanks for the review! I chose a simple task for my first contribution in this repository but I'll try to help with more improvements 🚀 

I wasn't sure if `CreateNewUser` is tested anywhere - it would be good to verify it's working as expected after my changes but I'm quite new to Fortify so I might need some guidance there.

Roardom:
> @Roardom thanks for the review! I chose a simple task for my first contribution in this repository but I'll try to help with more improvements 🚀
> 
> I wasn't sure if `CreateNewUser` is tested anywhere - it would be good to verify it's working as expected after my changes but I'm quite new to Fortify so I might need some guidance there.

Thanks for the PR! If I'm going to be honest, UNIT3D is lacking a lot of proper testing. There shouldn't be too many issues with the code though as it follows the same logic as the other items also created upon user creation. All of the attributes are mass assignable and there are no database fields without a default that are omitted. The field is properly validated to only be filled by staff and only shown to staff :+1: . But, I've been known to be wrong before.

Regarding Fortify, we've personally had a terrible experience with it and I definitely wouldn't recommend using it if you had a choice. Main reason is all the workarounds we've had to add to it to increase its security (lack of rate limits, timing attacks with emails, user enumeration) and handle our logic with account invitations as well as disabled and banned accounts. We'll probably migrate away from it at some point.

But basically, when a user registers using the form which posts to fortify's `register` route, it will call Fortify's RegisteredUserController::store which has the CreateNewUser dependency injected and calls the `create()` method.

With regards to testing it, we never had our tests migrated over when switching from laravel/ui. If you wish, you could add a new test covering it, but it'd be beyond the scope of this PR.

Roardom (approved):
Oh, I see you've added the Fortify tests since I last checked the PR last night. Thank you!

HDVinnie:
Thank you very much for the contribution. 🖖

PR #4292, opened 2024-10-30
(Refactor) Normalize forum aggregates

Roardom:
Even though the queries were ultra efficient before (always less than 5ms), the total combined cpu execution time for forums is not greater than 0.3% of all total execution time. Opt for cleaner code and a normalized database instead.

In a forum with 40k posts and 5k topics:

Topic index: 26 ms -> 83ms.
The forum category topic index: 29 ms -> 63 ms.
The forum index: 20 ms -> 55 ms.
The forum topic index: 19 ms -> 25 ms.
User topics: 18 ms -> 21 ms.
Subscriptions: 19 ms -> 20 ms.

These numbers are very reasonable and normalization should be fine here.

Unfortunately, Laravel doesn't have a function for hasManyThrough()->latestOfMany() so the eager load has to be done manually.

TODO: topic reads


Roardom:
The current system is performant, and isn't buggy. Although I'd prefer cleaner code, the existing code works fine as is, if not better than what's here.

PR #4293, opened 2024-10-31
(Fix) Claiming prize for last day of event

Roardom:


app/Http/Controllers/ClaimedPrizeController.php
```diff
@@ -32,7 +32,7 @@ public function store(Request $request, Event $event): \Illuminate\Http\Redirect
             return back()->withErrors('Event is not active.');
         }
 
-        if (!now()->isBetween($event->starts_at, $event->ends_at)) {
+        if (!now()->isBetween($event->starts_at, $event->ends_at->addDay())) {
```
IonBazan, line 35:
```suggestion
        if (!now()->isBetween($event->starts_at->startOfDay(), $event->ends_at->endOfDay())) {
```

Or even better, move this condition to a helper method in `Event` class:
```php
    public function isAvailable(): bool
    {
        return now()->isBetween($this->starts_at->startOfDay(), $this->ends_at->endOfDay());
    }
```

image

```

Comment (Roardom) [reply app/Http/Controllers/ClaimedPrizeController.php:35]: That's a *much* better idea. Thank you.

However, I'll probably do something like:

```php
$isAvailable = now()->isBetween($this->starts_at->startOfDay(), $this->ends_at->endOfDay());
```

and inline it into the controller so that it remains clear those columns are needed to be selected (We tend to avoid putting anything except casts/relations/etc into models)

PR #4296, opened 2024-10-31
(Refactor) Move welcome message to listener and add tests

IonBazan:
This PR moves the logic responsible for sending welcome message to newly registered user into a listener. This makes it much easier to test the registration flow as it no longer depends on `Bot` and `Chatroom` classes and allows to test that particular logic separately instead.

Using event listeners instead of controllers/actions makes it easier to decouple the logic and allows to conditionally run it in the future if needed. It is also easy to mock it using `Event::fake()` in tests.

### DIFF
--- a/app/Actions/Fortify/CreateNewUser.php
+++ b/app/Actions/Fortify/CreateNewUser.php
@@ -19,7 +19,6 @@ namespace App\Actions\Fortify;
 use App\Models\Group;
 use App\Models\Invite;
 use App\Models\User;
-use App\Repositories\ChatRepository;
 use App\Rules\EmailBlacklist;
 use Exception;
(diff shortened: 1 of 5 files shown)


HDVinnie:
@IonBazan great pull request. Fully agree with that changes here and really appreciate the test coverage which has been lacking unfortunately. Hopefully soon we can get a new core dev on board to manage tests and get our coverage 75% or greater. Thanks again.

HDVinnie (approved):

IonBazan:
Thanks for quickly merging the change. Glad to make a helpful contribution. I'll try to keep providing more similar improvements and hope to increase the coverage as the part of refactoring or fixing bugs too.

Issue #3763, opened 2024-04-18
[Request] Make PostgreSQL the Default Database System


EkoNesLeg:
I'd like to propose switching the default database to PostgreSQL, whilst retaining compatibility for htose who want to remian on MySQL. Postgre's better deadlock handling might address the daily deadlocks MySQL gets into on UNTI3D and it'd enable more granual control, most importatnly easy schema-specific restorations (a pain with MySQL). This aligns with the ongoing [borgmatic feature request](https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/3740) which would enable schema-specific restores directly from database dumps rather than only full db dumps/restorations.

Roardom:
TODO:

- Fix inline `@` variables in refund command
- Migrate `INTERVAL value unit` syntax to `INTERVAL 'value units'`
- Boolean expressions in aggregates need to be casted to ints (e.g. `SELECT COUNT(active AND visible) FROM peers`).

Find replacements for the following functions:

- `HEX`, `UNHEX`,
- `JSON_ARRAYAGG`, `JSON_OBJECT`,  `JSON_ARRAY`
- `SUBSTRING_INDEX`
- `UNIX_TIMESTAMP`, `DATE`, `FROM_DAYS`, `TO_DAYS`
- `INET6_NTOA`

Roardom:
Postgres does not support unsigned integer types, and laravel does not support check constraints, so this is a no go until one of the two can make themselves compatible.

PR #4300, opened 2024-11-01
(Refactor) Use ANSI-compatible syntax for raw database queries

Roardom:
In preparation for #3763 (and overall good practice)

The following still needs to be fixed before postgres is compatible:

- Fix inline `@` variables in refund command

Find replacements for the following functions:

- `HEX`, `UNHEX`,
- `JSON_ARRAYAGG`, `JSON_OBJECT`,  `JSON_ARRAY`
- `SUBSTRING_INDEX`
- `UNIX_TIMESTAMP`, `DATE`, `FROM_DAYS`, `TO_DAYS`
- `INET6_NTOA`


Roardom:
I runtime tested all the changed queries (except the migrations, but those are tested in ci already). We can worry about the remaining functions later in a "part 2" pr. The remaining functions will likely need to be duplicated in php checking the database system currently used and having 2 separate queries.

HDVinnie (approved):

PR #4310, opened 2024-11-02
Update tr user.php

kolankaya:
translate file en - tr


HDVinnie:
@kolankaya these should all be one big pull request and placed against the v8.x.x branch. Please read and follow the contribution guidelines

PR #4370, opened 2024-12-08
(Add) Allow internal options for internals in non-internal group

Obi-Wana:
Internals that are not member of the internal group (e.g., Editor) and not Moderator+ should still be able to set the internal options on their uploads. They need to be in at least one Internal group, checked using count of `InternalUsers`.

### DIFF
--- a/app/Http/Requests/StoreTorrentRequest.php
+++ b/app/Http/Requests/StoreTorrentRequest.php
@@ -44,6 +44,7 @@ class StoreTorrentRequest extends FormRequest
      */
     public function rules(Request $request): array
     {
+        $user = $request->user()->loadExists('internals');
         $category = Category::findOrFail($request->integer('category_id'));
 
         return [
@@ -223,19 +224,22 @@ class StoreTorrentRequest extends FormRequest
             'internal' => [
                 'sometimes',
                 'boolean',
-                Rule::when(!$request->user()->group->is_modo && !$request->user()->group->is_internal, 'prohibited'),
+                /** @phpstan-ignore property.notFound (Larastan doesn't yet support loadExists()) */
+                Rule::when(!$user->group->is_modo && !$user->internals_exists, 'prohibited'),
             ],
             'free' => [
                 'sometimes',
                 'integer',
                 'numeric',
                 'between:0,100',
-                Rule::when(!$request->user()->group->is_modo && !$request->user()->group->is_internal, 'prohibited'),
+                /** @phpstan-ignore property.notFound (Larastan doesn't yet support loadExists()) */
+                Rule::when(!$user->group->is_modo && !$user->internals_exists, 'prohibited'),
             ],
             'refundable' => [
                 'sometimes',
                 'boolean',
-                Rule::when(!$request->user()->group->is_modo && !$request->user()->group->is_internal, 'prohibited'),
+                /** @phpstan-ignore property.notFound (Larastan doesn't yet support loadExists()) */
+                Rule::when(!$user->group->is_modo && !$user->internals_exists, 'prohibited'),
             ],
         ];
     }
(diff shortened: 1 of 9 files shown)


Roardom (changes requested):
Great idea!

Can we use `$user->internals()->exists()` instead to make it more efficient?

Further, I think we can even drop the `is_internal` column on the `groups` table and use this relation as the source of truth instead for all cases the `is_internal` flag is checked. What do you think @HDVinnie?

Obi-Wana:
> Can we use `$user->internals()->exists()` instead to make it more efficient?

No, `->exists()` does not exist on this, just checked.


> Further, I think we can even drop the `is_internal` column on the `groups` table and use this relation as the source of truth instead for all cases the `is_internal` flag is checked. What do you think @HDVinnie?

I'll let this up to you, and think it should be going into its own PR.

HDVinnie:
LGTM. Just swap the count for exist.

HDVinnie (approved):

Roardom (changes requested):
When using `loadExists`, you need to specify the relation to load, i.e. `->loadExists('internals')`. Then, instead of checking for if `$user->internals()->exists(), you only have to check the `$user->internals_exists` property that's added to the user without needing further database calls.

Obi-Wana:
> When using `loadExists`, you need to specify the relation to load, i.e. `->loadExists('internals')`. Then, instead of checking for if `$user->internals()->exists()`, you only have to check the `$user->internals_exists` property that's added to the user without needing further database calls.

Added this to the APIController, but Larastan doesn't seem to like it...

Roardom:
I checked through the larastan repo and you're right: unfortunately larastan doesn't yet support relation_exists (although it does support relation_count).

Try adding `/** @phpstan-ignore property.notFound (Larastan doesn't yet support loadExists())` wherever the property is accessed.
Relevant documentation can be found here: https://phpstan.org/user-guide/ignoring-errors

app/Http/Requests/StoreTorrentRequest.php
```diff
@@ -44,6 +44,7 @@ public function authorize(): bool
      */
     public function rules(Request $request): array
     {
+        $user = $request->user();
```
Roardom, line 47: Can `->loadExists('internals')` and `->internals_exists` be used here as well?

resources/views/torrent/create.blade.php
Roardom, line 1: Can `->loadExists('internals')` and `->internals_exists` be used here as well (Inside `TorrentController::update()` at the bottom of the function)

Roardom (changes requested):

resources/views/torrent/create.blade.php
Obi-Wana: As far as I can tell, we'd need to change all `auth()->user()...` to use `$request->user()->loadExists...` from the `TorrentController::update()` function.

app/Http/Requests/StoreTorrentRequest.php
Obi-Wana: Added.

resources/views/torrent/create.blade.php
Roardom: Oops, sorry. I meant `TorrentController::create()`:

```php
            'user'         => $request->user()->loadExists('internals'),
```

However, that made me look again and realize you probably want to also add the same logic to `TorrentController::edit()` and `torrent/edit.blade.php`

Roardom (changes requested):
Can you do a ctrl+f in your IDE for `->group->is_internal` and replace all the instances with the `->internals()->exists()` logic? (And use eager loading if the query is used multiple times in a single http request).

Thanks!

Roardom (approved):

Roardom:
Thank you!

PR #4371, opened 2024-12-09
(Add) Latest torrent comments to homepage

Obi-Wana:
Add latest comments of type `Torrent` and `TorrentRequest` to the home page and display in a minimal list.

### DIFF
--- a/database/migrations/2024_12_09_175613_add_index_on_created_at_for_comments.php
+++ b/database/migrations/2024_12_09_175613_add_index_on_created_at_for_comments.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * NOTICE OF LICENSE.
+ *
+ * UNIT3D Community Edition is open-sourced software licensed under the GNU Affero General Public License v3.0
+ * The details is bundled with this project in the file LICENSE.txt.
+ *
+ * @project    UNIT3D Community Edition
+ *
+ * @author     HDVinnie <hdinnovations@protonmail.com>
+ * @license    https://www.gnu.org/licenses/agpl-3.0.en.html/ GNU Affero General Public License v3.0
+ */
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('comments', function (Blueprint $table) {
+            $table->index('created_at');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('comments', function (Blueprint $table) {
+            $table->dropIndex(['created_at']);
+        });
+    }
+};
--- a/database/migrations/2024_12_09_175613_add_index_on_created_at_for_comments.php
+++ b/database/migrations/2024_12_09_175613_add_index_on_created_at_for_comments.php
@@ -21,14 +21,14 @@ use Illuminate\Support\Facades\Schema;
 return new class () extends Migration {
     public function up(): void
     {
-        Schema::table('comments', function (Blueprint $table) {
+        Schema::table('comments', function (Blueprint $table): void {
             $table->index('created_at');
         });
     }
 
     public function down(): void
     {
-        Schema::table('comments', function (Blueprint $table) {
+        Schema::table('comments', function (Blueprint $table): void {
             $table->dropIndex(['created_at']);
         });
     }
--- a/resources/views/blocks/latest_comments.blade.php
+++ b/resources/views/blocks/latest_comments.blade.php
@@ -1,0 +1,18 @@
+<section class="panelV2 blocks__posts">
+    <h2 class="panel__heading">
+        <a href="{{ route('posts.index') }}">
+            {{ __('blocks.latest-comments') }}
+        </a>
+    </h2>
+    <div class="panel__body">
+        <ul class="comment-list">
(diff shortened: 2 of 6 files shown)


resources/views/blocks/latest_comments.blade.php
```diff
@@ -0,0 +1,22 @@
+<section class="panelV2 blocks__posts">
+    <header class="panel__header">
+        <h2 class="panel__heading">
+            <a href="{{ route('posts.index') }}">
+                {{ __('blocks.latest-comments') }}
+            </a>
+        </h2>
+    </header>
+    <div class="panel__body">
```
HDVinnie, line 9: Can we simplify this a bit to
```php
    <div class="panel__body">
        <ul class="comment-list">
            @forelse ($comments as $comment)
                <li class="comment__list-item">
                    <x-torrent.comment-listing :comment="$comment" />
                </li>
            @empty
                No comments.
            @endforelse
        </ul>
    </div>
```

resources/views/components/torrent/comment-listing.blade.php
```diff
@@ -0,0 +1,72 @@
...
+    <header class="comment__header">
+        <time
+            class="comment__datetime"
+            datetime="{{ $comment->created_at }}"
+            title="{{ $comment->created_at }}"
+        >
+            {{ $comment->created_at?->diffForHumans() }}
+        </time>
+
+        <span class="comment__topic">
```
HDVinnie, line 15: Thinking a switch here would be better. Something like
```php
<span class="comment__topic">
    On
    @switch(basename(str_replace('\\', '/', $comment->commentable_type)))
        @case('Torrent')
            Torrent
            <a href="{{ route('torrents.show', ['id' => $comment->commentable->id]) }}">
                {{ $comment->commentable->name }}
            </a>
            @break

        @case('TorrentRequest')
            Request
            <a href="{{ route('requests.show', ['torrentRequest' => $comment->commentable->id]) }}">
                {{ $comment->commentable->name }}
            </a>
            @break

        @default
            Unknown
    @endswitch
</span>
```

Use language files of course. `{{ __('torrent.torrent') }}`

HDVinnie (changes requested):

resources/views/components/torrent/comment-listing.blade.php
Roardom: Or actually probably the following since the fully qualified path to the related class name is saved:

```blade
@switch($comment->commentable_type)
    @case(\App\Models\Torrent::class)
```

resources/views/blocks/latest_comments.blade.php
Obi-Wana: Done

resources/views/components/torrent/comment-listing.blade.php
Obi-Wana: Done with:
> Or actually probably the following since the fully qualified path to the related class name is saved:
> 
> ```
> @switch($comment->commentable_type)
>     @case(\App\Models\Torrent::class)
> ```

Obi-Wana:
> Can you make a migration to add an index on the `created_at` field on the comments table?

Done that.

database/migrations/2024_12_09_175613_add_index_on_created_at_for_comments.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
```
Roardom, line 2: Can you add the copyright disclaimer here (can be copy pasted from any other file)?

resources/views/blocks/latest_comments.blade.php
```diff
@@ -0,0 +1,20 @@
+<section class="panelV2 blocks__posts">
+    <header class="panel__header">
```
Roardom, line 2: No need to surround it in a `header` if it doesn't have the action buttons.

resources/views/components/torrent/comment-listing.blade.php
```diff
@@ -0,0 +1,74 @@
...
+                    {{ __('torrent.torrent') }}
+                    <a href="{{ route('torrents.show', ['id' => $comment->commentable->id]) }}">
+                        {{ $comment->commentable->name }}
+                    </a>
+
+                    @break
+                @case(\App\Models\TorrentRequest::class)
+                    {{ __('request.request') }}
+                    <a
+                        href="{{ route('requests.show', ['torrentRequest' => $comment->commentable->id]) }}"
```
Roardom, line 29:
```suggestion
                        href="{{ route('requests.show', ['torrentRequest' => $comment->commentable]) }}"
```

resources/views/components/torrent/comment-listing.blade.php
```diff
@@ -0,0 +1,74 @@
...
+        @if (! empty($comment->user->title))
+            <p class="comment__author-title">
+                {{ $comment->user->title }}
+            </p>
+        @endif
+    </aside>
+    <div
+        class="comment__content bbcode-rendered"
+        x-ref="content"
+        data-base64-bbcode="{{ base64_encode($comment->content) }}"
```
Roardom, line 70: These two lines are only needed for the copy button scripts. You can remove them if you're not adding the button.

Roardom (changes requested):
A few last things

Thanks!

resources/views/components/torrent/comment-listing.blade.php
Obi-Wana: Done

PR #4374, opened 2024-12-11
(Add) Use shouldSend for notifications

Obi-Wana:
Use shouldSend for notifications.

### DIFF
--- a/app/Http/Livewire/Comments.php
+++ b/app/Http/Livewire/Comments.php
@@ -128,24 +128,27 @@ class Comments extends Component
         // New Comment Notification
         switch (true) {
             case $this->model instanceof Ticket:
-                $ticket = $this->model;
+                // Notify assigned staff if needed
+                User::find($this->model->staff_id)?->notify(new NewComment($this->model, $comment));
 
-                if ($this->user->id !== $ticket->staff_id && $ticket->staff_id !== null) {
-                    User::find($ticket->staff_id)?->notify(new NewComment($this->model, $comment));
-                }
-
-                if ($this->user->id !== $ticket->user_id) {
-                    User::find($ticket->user_id)?->notify(new NewComment($this->model, $comment));
-                }
+                // Notify ticket creator if needed
+                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
 
                 break;
             case $this->model instanceof Article:
+                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
+
+                break;
             case $this->model instanceof Playlist:
+                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
+
+                break;
             case $this->model instanceof TorrentRequest:
+                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
+
+                break;
             case $this->model instanceof Torrent:
-                if ($this->user->id !== $this->model->user_id) {
-                    User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
-                }
+                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
 
                 break;
         }
--- a/app/Http/Livewire/Comments.php
+++ b/app/Http/Livewire/Comments.php
@@ -136,17 +136,8 @@ class Comments extends Component
 
                 break;
             case $this->model instanceof Article:
-                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
-
-                break;
             case $this->model instanceof Playlist:
-                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
-
-                break;
             case $this->model instanceof TorrentRequest:
-                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
-
-                break;
             case $this->model instanceof Torrent:
                 User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
 
--- a/app/Notifications/NewBon.php
+++ b/app/Notifications/NewBon.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 namespace App\Notifications;
 
 use App\Models\Gift;
+use App\Models\User;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Notifications\Notification;
@@ -40,6 +41,29 @@ class NewBon extends Notification implements ShouldQueue
     public function via(object $notifiable): array
     {
         return ['database'];
+    }
+
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
+        // Enforce non-anonymous staff notifications to be sent
+        if ($this->gift->sender->group->is_modo) {
+            return true;
+        }
+
+        if ($notifiable->notification?->block_notifications == 1) {
+            return false;
+        }
+
+        if (!$notifiable->notification?->show_bon_gift) {
+            return false;
+        }
+
+        // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
+        // the expression will return false.
+        return ! \in_array($this->gift->sender->group_id, $notifiable->notification->json_bon_groups, true);
     }
 
     /**
--- a/app/Notifications/NewComment.php
+++ b/app/Notifications/NewComment.php
@@ -17,12 +17,12 @@ declare(strict_types=1);
 namespace App\Notifications;
 
 use App\Models\Article;
-use App\Models\Collection;
 use App\Models\Comment;
 use App\Models\Playlist;
 use App\Models\Ticket;
 use App\Models\Torrent;
 use App\Models\TorrentRequest;
+use App\Models\User;
 use Illuminate\Bus\Queueable;
 use Illuminate\Notifications\Notification;
 
@@ -33,7 +33,7 @@ class NewComment extends Notification
     /**
      * NewComment Constructor.
      */
-    public function __construct(public Torrent|TorrentRequest|Ticket|Playlist|Collection|Article $model, public Comment $comment)
+    public function __construct(public Torrent|TorrentRequest|Ticket|Playlist|Article $model, public Comment $comment)
     {
     }
 
@@ -45,6 +45,59 @@ class NewComment extends Notification
     public function via(object $notifiable): array
     {
         return ['database'];
+    }
+
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
+        // Do not notify self
+        if ($this->comment->user_id === $notifiable->id) {
+            return false;
+        }
+
+        // Enforce non-anonymous staff notifications to be sent
+        if ($this->comment->user->group->is_modo &&
+            ! $this->comment->anon) {
+            return true;
+        }
+
+        // Evaluate general settings
+        if ($notifiable->notification?->block_notifications == 1) {
+            return false;
+        }
+
+        // Evaluate model based settings
+        switch (true) {
+            case $this->model instanceof Torrent:
+                if (!$notifiable->notification?->show_torrent_comment) {
+                    return false;
+                }
+
+                // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
+                // the expression will return false.
+                return ! \in_array($this->comment->user->group_id, $notifiable->notification->json_torrent_groups, true);
+
+            case $this->model instanceof TorrentRequest:
+                if (!$notifiable->notification?->show_request_comment) {
+                    return false;
+                }
+
+                // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
+                // the expression will return false.
+                return ! \in_array($this->comment->user->group_id, $notifiable->notification->json_request_groups, true);
+
+            case $this->model instanceof Ticket:
+                return ! ($this->model->staff_id === $this->comment->id && $this->model->staff_id !== null)
+                ;
+
+            case $this->model instanceof Playlist:
+            case $this->model instanceof Article:
+                break;
+        }
+
+        return true;
     }
 
     /**
@@ -76,11 +129,6 @@ class NewComment extends Notification
                 'title' => 'New Playlist Comment Received',
                 'body'  => $username.' has left you a comment on Playlist '.$this->model->name,
                 'url'   => '/playlists/'.$this->model->id.'#comment-'.$this->comment->id,
-            ],
-            $this->model instanceof Collection => [
-                'title' => 'New Collection Comment Received',
-                'body'  => $username.' has left you a comment on Collection '.$this->model->name,
-                'url'   => '/mediahub/collections/'.$this->model->id.'#comment-'.$this->comment->id,
             ],
             $this->model instanceof Article => [
                 'title' => 'New Article Comment Received',
--- a/app/Notifications/NewComment.php
+++ b/app/Notifications/NewComment.php
@@ -78,7 +78,6 @@ class NewComment extends Notification
                 // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
                 // the expression will return false.
                 return ! \in_array($this->comment->user->group_id, $notifiable->notification->json_torrent_groups, true);
-
             case $this->model instanceof TorrentRequest:
                 if (!$notifiable->notification?->show_request_comment) {
                     return false;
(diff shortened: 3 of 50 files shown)


app/Notifications/NewRequestBounty.php
```diff
@@ -42,6 +43,32 @@ public function via(object $notifiable): array
...
+            return false;
+        }
+
+        if (\is_array($notifiable->notification->$targetGroup)) {
+            // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
+            // the expression will return false.
+            return !\in_array($this->bounty->user->group->id, $notifiable->notification->$targetGroup, true);
+        }
+
+        return true;
```
Roardom, line 69: We should be able to assume that it will always be an array (We made migrations to ensure this a few years ago). Also, it's better to directly access the UserNotification property `json_request_groups`. Also `->group->id` -> `->group_id`. Also remove `return @bool` since it's already indicated in the function signature. This goes for all the other notifications too.

```suggestion
     */
    public function shouldSend(User $notifiable): bool
    {
        if ($notifiable->notification?->block_notifications == 1) {
            return false;
        }

        if (!$notifiable->notification?->show_request_bounty) {
            return false;
        }

        // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
        // the expression will return false.
        return !\in_array($this->bounty->user->group_id, $notifiable->notification->json_request_groups, true);
```

app/Http/Livewire/Comments.php
```diff
@@ -128,24 +128,27 @@ final public function postComment(): void
...
-                    User::find($ticket->user_id)?->notify(new NewComment($this->model, $comment));
-                }
+                // Notify ticket creator if needed
+                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
 
                 break;
             case $this->model instanceof Article:
+                User::find($this->model->user_id)?->notify(new NewComment($this->model, $comment));
+
+                break;
```
Roardom, line 141: This code isn't necessary. The last 4 cases can all run the same code at the bottom.

app/Notifications/NewComment.php
```diff
@@ -47,6 +48,61 @@ public function via(object $notifiable): array
...
+                break;
+            case $this->model instanceof Ticket:
+                return ! ($this->model->staff_id === $this->comment->id && $this->model->staff_id !== null)
+                ;
+
+            case $this->model instanceof Playlist:
+            case $this->model instanceof Article:
+                return true;
+            case $this->model instanceof Collection:
+                break;
```
Roardom, line 87: It's not possible to notify on collection comments since they aren't owned by users. Probably best to remove this case, and, remove the Collection type from the constructor, and remove the message at the bottom too.

app/Notifications/NewComment.php
```diff
@@ -47,6 +48,61 @@ public function via(object $notifiable): array
...
+            default:
+                // In the unlikely case none matches, fallback to true to not lose a notification
+                return true;
```
Roardom, line 90: PHPStan should be able to guarantee us that the default case will never occur, so this can be removed.

app/Notifications/NewComment.php
```diff
@@ -47,6 +48,61 @@ public function via(object $notifiable): array
...
+
+        if (!$notifiable->notification?->$targetSetting) {
+            return false;
+        }
+
+        if (\is_array($notifiable->notification->$targetGroup)) {
+            // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
+            // the expression will return false.
+            return !\in_array($this->comment->user->group->id, $notifiable->notification->$targetGroup, true);
+        }
```
Roardom, line 101: Can these be checked directly inside the case statement

tests/Feature/Notifications/NewUploadNotificationTest.php
```diff
@@ -0,0 +1,361 @@
...
+        'movie_meta' => true,
+        'tv_meta'    => false,
+        'game_meta'  => false,
+    ]);
+    $type = Type::factory()->create([]);
+
+    $torrent = Torrent::factory()->make();
+    $name = 'Pony Music';
+    $filePath = base_path('tests/Resources/Pony Music - Mind Fragments (2014).torrent');
+    $file = new UploadedFile($filePath, 'Pony Music.torrent', 'text/plain', null, true);
```
Roardom, line 87: Did you try using `UploadedFile::fake()->createWithContent()`

Roardom (changes requested):
Thank you for all the tests! I just have a few syntax/style comments.

app/Http/Livewire/Comments.php
Obi-Wana: Done.

app/Notifications/NewComment.php
Obi-Wana: Done

app/Notifications/NewComment.php
Obi-Wana: Done.

app/Notifications/NewComment.php
Obi-Wana: Done.

tests/Feature/Notifications/NewUploadNotificationTest.php
Obi-Wana: No, I did not try that. Would this be preferred? If so, I can check if that would work.

app/Notifications/NewRequestBounty.php
Obi-Wana: Done most of your suggestions here too.

Unfortunately, with your code example, some tests keep failing (notifications are not sent tho they should).
When keeping the if:
```
        if (\is_array($notifiable->notification->json_bon_groups)) {
            // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
            // the expression will return false.
            return !\in_array($this->gift->sender->group_id, $notifiable->notification->json_bon_groups, true);
        }

        return true;
```
Larastan does not like the return true, as it thinks code will be terminated before that, but tests work again.

Obi-Wana: I ended up keeping the check for array. In some cases, the tests failed with something like `argument #2 (haystack) must be of type array, null given`.

Roardom: From a quick glance, it looks to be because the factory is generating a word instead of a json array of ints: https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/f554ae3acb204b705c0ccc48cf50098e8f7d8203/database/factories/UserNotificationFactory.php#L60-L67

tests/Feature/Notifications/NewUploadNotificationTest.php
Roardom: Yeah, it would be good to give it a quick go. Otherwise, I'm personally not too worried about it.

app/Notifications/NewPost.php
```diff
@@ -45,21 +45,28 @@ public function via(object $notifiable): array
...
+
+                break;
+            case 'topic':
+                $targetGroup = 'json_forum_groups';
+
+                break;
+            default:
+                $targetGroup = 'json_forum_groups';
+
+                break;
```
Roardom, line 69: Can this code maybe be inlined below? Something like the following:

```php
return match ($this->type) {
    'subscription' => !\in_array($this->user->group->id, $notifiable->notification->json_subscription_groups, true);
    'staff'        => true;
    default        => !\in_array($this->user->group->id, $notifiable->notification->json_forum_groups, true);
}
```

This should make the code a lot cleaner (once the factory's array property is fixed)

Roardom (changes requested):
One small improvement I noticed the second time around. Thanks!

app/Notifications/NewRequestBounty.php
Obi-Wana: With this and the fixed factory for `UserNotifications`, I managed to update all the Notifications and got all the tests passing again.

app/Notifications/NewUnfollow.php
```diff
@@ -42,6 +42,24 @@ public function via(object $notifiable): array
...
+            return false;
+        }
+
+        if (!$notifiable->notification?->show_account_unfollow) {
+            return false;
+        }
+
+        // If the sender's group ID is found in the "Block all notifications from the selected groups" array,
+        // the expression will return false.
+        return ! \in_array($this->unfollower->group_id, $notifiable->notification->json_following_groups, true);
```
Roardom, line 60: Should be `json_account_groups` to match the above and the settings page.

app/Notifications/NewPostTag.php
Roardom, line 1: Missing check to not send notification if sender and receiver are the same (e.g. if you quote a message that contains your own quote). The check is also done elsewhere, but probably best to do it here as well

app/Notifications/NewUpload.php
```diff
@@ -42,6 +43,33 @@ public function via(object $notifiable): array
...
+     * @return bool
+     */
+    public function shouldSend(User $notifiable): bool
+    {
+        $targetGroup = 'json_following_groups';
+
+        // Do not notify the uploader theirself
+        if ($this->torrent->user_id === $notifiable->id) {
+            return false;
+        }
```
Roardom, line 54: Can't follow yourself anyways, so we can probably remove this, similarly to how you've done elsewhere.

app/Notifications/NewRequestFill.php
```diff
@@ -42,6 +43,24 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
+        if ($notifiable->notification?->block_notifications == 1) {
+            return false;
```
Roardom, line 52: Should probably add the check to not notify yourself filling your own request, similar to how it's done before this pr.

app/Notifications/NewRequestFillReject.php
```diff
@@ -42,6 +43,24 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
+        if ($notifiable->notification?->block_notifications == 1) {
```
Roardom, line 51: Should add the check to not notify the filler if they're the rejecter

app/Notifications/NewRequestFillApprove.php
```diff
@@ -42,6 +43,24 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
```
Roardom, line 49: Should probably not notify the filler if they're the approver.

app/Notifications/NewRequestClaim.php
```diff
@@ -42,6 +43,24 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
+        if ($notifiable->notification?->block_notifications == 1) {
```
Roardom, line 51: Should probably not notify the requester if the claimer is the same user

app/Notifications/NewRequestUnclaim.php
```diff
@@ -42,6 +43,24 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
```
Roardom, line 50: Should probably not notify the requester if the unclaimer is the same user

app/Notifications/NewRequestBounty.php
```diff
@@ -42,6 +43,24 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
```
Roardom, line 50: Probably shouldn't notify requester if they add bounty to their own request.

app/Notifications/NewBon.php
```diff
@@ -42,6 +43,24 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
```
Roardom, line 50: Should have an is_modo check check to force send.

app/Notifications/NewComment.php
```diff
@@ -47,6 +47,48 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
```
Roardom, line 54: Should have an is_modo check check to force send (if comment is not anon).

app/Notifications/NewCommentTag.php
```diff
@@ -48,6 +49,55 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
```
Roardom, line 56: Should have an is_modo check check to force send (if comment is not anon).

app/Notifications/NewPostTag.php
```diff
@@ -42,6 +43,24 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     */
+    public function shouldSend(User $notifiable): bool
+    {
```
Roardom, line 50: Should have an is_modo check check to force send.

app/Notifications/NewUpload.php
```diff
@@ -42,6 +43,33 @@ public function via(object $notifiable): array
         return ['database'];
     }
 
+    /**
+     * Determine if the notification should be sent.
+     *
+     * @return bool
+     */
+    public function shouldSend(User $notifiable): bool
+    {
```
Roardom, line 50: Should have a check to not send if the uploader is anon.

Roardom (changes requested):
Thanks so much for sticking through with this so far (I really like how it's coming along). I noticed that some of the original logic for which fields are checked to determine if a notification should be sent were changed, seemingly unintentionally. Can you take a look?

app/Notifications/NewUnfollow.php
Obi-Wana: Corrected.

app/Notifications/NewPostTag.php
Obi-Wana: Added.

app/Notifications/NewUpload.php
Obi-Wana: Removed self part.

app/Notifications/NewRequestFill.php
Obi-Wana: Added self check.

app/Notifications/NewRequestFillReject.php
Obi-Wana: Added self check.

app/Notifications/NewRequestFillApprove.php
Obi-Wana: Added self check.

app/Notifications/NewRequestClaim.php
Obi-Wana: Aded self check.

app/Notifications/NewRequestUnclaim.php
Obi-Wana: Added self check.

app/Notifications/NewRequestBounty.php
Obi-Wana: Added self check.

app/Notifications/NewBon.php
Obi-Wana: Added

app/Notifications/NewComment.php
Obi-Wana: Added.

app/Notifications/NewCommentTag.php
Obi-Wana: Added.

app/Notifications/NewPostTag.php
Obi-Wana: Added.

app/Notifications/NewUpload.php
Obi-Wana: Added

Obi-Wana:
> I think it improves the code quality if you separate each check into its own if statement instead of combining the and's and the or's.

> E.g. this:
```
        if ($this->bounty->user_id === $notifiable->id ||
            $notifiable->notification?->block_notifications == 1) {
            return false;
        }
```

> would become this:
```
        if ($this->bounty->user_id === $notifiable->id) {
            return false;
        }

        if ($notifiable->notification?->block_notifications == 1) {
            return false;
        }
```

Roardom (approved):
Thank you!

PR #4380, opened 2024-12-16
(Add) Edit forum permissions when creating/editing group

Roardom:
Also a few touchups to checkboxGrid component to select rowspans, individual cells, and disable selecting column headers (when double clicking for example).

### DIFF
--- a/app/Http/Requests/Staff/UpdateGroupRequest.php
+++ b/app/Http/Requests/Staff/UpdateGroupRequest.php
@@ -42,150 +42,174 @@ class UpdateGroupRequest extends FormRequest
         $group = $request->route('group');
 
         return [
-            'name' => [
+            'group.name' => [
                 Rule::when(! $group->system_required, [
                     'required',
                     'string',
                 ]),
-                Rule::prohibitedIf($group->system_required && $request->name !== $group->name),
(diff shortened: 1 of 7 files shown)


HDVinnie (approved):

HDVinnie:
good one thanks

app/Http/Requests/Staff/UpdateGroupRequest.php
```diff
@@ -42,151 +42,175 @@ public function rules(Request $request): array
...
 
         return [
-            'name' => [
+            'group.name' => [
                 Rule::when(! $group->system_required, [
                     'required',
                     'string',
                 ]),
-                Rule::prohibitedIf($group->system_required && $request->name !== $group->name),
+                Rule::prohibitedIf($group->system_required && $request->group->name !== $group->name),
```
Slikkster, line 50: This rule gives a 500 error when attempting to update a system required group, because `$request->group` is an array. Correct syntax would be `$request->group['name']`.

Roardom: Thanks for pointing this out. I remember fixing this exact error, but not sure where my fix went...

PR #4384, opened 2024-12-17
(Fix) Every home block needs to have its own class

Roardom:


resources/views/blocks/latest_comments.blade.php
```diff
@@ -1,4 +1,4 @@
-<section class="panelV2 blocks__posts">
+<section class="panelV2 blocks__comments">
     <h2 class="panel__heading">
         <a href="{{ route('posts.index') }}">
```
HDVinnie, line 3: This is wrong too.

HDVinnie (changes requested):

HDVinnie (approved):

PR #4388, opened 2024-12-19
Adds cover_url and banner_url to API , fixes blades

sometimes-coding:
Expects url hence `cover_url` and `banner_url`, no change has been made to `http/TorrentController.php` which could probably benefit from this change as well (though you may want to add it as additional option rather than replacement) 

Considerations: Imagick needed for compression (if enabled) , better support for varying filetypes while preserving alpha channels if applicable. 

Conversion to .webp its old news but lets optimize storage, load times and leverage capabilities to  include transparency for applicable no_meta categories. (yes, I've made sure it can grab .jpg for backward compatibility) 

____

Whats with this UrlHelper?? 

I'm a php baby, so might be a better way to implement this. The idea is to try to keep things read-only so externally hosted is great. However not so reliable. DMCA's anyone? So while whitelists is great for hosting screenshots and whatnot, having nuked covers and banners might not be ideal. Thus this would be a helper where you can configure trusted external hosts (that wont be here today, gone tomorrow) and just display 'em if trusted. Else, we can just download the image, compress and serve it. Obviously safeguards need to be in place so this feature should be able to be turned off. Also could skip download if content-header not known to prevent abuse.

Hope this helps.

### DIFF
--- a/app/Helpers/UrlHelper.php
+++ b/app/Helpers/UrlHelper.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace App\Helpers;
+
+class UrlHelper
+{
+    /**
+     * Array of trusted URLs or domains.
+     *
+     * @var array
+     */
+    private static $trustedUrls = [
+        'imdb.com', 'imdb.org', 'imdb.co', 'imdb.idv', 'imdbshow.com', 'media-imdb.com',
+        'archive.org',
+        'coverartarchive.org',
+        'dvdempire.com', 'adultdvdempire.com', 'adultempire.com',
+        'cduniverse.com',
+
+    ];
+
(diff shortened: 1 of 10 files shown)


bitchcrazy1:
this would be a welcomed change so bots can upload cover pics when doing music etc

Vallesuski:
This is totally a good improvement to the API when uploading content not related to Movies/TV.

Roardom:
Hi, there were significant recent changes to the way files are stored in UNIT3D. Do you mind rebasing your changes? Also, I don't think your urlhelper is necessary anymore as images are only viewable when logged in (If I'm understanding your problem correctly)

PR #4406, opened 2025-01-03
(Update) Make bonus point allocation more customizable

Roardom:
Take 2 of #2411


database/migrations/2024_12_26_031647_bon_earnings_v2.php
```diff
@@ -0,0 +1,280 @@
...
+            [
+                'bon_earning_id' => 11,
+                'operand1'       => 'seedtime',
+                'operator'       => '>=',
+                'operand2'       => 1 * 30 * 24 * 3600,
+            ],
+        ]);
+    }
+
+    public function down(): void
```
HDVinnie, line 275: You can drop the down method.

HDVinnie (changes requested):

database/migrations/2024_12_26_031647_bon_earnings_v2.php
Roardom: Done. And rebased.

HDVinnie (approved):

PR #4409, opened 2025-01-04
(Update) Mediainfo Flags

ninboy:
Adds `es-CL` and `es-VE` flags, the other two dub-producing countries I know of in Latin America


HDVinnie:
Thanks for the contribution

PR #4433, opened 2025-01-19
add: category column and category icons to helpdesk

AnabolicsAnonymous:
This adds a category column header to the helpdesk index, updates the priorities icons to match the font awesome theme, and also adds icons matching the default categories.


Roardom (commented):
I think it would be a better idea to add another column to the ticket categories table to save the icon, similar to the categories table for example.

HDVinnie:
I agree with @Roardom here. It will keep consistency with other tables like types, categories, groups, etc. Feel free to open a new PR. Thank you.

PR #4424, opened 2025-01-12
(Add) Quoting in Livewire Comments

clandestine8:
Ported the Quoting Implementation from the Forums to Livewire Comments so Torrent Comments and any other comments using Livewire Comments can be quoted. Functionally identical to Forums.

This should close this issue: https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/4385

PR #4435, opened 2025-01-19
(Fix) Remove htmlspecialchars from comment quote function

AnabolicsAnonymous:
When replying to a message that has any special characters like quotes or ampersands it would html encode them in plain text, this removes that operation to make it work properly.

### DIFF
--- a/resources/views/components/forum/post.blade.php
+++ b/resources/views/components/forum/post.blade.php
@@ -110,16 +110,7 @@
                             document.getElementById('forum_reply_form').style.display = 'block';
                             input = document.getElementById('bbcode-content');
                             input.value += '[quote={{ \htmlspecialchars('@' . $post->user->username) }}]';
-                            input.value += (() => {
-                                var text = document.createElement('textarea');
-                                text.innerHTML = decodeURIComponent(
-                                    atob($refs.content.dataset.base64Bbcode)
-                                        .split('')
-                                        .map((c) => '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2))
-                                        .join('')
-                                );
-                                return text.value;
-                            })();
+                            input.value += decodeURIComponent(escape(atob('{{ base64_encode($post->content) }}')));
                             input.value += '[/quote]';
                             input.dispatchEvent(new Event('input'));
                             input.focus();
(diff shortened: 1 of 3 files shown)


Roardom (changes requested):
We should use the logic that's already used for forum posts: https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/c64275f0b5dcb3c4c845d5204871adfe24f359d6/resources/views/components/forum/post.blade.php#L109-L126

I figured there was a high chance something would break when modifying #4424

AnabolicsAnonymous:
> We should use the logic that's already used for forum posts:
> 
> https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/c64275f0b5dcb3c4c845d5204871adfe24f359d6/resources/views/components/forum/post.blade.php#L109-L126
> 
> I figured there was a high chance something would break when modifying #4424

Assumed that's what he would've used, what exactly would we be taking out of this function that isn't already here though?

Roardom:
> > We should use the logic that's already used for forum posts:
> > https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/c64275f0b5dcb3c4c845d5204871adfe24f359d6/resources/views/components/forum/post.blade.php#L109-L126
> > 
> > I figured there was a high chance something would break when modifying #4424
> 
> Assumed that's what he would've used, what exactly would we be taking out of this function that isn't already here though?

Okay now you've made me unsure of the way to go about this properly :) (I'm not a js person). I've fixed issues similar to this in the past such as #2933 and #3165. Also this commit 6bd691c9217eee3f473eec6dc562478fc0976d84 where it was undone looks like it was quite intentional for it to be added.

But looking into this further and it seems that the solution implemented in #4424 is deprecated in js, but it still works, and it bypasses the need for the alternative I used in the previous solutions of creating a textarea, putting the text in, and it solving the decoding aspect on its own. So in that sense, it'd be a good idea to change the other 2 locations where this is used (resources/views/components/forum/post.blade.php, and resources/views/torrent/partials/description.blade.php) to match what you've implemented here, alongside the escape changes. Would you be willing to do that?

(fwiw, the root problem surrounding all these issues is that atob returns bytes in js, but js strings use utf-16)

Roardom (approved):

PR #4468, opened 2025-02-07
(Add) Middle language codes

Blackflexxy:
Currently, mediainfo only shows language codes for these languages, it may have to be changed again in the future if mediainfo starts writing the full name like mkvtoolnix does.
middle languages


Roardom:
On further consideration and after speaking with others, none of these codes are used in the wild, nor should they refer to their modern-day flags. Therefore, I don't think this PR should be merged.

Blackflexxy:
>  nor should they refer to their modern-day flags

At first I wanted to add a flag for enm only, but I saw that there are more middle language codes, so I added all of them that have flags. If you think that new flags should be created for these languages, then I'd ask to keep at least enm, because from time to time I see these subtitles in anime releases from MTBB, which are essentially simplified English, and I think the us flag can be used in this case

Roardom:
> I see these subtitles in anime releases from MTBB, which are essentially simplified English

They use them for honorifics, which is not simplified English, nor is it Middle English. So, I still don't think that's correct.

PR #4484, opened 2025-02-16
(Update) Hide 'flush ghost peers' button when external tracker is enabled

Roardom:
The button doesn't do anything anyways for the external trackers. Users have to just wait it out.


HDVinnie (changes requested):
Same should be done for flush peers button in staff dashboard.

HDVinnie (approved):

PR #4485, opened 2025-02-17
(Fix) Migrate boolean columns to tinyint(1)

Roardom:
These columns were previously tiny ints without the (1) display width or regular ints, or smallints. They should all be tinyint(1).


HDVinnie:
Conflicts

PR #4548, opened 2025-03-11
(Update) Use "Powered by" instead of "Built using" in the footer

Roardom:
This language is more commonly used.


HDVinnie:
I almost feel the words should be removed all together. Just icons. Or just remove it all.

PR #4559, opened 2025-03-16
(Refactor) Use unsigned int for igdb

Roardom:
Use the correct type for the external id.

Step 3 in cleaner meta fetching code.


HDVinnie (commented):
Shouldn't this be for requests too?

Roardom:
> Shouldn't this be for requests too?

Yes, it should. Thank you.

PR #4656, opened 2025-04-14
(Update) Always notify uploader/requester on replies

Roardom:
But don't send two notifications, one to the top level commenter, and one to the uploader, if they happen to be the same user. Keep the logic that notifies the top level commenter of any replies within.


HDVinnie:
Any reason this is still draft?

PR #4676, opened 2025-04-27
(Fix) Topics notifications search

AnabolicsAnonymous:
Seems like this button wasn't formatted the same with the bunch which caused it to not do a Livewire search when ticked.


HDVinnie (approved):

HDVinnie:
thank you

PR #4681, opened 2025-04-29
(Fix) donor expiries overlapping instead of extending

AnabolicsAnonymous:
Apologies for the dupe #4673 still getting used to git.

> My last PR regarding this was abandoned and removed after I reset the fork, that being said this pull request updates the functionality of the donation approval so that when a user with an active donation decides to add another donation, instead of simply adding another donation to the table with the same values, it checks for all previous **active** donations and appends their remaining expiry to the new donation.

### DIFF
--- a/app/Http/Controllers/Staff/DonationController.php
+++ b/app/Http/Controllers/Staff/DonationController.php
@@ -74,16 +74,30 @@ class DonationController extends Controller
         $donation->status = ModerationStatus::APPROVED;
         $donation->starts_at = $now;
 
-        if ($donation->package->donor_value > 0) {
-            $donation->ends_at = $now->addDays($donation->package->donor_value);
-        } else {
+        if ($donation->package->donor_value === null) {
             $donation->ends_at = null;
+        } else {
+            $activeDonations = Donation::where('user_id', $donation->user_id)
+                ->where('status', ModerationStatus::APPROVED)
+                ->where('ends_at', '>', $now)
+                ->get();
+
+            if ($activeDonations->isNotEmpty()) {
+                if ($donation->user->is_lifetime) {
+                    $donation->ends_at = null;
+                } else {
+                    $furthestExpiry = $activeDonations->max('ends_at');
+                    $donation->ends_at = $furthestExpiry->addDays($donation->package->donor_value);
+                }
+            } else {
+                $donation->ends_at = $now->addDays($donation->package->donor_value);
+            }
         }
 
         $donation->user->invites += $donation->package->invite_value ?? 0;
         $donation->user->uploaded += $donation->package->upload_value ?? 0;
         $donation->user->is_donor = true;
-        $donation->user->is_lifetime = $donation->package->donor_value === null;
+        $donation->user->is_lifetime = $donation->ends_at === null;
         $donation->user->seedbonus += $donation->package->bonus_value ?? 0;
         $donation->user->save();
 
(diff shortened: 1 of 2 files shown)


Roardom (changes requested):
Instead of modifying the existing donation, can you please create a new donation that starts at the time the newest one ends and modify the logic in the AutoRemoveExpiredDonors.php command to match?

Do you mind also updating the docblock of Donation.php to match the database schema? It seems some of the nullable properties were missed when it was first added.

AnabolicsAnonymous:
> Can you please ensure the donation starts at the time the furthest out donation ends and modify the logic in the AutoRemoveExpiredDonors.php command to match (use both `<` and `>`)?

Could you clarify what needs to be changed in `AutoRemoveExpiredDonors.php`, I think the logic should be fine the way it is but clearly I'm not very good at reading that or maybe there's different reasoning going over my head?

Roardom:
> > Can you please ensure the donation starts at the time the furthest out donation ends and modify the logic in the AutoRemoveExpiredDonors.php command to match (use both `<` and `>`)?
> 
> Could you clarify what needs to be changed in `AutoRemoveExpiredDonors.php`, I think the logic should be fine the way it is but clearly I'm not very good at reading that or maybe there's different reasoning going over my head?

The logic should check if the current time is between the starts_at and ends_at, instead of only checking the ends_at.

AnabolicsAnonymous:
> The logic should check if the current time is between the starts_at and ends_at, instead of only checking the ends_at.

I'm so confused on this one, are you looking for this? That second check doesn't make sense to me so I'm just verifying.
```php
        $expiredDonors = User::where('is_donor', '=', true)
            ->where('is_lifetime', '=', false)
            ->whereHas('donations')
            ->whereDoesntHave('donations', function ($query): void {
                $query->where('ends_at', '>', Carbon::now());
                      ->where('starts_at', '<', Carbon::now());
            })->get();

File: app/Http/Controllers/Staff/DonationController.php [hunk 1]
```diff
@@ -74,16 +74,31 @@ public function update(Request $request, int $id): \Illuminate\Http\RedirectResp
...
-        if ($donation->package->donor_value > 0) {
-            $donation->ends_at = $now->addDays($donation->package->donor_value);
-        } else {
+        if ($donation->package->donor_value === null) {
             $donation->ends_at = null;
+        } else {
+            $activeDonations = Donation::where('user_id', $donation->user_id)
+                ->where('status', ModerationStatus::APPROVED)
+                ->where('ends_at', '>', $now)
+                ->get();
```
Roardom, line 83: If a user has 10000+ donations, you don't need them all. You only need the one that has the max ends_at (and to check if it's not null).

app/Http/Controllers/Staff/DonationController.php
```diff
@@ -74,16 +74,31 @@ public function update(Request $request, int $id): \Illuminate\Http\RedirectResp
         $donation->status = ModerationStatus::APPROVED;
         $donation->starts_at = $now;
```
Roardom, line 75:
```suggestion
```

Instead of setting it here and overwriting it in specific cases later, it would be good if it was explicitly set inside each edge case to make things more clear.

app/Http/Controllers/Staff/DonationController.php
Roardom, line 83: In the supposed case that a user has donated 10000+ times, we don't need to pull all those records into memory. Can you query just the record with the latest (or null) ends_at instead? Then comparing it to null can be used for if there's any active donation or not.

Roardom (changes requested):
A couple more things

app/Http/Controllers/Staff/DonationController.php
Roardom: Oops I reviewed this twice

Issue #3897, opened 2024-06-03
api poster uploads [Request]


BitPyrate:
We are starting have a thriving unit3d api scripting community that has found a home on our tracker. 

they are requesting api upload support for
Table:movies
column:poster
to link to a poster offsite or perhaps upload a poster 

suggestion:
perhaps a condition when tmdb and the other meta sources = 0

z-ink:
Hi, ignorant to php code here: but as the [http controller](https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/master/app/Http/Controllers/TorrentController.php) has provision in lines 227-247 and again in lines 463-483 

`     // Cover Image for No-Meta Torrents
        if ($request->hasFile('torrent-cover')) {
            $image_cover = $request->file('torrent-cover');

            abort_if(\is_array($image_cover), 400);

            $filename_cover = 'torrent-cover_'.$torrent->id.'.jpg';
            $path_cover = public_path('/files/img/'.$filename_cover);
            Image::make($image_cover->getRealPath())->fit(400, 600)->encode('jpg', 90)->save($path_cover);
        }

        // Banner Image for No-Meta Torrents
        if ($request->hasFile('torrent-banner')) {
            $image_cover = $request->file('torrent-banner');

            abort_if(\is_array($image_cover), 400);

            $filename_cover = 'torrent-banner_'.$torrent->id.'.jpg';
            $path_cover = public_path('/files/img/'.$filename_cover);
            Image::make($image_cover->getRealPath())->fit(960, 540)->encode('jpg', 90)->save($path_cover);
        }`
        
        

> The api however has no access to post a 'torrent-cover' or 'torrent-banner', categories that are unlinked to TMDb would greatly benefit from this. 

_Quoted above because it kept throwing it into the code block_

Roardom:
The movie and tv tables are strictly linked to tmdb currently. You can change the meta on the category to "no meta", and then you can upload your own banner/poster and it will be saved on the individual torrent. If you find content missing on tmdb, I would recommend adding it so that others can benefit. Currently the unit3d api doesn't accept those parameters, but if it did, it would only be allowed for categories with "no meta" selected. That being said, a PR to add the missing functionality to the api of uploading poster/banner for torrents with the "no meta" category would be appreciated.

BitPyrate:
The main point would be to upload to the "No meta" categories with a poster. right now if you mass upload to a no meta group you have to manually go back and add it. 
[...]

Issue #3688, opened 2024-03-25
[Request] API Update - Cover File for Music


MaddyTP:
Would it be possible to add cover file similar to nfo file to the API?  Currently unable to fully automate uploading music since cover file isn't included in API post request.

Neodkbits:
Hi! 

What i do is...(workaround)

i just have a python script renaming and ftp uploading to \var\www\html\public\files\img\ so the cover_image.jpg is name torrent_cover_54563.jpg and so on

torrentid and filenameid should match.. 
then it will be shown as cover :)

PR #4682, opened 2025-04-30
(add) Support for Cover and Banner Image URLs

sometimes-coding:
# Add Support for Cover and Banner Image URLs 

## Description
This PR introduces support for uploading cover and banner images via URLs or files for `no_meta` and `music_meta` torrents, enhancing flexibility for users. It centralizes image processing, improves validation, and adds WebP support with fallback to JPG.

## Key Changes
- Added `cover_url` and `banner_url` fields to the `Torrent` model and API responses.
- Introduced `ImageHelper` class for centralized image processing, handling both uploaded files and remote URLs.
- Implemented validation for image files and mutual exclusivity between file uploads and URLs.
- Configured remote image handling with a `whitelist_or_save` option, allowing whitelisted URLs to be used directly or saving images to the server otherwise. (Else `save_to_server` or `use_url` configurable in `config\Image.php`)
- Updated views and controllers to support new image fields for `no_meta` and `music_meta` torrents.
- Added cleanup of unused images on torrent deletion to prevent storage bloat.
- Enhanced `AuthenticatedImageController` to prioritize WebP files with JPG fallback.
- Included migration to add `cover_url` and `banner_url` columns to the `torrents` table.
- Fixed phpStan errors and ran prettier on Blade templates via CI.

## Motivation
- Provide the API with ability to handle both torrent-cover and torrent-banner as `Http\TorrentController.php` does
- Provide users with the option to use external image URLs instead of uploading files, reducing server load for trusted sources.
- Improve image handling consistency and support modern WebP format for better performance.
- Enhance the user experience by supporting both `no_meta` and `music_meta` categories.

## Testing
- Tested uploading cover and banner images via files and URLs for `no_meta` and `music_meta` torrents.
- Verified mutual exclusivity validation (file vs. URL) in `StoreTorrentRequest` and `UpdateTorrentRequest`.
- Confirmed WebP conversion and fallback to JPG for stored images.
- Checked remote image handling with whitelisted and non-whitelisted URLs.
- Ensured unused images are deleted when torrents are removed.
- Validated UI updates in torrent creation/edit views and image display in torrent cards/rows.

## Dependencies
__None Added__ but for clarity..
- Requires the `intervention/image` package for image processing.
- Depends on the `WhitelistedImageUrl` model for URL whitelisting.

## Related Issues
Fixes Issue #3897 and Issue #3688 
Complete overhaul of Pull #4388, rebased. Dropped needless dependencies. Also added URL support for the site.

## Considerations
It is recommended that the API primarily use `cover_url` and `banner_url `for image input. By default, PHP has an `upload_max_filesize` limit of 2MB, but Nginx often enforces a stricter `client_max_body_size` of 1MB. This means larger uploads may be rejected by Nginx before Laravel validation can occur. To avoid these issues, API users should avoid sending files via `torrent-cover` and `torrent-banner`, and instead rely on the URL-based fields.

Using WebP is also recommended for its flexibility and efficiency. WebP can significantly reduce file sizes—for example, compressing a 10MB PNG to under 1MB—making it especially useful if users choose to self-host all images.

## Checklist
- [x] Tests pass locally and in CI.
- [x] Migration included for database changes.
- [x] Reviewed for security (e.g., URL validation, file type checks).

### DIFF
--- a/app/Helpers/ImageHelper.php
+++ b/app/Helpers/ImageHelper.php
@@ -21,15 +21,17 @@ use Illuminate\Http\UploadedFile;
 use Illuminate\Support\Facades\Http;
 use Intervention\Image\Facades\Image;
 use Illuminate\Support\Facades\Storage;
+use Exception;
+use Log;
 
 class ImageHelper
 {
     /**
      * Process an image (file or URL) for torrent cover or banner.
      *
-     * @param UploadedFile|string $source Image file or URL
-     * @param string $torrentId Torrent ID for naming
-     * @param string $type 'cover' or 'banner'
+     * @param  UploadedFile|string                        $source    Image file or URL
+     * @param  string                                     $torrentId Torrent ID for naming
+     * @param  string                                     $type      'cover' or 'banner'
      * @return array{path: string|null, url: string}|null
      */
     public static function processTorrentImage($source, string $torrentId, string $type): ?array
@@ -49,11 +51,12 @@ class ImageHelper
         }
 
         if (!self::isValidImageUrl($source)) {
-            \Log::error("Invalid image URL in processTorrentImage", [
-                'url' => $source,
+            Log::error("Invalid image URL in processTorrentImage", [
+                'url'        => $source,
                 'torrent_id' => $torrentId,
-                'type' => $type,
+                'type'       => $type,
             ]);
+
             return null;
         }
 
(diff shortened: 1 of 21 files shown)


Roardom (changes requested):
Hi, thanks for the PR. I know you've put in a lot of work into this, but this might be one of the few things that we (as maintainers) don't agree should be part of UNIT3D. Specifically the ability to provide a url and the site fetches the images on its own. We'd rather have the user be responsible for uploading the image themselves (and the sysop should adjust the nginx config as necessary) with server side validation. There is a lot of complexity added here that will be difficult to maintain. A small list of things I have issues with:

- Exposes server ip address
- Defaults to webp
- Complexity in the maze of possible paths taken during execution

I would like to see the issues linked in your PR fixed, but I'd rather see them fixed by simply copy-pasting the ~30 lines of code that makes them possible in the web-based `TorrentController::store`, and have it mirrored to the `\API\TorrentController::store`, using the same names for the fields that the web form already uses. Will that suffice?

PR #4691, opened 2025-05-06
(Add) Event deletion condition

HDVinnie:
- One cannot delete an event where users have already claimed a prize. Foreign key constraint stops this. So we implement a condition to prevent this and return message to mark the event as inactive instead.


Roardom (changes requested):
Can you use `to_route()` instead?

Roardom (approved):

PR #4699, opened 2025-05-08
(Update) Implement proper tmdb rate limiting

Roardom:
Allows the fetch meta command to work without hitting rate limits on tmdb. Tmdb's rate limit is ~40 per second, but each movie/tv has to send additional http request for each collection, network, company and person, so I figured 2 per second was safe.


EkoNesLeg:
Tested and works well.

PR #4703, opened 2025-05-11
(Add) Componentize nav-bar

LostRager:


resources/views/partials/nav-bar.blade.php
Roardom, line 1: Can you rename this to `/resources/views/torrents/partials/_nav-tabs.blade.php` instead?

Roardom (changes requested):
Great refactor. Thank you. I just have one request.

Roardom (changes requested):
Ideally all partials should be local to the subfolder of the pages including them. Even if rss and top10 are under different pages. (i.e. under the `/resources/views/torrents/partials` directory) I'll be going through all the views at some point to standardize this approach.

HDVinnie:
Stale

PR #4704, opened 2025-05-11
(Fix) Comparison image size when posted in chatbox

LostRager:


resources/sass/components/_chatbox.scss
```diff
@@ -104,7 +104,7 @@
     padding: 0 8px 8px 8px;
     font-size: 16px;
 
-    img {
+    img:not(.comparison img) {
```
Roardom, line 107:
```suggestion
    img:not(.comparison__image) {
```

Roardom (changes requested):
Thanks for the PR. My only suggestion is to use the class name directly.

PR #4705, opened 2025-05-11
(Fix) Comparison blocking arrow keys from working in chat

LostRager:
Makes it so the arrow keys are only blocked when comparison is opened.


Roardom (approved):
So *that's* what's been breaking chatbox arrow keys every now and then. Great find!

Roardom:
Could you also please run prettier before committing?

```sh
./node_modules/.bin/prettier --write ./resources/views/partials/comparison.blade.php
```

Roardom (approved):

PR #4716, opened 2025-05-20
(Fix) RSS Genres

BradleyRL:
It had the old table


HDVinnie:
Thank you for the contribution.

PR #4719, opened 2025-05-24
(Add) user chatbox message stats

AnabolicsAnonymous:
This adds a user stats page listing the top 100 users by the number of messages they sent in the chatbox so users can see who chats the most.


Roardom (changes requested):
Thanks. Can you modify the query to exclude private chatbox messages, as well as those belonging to the system user?

app/Http/Controllers/StatsController.php
```diff
@@ -354,4 +354,22 @@ public function themes(): \Illuminate\Contracts\View\Factory|\Illuminate\View\Vi
...
         ]);
     }
+
+    /**
+     * Show Extra-Stats User Messages.
+     */
+    public function messages(): \Illuminate\Contracts\View\Factory|\Illuminate\View\View
+    {
+        $users = User::withCount(['messages' => function ($query): void {
+            $query->where('chatroom_id', '!=', 0);  // exclude chatroom_id 0/PMs;
```
Roardom, line 364:
```suggestion
            $query->where('chatroom_id', '!=', 0);  // Exclude private chatbox messages;
```

app/Http/Controllers/StatsController.php
```diff
@@ -354,4 +354,22 @@ public function themes(): \Illuminate\Contracts\View\Factory|\Illuminate\View\Vi
...
+        }])
+            ->orderByDesc('messages_count')
```
Roardom, line 366:
```suggestion
            ->orderByDesc('messages_count')
            ->where('id', '!=', User::SYSTEM_USER_ID)
```

resources/views/stats/users/messages.blade.php
```diff
@@ -0,0 +1,55 @@
...
+    <li class="breadcrumb--active">
+        {{ __('common.users') }}
+    </li>
+@endsection
+
+@section('nav-tabs')
+    @include('partials.statsusermenu')
+@endsection
+
+@section('page', 'page__stats--message-count')
```
Roardom, line 22:
```suggestion
@section('page', 'page__stats--messages')
```

Roardom (changes requested):
Just a few nitpicks

app/Http/Controllers/StatsController.php
AnabolicsAnonymous: Guess I didn't think about bots being set under a different group, thanks.

app/Http/Controllers/StatsController.php
```diff
@@ -354,4 +354,23 @@ public function themes(): \Illuminate\Contracts\View\Factory|\Illuminate\View\Vi
...
+    /**
+     * Show Extra-Stats User Messages.
+     */
+    public function messages(): \Illuminate\Contracts\View\Factory|\Illuminate\View\View
+    {
+        $users = User::withCount(['messages' => function ($query): void {
+            $query->where('chatroom_id', '!=', 0);  // Exclude private chatbox messages;
+        }])
+            ->orderByDesc('messages_count')
+            ->where('id', '!=', User::SYSTEM_USER_ID)
```
Roardom, line 367: You'll have to `use App\Models\User;` at the top of the file as well

app/Http/Controllers/StatsController.php
Roardom, line 1: You'll have to update the phpstan baseline to fix that ci error:

> Ignored error pattern #^Anonymous function should return Illuminate\\Database\\Eloquent\\Builder\<Illuminate\\Database\\Eloquent\\Model\> but returns Illuminate\\Database\\Eloquent\\Builder\<App\\Models\\Group\>\.$# (return.type) in path /home/runner/work/UNIT3D/UNIT3D/app/Http/Controllers/StatsController.php is expected to occur 3 times, but occurred 4 times.

It's caused by larastan/phpstan not supporting the `whereDoesntHave` method and you adding an additional instance of that method to the file.

Roardom (changes requested):
A few CI issues remaining

Roardom (approved):
Thank you!

PR #4739, opened 2025-06-02
(Update) Use Modern Apexcharts on Donation Stats

AnabolicsAnonymous:
I've always been bugged by the donation graph since it's buggy and looks out of place with unit3d, this replaces the chart.js with apexcharts and adds a dropdown menu to select the time frame. It's highly reactive and mobile friendly while looking 🤤.

Let the screenshot speak for itself:
Screenshot 2025-06-01 234837


Roardom (changes requested):
I'm not entirely sure if Apexcharts is actually better enough for it to be worth to switch it out (I'll leave that to vinnie to approve), but just a few suggestions to match UNIT3D's current structure. 

I'm also not a JS dev, but the code looks overly complicated for what it does. This code really only needs to be given certain x/y values and not need for much else.:

- Why are you setting custom styles? Are the defaults not good enough? Setting a fontFamily to undefined doesn't make much sense instead of omitting it altogether (I haven't read the docs mind you, it just seems like it shouldn't be necessary)
- You also specified a custom date format. There's an international standard that already exists - please follow that instead.
- The colors should also not be hardcoded, but use the CSS variables that all the themes already provide. If you need more colours, please find good colours for each theme and add them to all stylesheets. This is so that external stylesheets can also modify the colors.

HDVinnie:
I have looked at apex charts in past. I was going to review this PR locally but see it's closed now? So no longer wanted or?

PR #4740, opened 2025-06-02
(Add) Turnstile

AnabolicsAnonymous:
While everybody LOVES the hidden captcha, it's got some major downsides that ruin the users experience. For one it frequently invalidates users especially on applications where they may be filling it out for a while and lose their progress only after clicking apply. Also it's not really a captcha but a repellent, which is why I think it's valuable for sites to invest in free turnstile for their domains and use it with unit3d. Turnstile is highly secure and cant be easily bypassed and it will prompt the user to reverify should there be any issues before continuing to the submit form. This is applied on all unauthenticated pages and only requires the following fields to be added to the sites environment:

```ini
CAPTCHA_TYPE=turnstile # by default it's the hidden captcha
TURNSTILE_SITE_KEY=0x00000000000000000 # these keys can be obtained from cloudflares dashboard, all it requires is a FQDN
TURNSTILE_SECRET_KEY=0x000000000000000
```
image

### DIFF
--- a/app/Actions/Fortify/CreateNewUser.php
+++ b/app/Actions/Fortify/CreateNewUser.php
@@ -58,7 +58,7 @@ class CreateNewUser implements CreatesNewUsers
             ],
             'captcha' => [
                 Rule::excludeIf(config('captcha.enabled') === false),
-                Rule::when(config('captcha.enabled') === true, 'hiddencaptcha'),
+                Rule::when(config('captcha.enabled') === true, 'captcha'),
             ],
             'code' => [
                 Rule::when(config('other.invite-only') === true, [
--- a/app/Helpers/TurnstileCaptcha.php
+++ b/app/Helpers/TurnstileCaptcha.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Helpers;
+
+use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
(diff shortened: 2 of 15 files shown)


Roardom (changes requested):
I strongly oppose this. Not only is this privacy-invasive and not directly hosted, but cloudflare's captchas are beyond broken. I have to solve them daily for things that should absolutely not be behind captchas, and do not want that experience inside UNIT3D.

AnabolicsAnonymous:
> I strongly oppose this. Not only is this privacy-invasive and not directly hosted, but cloudflare's captchas are beyond broken. I have to solve them daily for things that should absolutely not be behind captchas, and do not want that experience inside UNIT3D.

Being directly hosted sounds great and all there should at least be a proper bot fight that doesn't have a low CSR and only serve as a roadblock for real users and practically useless against scripts. I haven't had any negative experiences with turnstile captchas since most of the time it's either no-click or one click and done for the one time I'd need to login.

Roardom:
> mysterious

It's caused by specific password managers filling in a field they're not supposed to

AnabolicsAnonymous:
Could've sworn the network tab showed that field as empty but maybe not the best idea to have a field that browsers can fill with username in the tag? It's just quirks like this that make the hidden captcha pointless if the way to fail is by using a password manager while any jimbob can waltz in and scrape 3 random strings and slip in.

PR #4779, opened 2025-06-17
(Refactor) Simplify UpdateLastAction middleware

Roardom:
Removes second return statement.


HDVinnie (changes requested):
Can we leave $user = $request->user outside the condition as technically placing assignments in if conditions is bad practice.

Roardom:
> Can we leave $user = $request->user outside the condition as technically placing assignments in if conditions is bad practice.

Sure. Updated.

HDVinnie (approved):

PR #4780, opened 2025-06-17
(Update) Allow requesting torrents with any resolution or type

Roardom:


resources/views/requests/create.blade.php
```diff
@@ -90,13 +90,9 @@ class="form__option"
...
-                        <select
-                            id="resolution_id"
-                            class="form__select"
-                            name="resolution_id"
-                            required
-                        >
-                            <option hidden disabled selected value=""></option>
+                        <select id="resolution_id" class="form__select" name="resolution_id">
+                            <option selected value="">Any</option>
+                            <option value="">Any</option>
```
HDVinnie, line 95: This should be removed?

resources/views/requests/show.blade.php
```diff
@@ -82,16 +82,22 @@
...
-            </li>
-            <li class="request__resolution">
-                <span>
-                    {{ $torrentRequest->resolution->name ?? 'No Res' }}
-                </span>
-            </li>
+            @if ($torrentRequest->type !== null)
+                <li class="request__type">
+                    <span>
+                        {{ $torrentRequest->type->name }}
```
HDVinnie, line 88: Instead of wrapping in condition why not use ?? 'Any' like already do in other places.

resources/views/requests/show.blade.php
```diff
@@ -82,16 +82,22 @@
...
+                    </span>
+                </li>
+            @endif
+
+            @if ($torrentRequest->resolution !== null)
+                <li class="request__resolution">
+                    <span>
+                        {{ $torrentRequest->resolution->name }}
```
HDVinnie, line 96: Same here

HDVinnie (changes requested):

HDVinnie (approved):

PR #4781, opened 2025-06-17
(Add) Store unread news notifications

Roardom:


HDVinnie (changes requested):
I feel like UnreadNews is a bad choice. UnreadArticle make more sense since base model is Article.

HDVinnie (approved):

PR #4784, opened 2025-06-18
(Refactor) Chatbox from Vue2 to AlpineJS

HDVinnie:
- This fully removes Vue from UNIT3D.


resources/sass/components/_chatbox.scss
```diff
@@ -245,3 +272,64 @@
...
+        transform: scale(1);
+        opacity: 1;
+    }
+}
+
+.spinner__text {
+    font-size: 16px;
+    color: #65a6c2;
+    font-weight: 500;
+}
```
Roardom, line 335: Can this be added to a separate spinner component?

Roardom (changes requested):
Wonderful job, thank you. Haven't tested, but read through all the code. I like it. I assumed while reading it that most of the logic is copied identically from the original .vue so I didn't mark down any bad js logic I saw. Only saw one thing where there was new code that could be improved.

HDVinnie:
Merging even with failing CI as its due to article system changes and has nothing to do with chat PR. @Roardom

PR #4787, opened 2025-06-19
(Refactor) Create zips in-memory instead of on filesystem

HDVinnie:


HDVinnie:
Merging even with failing CI as its due to article system changes and has nothing to do with this PR. @Roardom

PR #4794, opened 2025-06-20
(Refactor) Use laravel-data for typed form requests and DTOs

Roardom:
Just a draft right now, but I'm very impressed by this package. The developer experience is much improved with the typed form requests, and matches behaviour common in other type-safe languages. It allows accessing properties directly off the dto that came from the form request with ide autocomplete, and it validates that the properties exist and are not null. It might also be the solution to fixing the livewire type errors where a user supplies a string to an int variable as well.


Roardom:
Use https://www.github.com/laravel/framework/pull/58676 instead once it's released.

Roardom:
Unfortunately laravel tabled that PR :(

PR #4822, opened 2025-07-11
(Fix) Chatbox pms not showing up

Roardom:
Doesn't affect every user. Logs from an affected user suggest that there's a duplicate key in the x-for. Untested since I can't reproduce the error to begin with.


Roardom:
Affected users have reported this fix works

PR #4823, opened 2025-07-12
(Update) Simultaneous chatter typing indicator

AnabolicsAnonymous:
This formats typing indicators to support multiple users talking so instead of flicking between users typing, it shows user1 and user2 are typing, and concatenates with commas if more than two chatters type.

### DIFF
--- a/resources/js/components/alpine/chatbox.js
+++ b/resources/js/components/alpine/chatbox.js
@@ -121,10 +121,10 @@ const channelHandler = {
             })
             .listenForWhisper('typing', (e) => {
                 if (context.state.chat.target > 0 || context.state.chat.bot > 0) return;
-                const name = e.username || e;
-                clearTimeout(context.activePeer.get(name));
-                const t = setTimeout(() => context.activePeer.delete(name), 15000);
-                context.activePeer.set(name, t);
+                const username = e.username || e;
+                clearTimeout(context.activePeer.get(username));
+                const messageTimeout = setTimeout(() => context.activePeer.delete(username), 15000);
+                context.activePeer.set(username, messageTimeout);
             });
 
         context.channel.error((error) => {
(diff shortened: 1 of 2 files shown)


Roardom (commented):
Great change, thank you.

What do you think about using a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead of an array?

resources/js/components/alpine/chatbox.js
```diff
@@ -121,11 +121,10 @@ const channelHandler = {
...
             .listenForWhisper('typing', (e) => {
                 if (context.state.chat.target > 0 || context.state.chat.bot > 0) return;
-                context.activePeer = e;
-                clearTimeout(context.typingTimeout);
-                context.typingTimeout = setTimeout(() => {
-                    context.activePeer = false;
-                }, 15000);
+                const name = e.username || e;
+                clearTimeout(context.activePeer.get(name));
+                const t = setTimeout(() => context.activePeer.delete(name), 15000);
```
Roardom, line 126: Any potential of using a few more letters in the variable name?

resources/views/blocks/chat.blade.php
```diff
@@ -404,12 +404,16 @@ class="chatroom-users__button"
...
-                    ></span>
-                </span>
+                    x-show="state.chat.target < 1 && state.chat.bot < 1 && activePeer && activePeer.size > 0"
+                    x-text="
+                        activePeer.size === 1
+                            ? [...activePeer.keys()][0] + ' is typing ...'
+                            : [...activePeer.keys()].slice(0, -1).join(', ') +
+                              ' and ' +
+                              [...activePeer.keys()][activePeer.size - 1] +
+                              ' are typing ...'
```
Roardom, line 414: It would be extra nice to add some logic here so that when there's more than three people typing, it could change to "Several people are typing...". Also can the space be removed before the ellipsis?

Roardom (changes requested):
Looking good! Just a couple of nitpicks

resources/views/blocks/chat.blade.php
AnabolicsAnonymous: I've added my change for this that looks right to me but I can't get more than 2-3 people to type at a time to work on my environment.

Roardom (approved):
Thank you!

PR #4829, opened 2025-07-15
(Update) Improve appearance of homepage block settings

Roardom:
Remove unnecessary class that adds padding between each checkbox, and remove repetition of wording from each line.


HDVinnie:
Done in #4831

PR #4848, opened 2025-07-18
Style: User list toggle & chat textbox

AnabolicsAnonymous:
A few style tweaks, the user list toggle background color on the navy theme was missed when I added it, there's two pointless scrollbars that show up on the user list toggle that this removes, leaving only the vertical main one. The chat message form now only resizes vertically and the textbox label gets a background when it's focused on so it doesn't overlap with the textbox border.

### DIFF
--- a/resources/sass/components/_chatbox.scss
+++ b/resources/sass/components/_chatbox.scss
@@ -48,7 +48,8 @@
 }
 
 #chatbox__messages-create:focus ~ .form__label--floating,
-#chatbox__messages-create:valid:not(:placeholder-shown) ~ .form__label--floating {
+#chatbox__messages-create:valid:not(:placeholder-shown)
+    ~ .form__label--floating {
     background-color: var(--panel-bg);
 }
 
--- a/resources/sass/components/_chatbox.scss
+++ b/resources/sass/components/_chatbox.scss
@@ -47,6 +47,11 @@
     max-height: 60vh;
 }
 
+#chatbox__messages-create:focus ~ .form__label--floating,
+#chatbox__messages-create:valid:not(:placeholder-shown) ~ .form__label--floating {
+    background-color: var(--panel-bg);
+}
+
 .chatbox-message {
     display: grid;
     grid-template-areas: 'aside header . menu' 'aside content content content';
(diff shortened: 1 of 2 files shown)


resources/sass/components/_chatbox.scss
```diff
@@ -184,8 +192,9 @@
 
 .chatroom-users__list {
     list-style-type: none;
-    overflow-y: scroll;
+    overflow-y: visible;
     padding: 0 0 0 20px;
+    overflow-x: hidden;
```
Roardom, line 197: This section doesn't actually need overflow at all since it's already handled in `.chatroom__users`. Can you remove both `overflow-y` and `overflow-x`?

Roardom (changes requested):
Thanks, just a couple of touch ups

resources/sass/components/_chatbox.scss
AnabolicsAnonymous: Thanks, `overflow-y` is redundant but removing `overflow-x` makes the bottom horizontal scrollbar appear, but that is only there because the gift and message icons aren't accounted for in the pane width.

AnabolicsAnonymous: I think it's because Firefox doesn't support the custom scrollbars.

resources/sass/components/_chatbox.scss
```diff
@@ -42,9 +42,17 @@
...
+    width: 100%;
     max-height: 60vh;
 }
 
+#chatbox__messages-create:focus ~ .form__label--floating,
+#chatbox__messages-create:valid:not(:placeholder-shown)
+    ~ .form__label--floating {
+    background-color: var(--panel-bg);
+}
+
```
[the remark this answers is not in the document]
AnabolicsAnonymous: Isn't inherit what's already being used? Using inherit here keeps the same issue where the border overlaps the text, doesn't work the same as the forum textboxes.

AnabolicsAnonymous:
Issues stem from different web browsers and style issues that I just can't figure out out, closing this one.

Roardom:
The one to only allow the chatbox to resize vertically and be full width was a good change.

The inheritance is sorta how the labels were designed. It has to inherit the background color behind the outline to be over top the outline. Just need to add `background-color: inherit` to all layers between the nearest `.panelV2` and the label. The background-color shouldn't be manually defined further up.

PR #4860, opened 2025-07-20
(Update) BBCode parser

Roardom:
Moved all bbcodes into the parser array. Moved callback support into the array. Automatically wrap list items in lists when users forget to. Properly escape bbcodes inside of code blocks.


HDVinnie:
CI needs work

PR #4872, opened 2025-07-28
(Add) configurable healthcheck badge for UNIT3D-Announce in footer

EkoNesLeg:
Option to display a health check badge in footer if `announce_healthcheck_badge_url` config is set.

Suggested use is UNIT3D-Announce (UP/DOWN); e.g. 

unit3d-announce

### See also

- https://shields.io/
- https://healthchecks.io/


EkoNesLeg:
Only tested on UNIT3D-Announce, not the PHP tracker. The URL constructed by the Python checker is

```bash
curl -X GET "<TRACKER_BASE_URL>/<PASSKEY>?info_hash=<INFO_HASH>&peer_id=<PEER_ID>&port=<PORT>&uploaded=<UPLOADED>&downloaded=<DOWNLOADED>&left=<LEFT>&numwant=<NUMWANT>" -H "User-Agent: HealthCheck/1.0"
```

If this is query is incompatible with the PHP tracker, then the config key name should be moved inside the `external_tracker `key in `config/announce`.

PR #4889, opened 2025-08-04
(Add) Store torrent deletion reason

Roardom:
To be later used on the unregistered info hash page, in the tracker's torrent deletion message for bittorrent clients and perhaps a work edit log.


Roardom:
This PR has now been revamped

app/Http/Controllers/Staff/TorrentDeletionReasonController.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Store a new torrent deletion reason.
+     */
+    public function store(StoreTorrentDeletionReasonRequest $request): \Illuminate\Http\RedirectResponse
+    {
+        TorrentDeletionReason::query()->create($request->validated());
+
+        return to_route('staff.torrent_deletion_reasons.index')
+            ->with('success', 'Type successfully added');
```
HDVinnie: Should probably say reason and not type?

app/Http/Controllers/Staff/TorrentDeletionReasonController.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Edit a torrent deletion reason.
+     */
+    public function update(UpdateTorrentDeletionReasonRequest $request, TorrentDeletionReason $torrentDeletionReason): \Illuminate\Http\RedirectResponse
+    {
+        $torrentDeletionReason->update($request->validated());
+
+        return to_route('staff.torrent_deletion_reasons.index')
+            ->with('success', 'Type successfully modified');
```
HDVinnie: Should probably say reason and not type?

app/Http/Controllers/Staff/TorrentDeletionReasonController.php
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @throws Exception
+     */
+    public function destroy(DestroyTorrentDeletionReasonRequest $request, TorrentDeletionReason $torrentDeletionReason): \Illuminate\Http\RedirectResponse
+    {
+        $torrentDeletionReason->torrents()->update($request->validated());
+        $torrentDeletionReason->delete();
+
+        return to_route('staff.torrent_deletion_reasons.index')
+            ->with('success', 'Type successfully deleted');
```
HDVinnie: Should probably say reason and not type?

HDVinnie (changes requested):

HDVinnie:
I like this. Three small comments. I do, however, think a seeder should be added before merging. Even if it's just some common default ones. Each site can add their own and remove if they don't like the defaults.

app/Http/Controllers/Staff/TorrentDeletionReasonController.php
Roardom: Fixed, thanks

Roardom:
> I like this. Three small comments. I do, however, think a seeder should be added before merging. Even if it's just some common default ones. Each site can add their own and remove if they don't like the defaults.

Fixed, thanks. Seeder added. Took some inspiration from some gazelle sites.

Roardom (commented):
A torrent often has more than one deletion reason. It should be a many-to-many relationship instead.

PR #4905, opened 2025-08-18
(Refactor) Replace multiple eloquent `->when()` with `match`

Roardom:
Only just found out this was a thing. Looks really nice.


Roardom:
Converting to draft until I can figure out the CI

PR #4919, opened 2025-08-23
(Fix) Rotten Tomatoes search generation

yafbot:
The previous form was not URL encoded meaning that a title with "special" characters like "&" would result in a broken URL that lead to confusing destinations.

Warning: This code isn't tested at all.

### DIFF
--- a/app/Helpers/Helpers.php
+++ b/app/Helpers/Helpers.php
@@ -56,6 +56,16 @@ if (!\function_exists('href_request')) {
     }
 }
 
+if (!\function_exists('href_rottentomatoes')) {
+    function href_rottentomatoes(string|null $title, string|null $date): string
+    {
+        $year = substr($date ?? '', 0, 4);
+        $query = "{$title}  ({$year}) site:rottentomatoes.com";
+
+        return "https://html.duckduckgo.com/html/?q=\\".rawurlencode($query);
+    }
+}
+
 if (!\function_exists('href_poll')) {
     function href_poll(App\Models\Poll $poll): string
     {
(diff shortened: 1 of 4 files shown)


Roardom (changes requested):
Thanks for catching this bug.

I really don't think a global function is needed for this very small task used in only two closely-related locations. Worst case scenario, a laravel accessor (https://laravel.com/docs/12.x/eloquent-mutators#defining-an-accessor) could be used on the movie/tv models. Can you either look into doing that, or just inlining the rawurlencode?

Also first_air_date is only available for tv, you have to use release_date for movies.

Thanks!

app/Helpers/Helpers.php
```diff
@@ -56,6 +56,16 @@ function href_request(App\Models\TorrentRequest $torrentRequest): string
     }
 }
 
+if (!\function_exists('href_rottentomatoes')) {
+    function href_rottentomatoes(string|null $title, string|null $date): string
+    {
+        $year = substr($date ?? '', 0, 4);
+        $query = "{$title}  ({$year}) site:rottentomatoes.com";
```
yafbot, line 63: In the future it may be nice to omit the `()` if there is no year. But for the initial version I just wanted to match existing behaviour with correct parameter encoding.

Roardom:
In my opinion, 3 instances is the cut off point before requiring abstracting out (and not abstracted to a global function, but something more contained), but Vinnie seems to be fine with it.

Globals generally are a bad programming practice and I'd like to refactor them all out at some point (probably to laravel accessors or property hooks on their respective models).

PR #4930, opened 2025-09-01
(Update) Set comparison colors on material themes to white

AnabolicsAnonymous:
Black is hardly visible on dark material design themes and impossible to see on amoled, this sets the dark material theme ones to white to match the other themes.


HDVinnie:
Thanks

Issue #3976, opened 2024-07-09
Anon Forum posts [Request]


BitPyrate:
users can post torrents Anon however they can not post in the forum anon. 
Can that be added. Some feel like they can be a bit more open with their comments if they don't have to worry about singled out. abuse is mitigated in that Admins and perhaps mods can see who the person is.

HDVinnie:
will be in v9.1.7

PR #4935, opened 2025-09-02
(Add) Anon forum posts

HDVinnie:
- Request #3976


app/Notifications/NewPost.php
```diff
@@ -90,25 +90,29 @@ public function shouldSend(User $notifiable): bool
      */
     public function toArray(object $notifiable): array
```
Roardom, line 91:
```suggestion
    public function toArray(User $notifiable): array
```

Roardom (changes requested):
4chan called and wants their anons back

PR #4940, opened 2025-09-07
(Update) Use glow effect around peer counts as torrent activity indicator

Roardom:
Requires less screen space, and easier to tell from a glance.


HDVinnie:
Conflicts

PR #4941, opened 2025-09-07
(Refactor) Warning expire to bundled notification

### LINKED ISSUE #4868: [Request] Bundle expired warning notifications into single notification
The AutoDeactivateWarning command should bundle all expired warnings into a single email like the AutoWarning command, Instead of sending an email per warning like it is now.

HDVinnie:
- closes #4868


Roardom (changes requested):
UserManualWarningExpire is dead code now.

Roardom (changes requested):
Since it's being changed anyways...

Also I think keeping it singular is best, that way a user doesn't assume all their warnings have expired.

PR #4960, opened 2025-09-12
(Fix) Authenticated Image Controller needs to bypass torrent ApprovedScope.

tetrahydroc:
When a mod queue enabled user uploads a torrent and  adds a custom cover for a non movie/tv meta category as a mod i would expect to be able to view that cover image correctly when reviewing the torrent. Currently it 404s due to the  global ApprovedScope on the torrent model. This changes the AuthenticatedImageController to bypass that scope when retrieving images the same way the TorrentController handles accessing unapproved torrents.

### DIFF
--- a/app/Http/Controllers/AuthenticatedImageController.php
+++ b/app/Http/Controllers/AuthenticatedImageController.php
@@ -19,6 +19,7 @@ namespace App\Http\Controllers;
 use App\Models\Article;
 use App\Models\Category;
 use App\Models\Playlist;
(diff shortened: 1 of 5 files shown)


HDVinnie:
@tetrahydroc thanks for the pr but can you please make sure to run `bunx prettier -w *` and `./vendor/bin/pint` against your code when contributing to make sure it follows UNIT3Ds code style. Also always open PR's against development branch.

Roardom (approved):

PR #4985, opened 2025-09-19
(Refactor) Save request last bounty timestamp separately

Roardom:
created_at is the wrong field to save this value in because the initial created_at date is lost. Using updated_at also doesn't work since updating the request fields will bump the request even if more bounty isn't added. So, we use a new field for this info. We use a field on the requests table and not the request_bounty table so that we can use an index to sort.


HDVinnie:
Honestly the ability to bump requests by adding bounty is to abused and should be removed. Users can just sort by bounty amount or new last bountied at field.

HDVinnie (approved):

PR #4986, opened 2025-09-19
(Fix) Load thanks count after thanking on torrent page

Roardom:
Livewire doesn't reload the relations when the button is clicked and the number disappears.


HDVinnie:
@Roardom we should really just convert this to an alpine component like what was don't with like, dislike and bookmark.

Roardom:
> @Roardom we should really just convert this to an alpine component like what was don't with like, dislike and bookmark.

Agreed. Especially with alpine's new CSP-compatibility. Probably worth fixing this temporarily as is until that's done though?

PR #4992, opened 2025-09-20
(Fix) Use correct model ID for my requests filtering

AnabolicsAnonymous:
Currently the checkbox does nothing and it errors out in console.


HDVinnie:
Thank you

PR #4993, opened 2025-09-20
(Fix) Only load reseed requests that have an existing torrent

AnabolicsAnonymous:


HDVinnie:
Honestly I missed adding the foreign key restraint but they don't work anyway with soft deletes since it's not actually deleted. That being said what was truly missed is adding the relation delete https://github.com/HDInnovations/UNIT3D/blob/master/app/Http/Controllers/TorrentController.php around line 375.

My thoughts is we should create new migration that adds the foreign and also deletes any reseeds that don't have a torrent along with adding the relation delete in controller to properly fix this. @Roardom thoughts?

PR #5027, opened 2025-10-03
(Update) CI

flowerey:
Updates actions. Please tell me if cspell needs testing.


Roardom:
I still need to go through all the releases notes of the ci actions since these are major breaking versions, but can you please use only a single commit per change? There should be no more than one commit per action updated.

PR #5050, opened 2025-10-08
(Fix) Perms given in docs for storage/views and bun

Roardom:
Some of these perms need to be changed to avoid errors when running `sudo php artisan set:all_cache && sudo systemctl restart php-fpm && sudo php artisan queue:restart`.


Roardom:
@EkoNesLeg Can you review when you have a chance please?

PR #5076, opened 2025-10-21
(Fix) Meter styles on chromium and safari browsers

Roardom:
It's not a perfect match from the intended styles in firefox, but it's better than the early 2000s styles that were there before.


flowerey:
Thanks :)

PR #5094, opened 2025-10-29
(Add) Chatbox Mention Feature

sebhal92:
Added mention icon like here: https://github.com/HDInnovations/UNIT3D/issues/5087

As Roardom said 1 feature as a time.

### DIFF
--- a/resources/views/blocks/chat.blade.php
+++ b/resources/views/blocks/chat.blade.php
@@ -262,6 +262,49 @@
                                                     class="fal fa-star text-gold"
                                                     title="Donor"
                                                 ></i>
+												<button
+													type="button"
+													class="bbc-btn mention-btn"
+													title="Mention user"
+													style="
+														font-size: 13.5px; 
+														font-family: inherit; 
+														background: #383868; 
+														color: #ffe266; 
+														border-radius: 4.2px; 
+														margin-left: 5px; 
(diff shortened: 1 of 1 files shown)


flowerey:
Hey can you change the PR title to something better?

Roardom (changes requested):
Thanks for the pr! Sorry for the delay. It seems like everybody and anybody is contributing this week :)

I notice that you're using the alpinejs directives inline within the chat, which is already alpinejs. It'd be better if you can just add another function, perhaps below the existing commands that add the `/gift` and `/msg` commands: https://github.com/HDInnovations/UNIT3D/blob/6929af868b3d0cc9212e1aaad78c0caf9775f030/resources/js/components/alpine/chatbox.js#L706-L720

A similar style of button would also be good (right now your button styles don't match the existing buttons). You can probably use the existing class: https://github.com/HDInnovations/UNIT3D/blob/6929af868b3d0cc9212e1aaad78c0caf9775f030/resources/sass/components/_chatbox.scss#L230-L238. You should also be able to use `:hover` css styles instead of javascript for the hover colors. You might also need to use css variables for the colors and add them to all the themes, since using white for hover might only look good for the dark themes, but not the light themes. It also looks like your existing `bbc-btn mention-btn` classes aren't defined anywhere. Maybe a remnant from a previous draft?

PR #5099, opened 2025-10-31
(Add) Implement support for external IRC announce services

zze0s:
This PR implements support for external IRC announce services, such as https://codeberg.org/ze0s/UNIT3D-Announcer which Im currently doing a lot of updates to. Variants of it has been used for a couple of years without issues.

This implementation is sort of generic but it does hardcode the http path, and how it does auth and the JSON payload format. It supports connecting via both tcp/http and over unix domain socket.

The implementation is heavily inspired by the unit3d announce class.

I went for a new separate implementation to not introduce any extra complexity in the existing IRC Bot implementation.

Currently untested, so I'm starting this as a draft.

Have previously discussed with @Roardom about implementing this.

### DIFF
--- a/app/Bots/IRCAnnounceBotExternal.php
+++ b/app/Bots/IRCAnnounceBotExternal.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
 
 namespace App\Bots;
 
+use App\Models\IgdbGame;
 use App\Models\TmdbMovie;
 use App\Models\TmdbTv;
 use App\Models\Torrent;
@@ -65,6 +66,7 @@ class IRCAnnounceBotExternal
             $meta = match (true) {
                 $category->tv_meta    => TmdbTv::find($torrent->tmdb_tv_id),
                 $category->movie_meta => TmdbMovie::find($torrent->tmdb_movie_id),
+                $category->game_meta  => IgdbGame::find($torrent->igdb),
                 default               => null,
             };
         }


EkoNesLeg:
Thank you!

zze0s:
I've done some testing with a local dev setup, using the unix_socket and it works as expected! Or it doesn't set [meta](https://github.com/zze0s/UNIT3D/blob/686c1779c51bb246252ea0e1e9780d24f3fb763a/app/Bots/IRCAnnounceBotExternal.php#L93-L102) and release_year seem to be 0, but that's probably due to no TMDB setup locally?

Marking it ready for review anyway :partying_face:

app/Bots/IRCAnnounceBotExternal.php
```diff
@@ -0,0 +1,207 @@
...
+                    $leechTypeEnum = 7;
+
+                    break;
+                case 3:
+                    $leechTypeEnum = 8;
+
+                    break;
+                case 4:
+                    $leechTypeEnum = 9;
+            }
```
Roardom:
```suggestion
            $leechTypeEnum = match ($leechTypeEnum) {
                0 => 5,
                1 => 6,
                2 => 7
                3 => 8,
                4 => 9,
            };
```

app/Bots/IRCAnnounceBotExternal.php
```diff
@@ -0,0 +1,207 @@
...
+
+        $originEnum = 0;
+
+        if ($torrent->internal) {
+            $originEnum = 1;
+        }
+
+        if ($torrent->personal_release) {
+            $originEnum = 2;
+        }
```
Roardom:
```suggestion
        $originEnum = match (true) {
            $torrent->personal_release => 2,
            $torrent->internal => 1,
            default => 0,
        };
```

app/Bots/IRCAnnounceBotExternal.php
```diff
@@ -0,0 +1,207 @@
...
+            'category_name'          => $category->name,
+            'origin_enum'            => $originEnum,
+            'leech_type_enum'        => $leechTypeEnum,
+            'upload_time_unix_epoch' => $torrent->created_at->getTimestamp(),
+            'freeleech'              => (bool) $torrent->free > 0,
+            'freeleech_percent'      => $torrent->free,
+            'double_up'              => $torrent->doubleup,
+            'resolution'             => $torrent->resolution?->name ?? '',
+            'type'                   => $torrent->type->name,
+            'release_year'           => isset($torrent->meta->release_date) ? $torrent->meta->release_date->format('Y') : (isset($torrent->meta->first_air_date) ? $torrent->meta->first_air_date->format('Y') : null),
```
Roardom: It looks like you intended to use the `$meta` variable here instead of `$torrent->meta`.

config/irc-bot-external.php
```diff
@@ -0,0 +1,81 @@
...
+    */
+
+    'is_enabled' => false,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Host
+    |--------------------------------------------------------------------------
+    |
+    | Host takes an ip/localhost as http endpoint. The api key is not required, but recommended.
```
Roardom:
```suggestion
    | Host takes an ip/localhost as http endpoint.
```

It looks like the apikey is already described further down

app/Bots/IRCAnnounceBotExternal.php
Roardom: Looks like `internal` is currently improperly casted as an int instead of a bool unfortunately so `match` won't work yet, but easy fix to cast it. However, would that break this? https://github.com/Prowlarr/Indexers/blob/52fc4eddc26f730a11d956238082afa6f23237c7/definitions/v11/blutopia-api.yml#L143-L147

app/Bots/IRCAnnounceBotExternal.php
```diff
@@ -0,0 +1,207 @@
...
+        if ($torrent->tmdb_movie_id > 0 || $torrent->tmdb_tv_id > 0) {
+            $meta = match (true) {
+                $category->tv_meta    => TmdbTv::find($torrent->tmdb_tv_id),
+                $category->movie_meta => TmdbMovie::find($torrent->tmdb_movie_id),
+                default               => null,
+            };
+        }
+
+        return self::post([
+            'id'                     => (string) $torrent->id,
```
Roardom: Just want to double check that you intend for this to be a string instead of an int.

zze0s: If it will ever only be used with UNIT3D then int is fine but if someone have string IDs then that won't be ideal. Probably better I handle that on the other end and we send as int here :+1:

Roardom: If you want maximum flexibility for in the very very rare chance we change to uuids in the future, then keep the string :). I just wanted to make sure you weren't copying the api which does in fact use a string for this if you didn't intend to match it.

app/Bots/IRCAnnounceBotExternal.php
```diff
@@ -0,0 +1,175 @@
...
+            'category_name'          => $category->name,
+            'origin_enum'            => $originEnum,
+            'leech_type_enum'        => $leechTypeEnum,
+            'upload_time_unix_epoch' => $torrent->created_at->getTimestamp(),
+            'freeleech'              => (bool) $torrent->free > 0,
+            'freeleech_percent'      => $torrent->free,
+            'double_up'              => $torrent->doubleup,
+            'resolution'             => $torrent->resolution?->name ?? '',
+            'type'                   => $torrent->type->name,
+            'release_year'           => isset($meta->release_date) ? $meta->release_date->format('Y') : (isset($meta->first_air_date) ? $meta->first_air_date->format('Y') : null),
```
Roardom:
```suggestion
            'release_year'           => $meta?->release_date?->format('Y') ?? $meta?->first_air_date?->format('Y'),
```

Roardom: Or:

```suggestion
            'release_year'           => $meta?->release_date?->format('Y') ?? $meta?->first_air_date?->format('Y') ?? $meta?->first_release_date?->format('Y'),
```

If you want to include igdb video games as well (would also require adding a `$category->game_meta => IgdbMeta::find($torrent->igdb)` to the match above)

zze0s:
Great, thanks for reviewing! Have fixed the last couple of things (hopefully) :smile:

Roardom:
Thank you!

PR #5111, opened 2025-11-07
(Fix) Various typos

flowerey:
Found while rechecking translations. Also used "typos" tool.


Roardom:
The keys can't be edited. You can only change the value

flowerey:
Will look onto that once I have time, in the meantime you can merge this if you want and I will make a seperate PR.

flowerey:
Okay

PR #5133, opened 2025-11-25
(Update) CI

flowerey:
Updates CI. Didn't update spellcheck since I already have a PR for that.

### DIFF
--- a/.github/workflows/compile-assets-test.yml
+++ b/.github/workflows/compile-assets-test.yml
@@ -12,7 +12,7 @@ jobs:
         runs-on: ${{ matrix.operating-system }}
         steps:
             -   name: Checkout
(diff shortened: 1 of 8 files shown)


Roardom (approved):
Approved.

I went through all the updates and as far as I can tell there are no breaking changes that affect UNIT3D.

Next time, please only update dependencies if there is a meaningful improvement. Some good examples of meaningful improvements might be runner compatibility, security fixes, performance improvements or newly added features that we want to use. This helps reduce unneeded churn when the current dependencies suffice for what we need. The reasoning behind the change could then be included as the reason for updating.

flowerey:
Noted. Will do another PR for casing and naming changes if you want.

PR #5142, opened 2025-11-29
(Refactor) Events should be named Giveaways

Obi-Wana:
Rename all Event stuff to Giveaway. Also prefix their related models with `gateway_` (I don't see this being modular and re-usable for other parts of code).


Obi-Wana:
> Tests pass... should we wait until the end of the holiday season before merging?

Unless you plan a release for mid-December, IMO we can merge now. It's going to be in the development branch and not in main anyway. Majority will not update from development.

PR #5180, opened 2025-12-28
New unit tests

hier0p:


tests/Unit/Actions/Fortify/UpdateUserPasswordTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $this->actingAs($user);
+        $action = new UpdateUserPassword();
+        
+        // This should fail but doesn't because 'confirmed' rule is missing from PasswordValidationRules
+        $action->update($user, [
+            'current_password' => $oldPassword,
+            'password' => $newPassword,
+            'password_confirmation' => $differentPassword,
+        ]);
```
hier0p, line 80: This test is configured to match the existing code but this looks like a bug -- should I fix the bug and update the test or do I have a gap in my understanding.

tests/Unit/Actions/Fortify/UpdateUserProfileInformationTest.php
```diff
@@ -0,0 +1,78 @@
...
+            'email' => fake()->freeEmail,
+        ]);
+
+        $action = new UpdateUserProfileInformation();
+        
+        // This action is broken because it expects 'name' field but User model uses 'username'
+        expect(fn() => $action->update($user, [
+            'name' => fake()->name,
+            'email' => fake()->freeEmail,
+        ]))->toThrow(QueryException::class);
```
hier0p, line 35: It looks like the command is using `name` but the `User` model has `username`. Is it OK to update the command or am I misunderstanding how things work?

tests/Unit/Actions/Fortify/UpdateUserPasswordTest.php
Roardom: Sorry for the delay, I missed this question. Yes, if you can add the additional rule to PasswordValidationRules that would be great.

tests/Unit/Actions/Fortify/UpdateUserProfileInformationTest.php
Roardom: Sorry for the delay. This action isn't actually used anywhere. It could probably just be deleted, along with the reference deleted in the fortify service provider, and feature disabled in the fortify configuration.

PR #5181, opened 2025-12-29
(Add) Extended user statistics to API endpoint

SlovakianCanon:
## Summary
Extends the `/api/user` endpoint with additional statistics already available on the profile page:

- `real_uploaded` / `real_downloaded` - actual transfer amounts
- `credited_uploaded` / `credited_downloaded` - with bonus multipliers applied
- `average_seedtime` - average seed time per torrent
- `seeding_size` - total size of torrents currently seeding
- `fl_tokens` - freeleech tokens
- `uploads_count` / `downloads_count` - torrent counts
- `bonus_uploaded` - upload purchased with bonus points

## Changes
- `app/Http/Controllers/API/UserController.php` - added history aggregation query
- `app/Http/Resources/UserResource.php` - added new fields to response
- `book/src/torrent_api.md` - updated API documentation


app/Http/Controllers/API/UserController.php
```diff
@@ -17,15 +17,38 @@
...
+            ->selectRaw('SUM(actual_downloaded > 0) as download_count')
+            ->selectRaw('COUNT(*) as count')
+            ->first();
+
+        $seedingSize = $user->seedingTorrents()->sum('size');
+
+        $bonusUploaded = BonTransactions::query()
+            ->where('sender_id', '=', $user->id)
+            ->where('name', 'like', '%Upload%')
+            ->sum('cost');
```
Roardom:
```suggestion
        $bonusUploaded = BonTransactions::query()
            ->where('sender_id', '=', $user->id)
            ->where('name', 'like', '%Upload%')
            ->sum('cost');
```

Wildcard searches shouldn't be permitted in the api.

app/Http/Controllers/API/UserController.php
```diff
@@ -17,15 +17,38 @@
...
+use Illuminate\Support\Facades\DB;
 
 class UserController extends BaseController
 {
     final public function show(): UserResource
     {
         $user = auth()->user();
 
+        $history = DB::table('history')
+            ->where('user_id', '=', $user->id)
```
Roardom:
```suggestion
        $history = History::query()
            ->withTrashed()
            ->where('user_id', '=', $user->id)
```

app/Http/Resources/UserResource.php
```diff
@@ -38,21 +50,45 @@ class UserResource extends JsonResource
...
+            'seeding'            => \count($this->seedingTorrents),
+            'leeching'           => \count($this->leechingTorrents),
+            'seedbonus'          => $this->seedbonus,
+            'hit_and_runs'       => $this->hitandruns,
+            'real_uploaded'      => str_replace("\u{00A0}", ' ', StringHelper::formatBytes((int) ($this->history?->upload_sum ?? 0), 2)),
+            'real_downloaded'    => str_replace("\u{00A0}", ' ', StringHelper::formatBytes((int) ($this->history?->download_sum ?? 0), 2)),
+            'credited_uploaded'  => str_replace("\u{00A0}", ' ', StringHelper::formatBytes((int) ($this->history?->credited_upload_sum ?? 0), 2)),
+            'credited_downloaded' => str_replace("\u{00A0}", ' ', StringHelper::formatBytes((int) ($this->history?->credited_download_sum ?? 0), 2)),
+            'average_seedtime'   => StringHelper::timeElapsed($avgSeedtime),
+            'seeding_size'       => str_replace("\u{00A0}", ' ', StringHelper::formatBytes($this->seedingSize, 2)),
```
Roardom: These should be raw values, not human ones.

app/Http/Resources/UserResource.php
```diff
@@ -38,21 +50,45 @@ class UserResource extends JsonResource
...
+            'fl_tokens'          => $this->fl_tokens,
+            'uploads_count'      => $this->uploadsCount,
+            'downloads_count'    => (int) ($this->history?->download_count ?? 0),
+            'bonus_uploaded'     => str_replace("\u{00A0}", ' ', StringHelper::formatBytes($this->bonusUploaded, 2)),
```
Roardom: ditto

app/Http/Resources/UserResource.php
```diff
@@ -38,21 +50,45 @@ class UserResource extends JsonResource
...
+     *     uploads_count: int,
+     *     downloads_count: int,
+     *     bonus_uploaded: string,
      * }
      */
     public function toArray(Request $request): array
     {
+        $avgSeedtime = $this->history?->count > 0
+            ? (int) (($this->history->seedtime_sum ?? 0) / $this->history->count)
+            : 0;
```
Roardom: You should be able to use resource loading instead of manual db queries and passing them through: https://laravel.com/docs/12.x/eloquent-resources#conditional-relationships

Roardom (changes requested):
This is a good PR. I'm currently working on a more refined compartmentalized framework for api usage/keys and will consider this PR as soon as that framework is in place.

Roardom (commented):
My review at this point is still the same:

> This is a good PR. I'm currently working on a more refined compartmentalized framework for api usage/keys and will consider this PR as soon as that framework is in place.

blixten85:
> This is a good PR. I'm currently working on a more refined compartmentalized framework for api usage/keys and will consider this PR as soon as that framework is in place.

Can the FL token be used (not only viewed) in the *arrs? So that can be chosen to be used first on every download?

oOStroudyOo:
Should number of invites also be added to this?
I know it's not really a "stat" but could be useful to many

Roardom:
> Should number of invites also be added to this? I know it's not really a "stat" but could be useful to many

Let's keep the stats in this PR as is for the time being. I think invites are one step too far considering the amount of apikeys that regularly get leaked from unprotected 3rd party instances. As is, the stats already indicate to malicious users just how much an account can be abused traffic-wise.

PR #5186, opened 2025-12-31
(Fix) Update TopTorrents newest tab

crKtv:
Since this was ordered by torrent_id, if a torrent was uploaded and pending moderation and other trusted torrents were uploaded after it, those would bury the pending moderation torrent once it was approved. Ordering by created_at fixes this.

### DIFF
--- a/app/Http/Livewire/TopTorrents.php
+++ b/app/Http/Livewire/TopTorrents.php
@@ -65,7 +65,7 @@ class TopTorrents extends Component
                     END AS meta
                 SQL)
                 ->withCount(['comments'])
-                ->when($this->tab === 'newest', fn ($query) => $query->orderByDesc('id'))
+                ->when($this->tab === 'newest', fn ($query) => $query->orderByDesc('created_at'))
                 ->when($this->tab === 'seeded', fn ($query) => $query->orderByDesc('seeders'))
                 ->when(
                     $this->tab === 'dying',
--- a/app/Http/Livewire/TopTorrents.php
+++ b/app/Http/Livewire/TopTorrents.php
@@ -65,7 +65,7 @@ class TopTorrents extends Component
                     END AS meta
                 SQL)
                 ->withCount(['comments'])
-                ->when($this->tab === 'newest', fn ($query) => $query->orderByDesc('created_at'))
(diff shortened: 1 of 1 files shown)


Roardom (commented):
Does the created_at index apply to this new query? What are the query times for the previous query and for this new query for a database with a few hundred thousand torrents? Given that this query is loaded on the home page, we try to make it as efficient as possible.

crKtv:
From my rough testing and lack of knowledge on how to check it properly, I didn't see any performance hit on a 100k torrent set. If you want to look further into it, I can close this and open an issue instead.

Roardom:
By default, when you have `APP_DEBUG` set to `true` in your `.env`, a debug bar will appear on the bottom of every page allowing you to see the db queries and the length of time it took to complete the query.  You might also need to add `DEBUGBAR_OPEN_STORAGE=true` to your .env to see past http requests since the chatbox will send additional http requests after the initial page load. You can use the folder button in the top right of the debugbar to view previous http requests to find the one that loaded the TopTorrents.

Roardom:
On my local setup, previously, it took 3.24ms. Now it's takes 394ms. That's a very big performance drop, especially for the frequently loaded home page. Can you try and get it below 10ms? For perspective, the other 50 home page queries finish in 93ms combined.

Roardom:
You could try using bumped_at instead. It only uses 5.24 ms. It doesn't fully satisfy having only the newest torrents when there are recently bumped torrents, but it solves the moderation issue.

HDVinnie:
You could always try a index on created_at

Lokilicious:
> You could try using bumped_at instead. It only uses 5.24 ms. It doesn't fully satisfy having only the newest torrents when there are recently bumped torrents, but it solves the moderation issue.

Won't this confuse users if bumped torrents are shown as new ones?

HDVinnie:
> > You could try using bumped_at instead. It only uses 5.24 ms. It doesn't fully satisfy having only the newest torrents when there are recently bumped torrents, but it solves the moderation issue.
> 
> Won't this confuse users if bumped torrents are shown as new ones? 

Bumped is not practical. Just need to add a proper index for created.

HDVinnie:
Stale

PR #5204, opened 2026-01-07
(Update) Use cache()->flexible() to improve response times

HDVinnie:
- continuation of https://github.com/HDInnovations/UNIT3D/pull/4957


Roardom (changes requested):
They're all backwards. Also I think it'd make more sense to keep the existing value as the lower end and double it for the higher end.

PR #5216, opened 2026-01-09
(Update) GitHub actions for speed, readability and maintenance

HDVinnie:
| Workflow | Before | After | Time Saved | Improvement |
|----------|--------|-------|------------|-------------|
| Format Blade Files (Prettier) | 1m 6s | 24s | 42s | **~63.6% faster**  |
| PHP Lint (Pint) | 2m 12s | 50s | 1m 22s | **~62.1% faster**  |
| PHP Static Analysis (Larastan) | 2m 31s | 1m 5s | 1m 26s | **~57.0% faster**  |

Commits can be squashed upon merge.


Roardom:
Not sure if you want it done now or later, but it'd probably be very useful to look into using composite actions, so that we don't have to duplicate the same code in each one.

cspell.json
```diff
@@ -205,6 +204,13 @@
 		"phpstan-baseline.neon",
 		"pint.json",
 		"*.svg",
+		"public",
+		"storage",
+		"vendor",
+		"node_modules",
+		"book",
+		"bootstrap",
+		"laravel-echo-server.json",
```
Roardom, line 213: Doesn't cspell already exclude everything in the .gitignore by default?

HDVinnie: It didn't on my local instance. As for composites I would rather wait.

PR #5229, opened 2026-01-14
(Update) Auto-freeleech feature

### LINKED ISSUE #4732: [Request] Add FL token usage to API
FL tokens seem like a missed opportunity for something greater, people wont run into them unless they revive a torrent or enter a special event and having to manually use them when a large portion of people rely on *arrs to fetch their movies makes most people who have them ignore it. Perhaps a way to point them into a different direction could be by doing something similar like gazelle trackers where an option could be ticked or a url argument where it would use an FL token on download if you have one available, and if not then it just continues like normal.

### LINKED ISSUE #5170: [Request] Automatic freeleech token application
If you download a torrent and you do not have freeleech on the torrent, and you have a freeleech token, there should be a mode where the system will automatically consume a freeleech token once you show up as a leecher.

Ideally, you'd be able to define a minimum threshold so that tokens aren't burned on smaller torrents.

nikoresu:
This creates a setting for enabling autofreeleech tokens, added inside the Settings/Torrent page (might be good to update the added UI element to be more _compliant_):

image

Added the logic to apply token when the feature is enabled and complies with the given min tokens rule

The main objective is to apply tokens for automated downloads (Prowlarr, Sonarr, Radarr, etc.), but also works when downloading directly from the web UI

### DIFF
--- a/app/Http/Controllers/TorrentDownloadController.php
+++ b/app/Http/Controllers/TorrentDownloadController.php
@@ -96,7 +96,7 @@ class TorrentDownloadController extends Controller
             !cache()->get("freeleech_token:{$user->id}:{$torrent->id}")
         ) {
             FreeleechToken::query()->create([
-                'user_id' => $user->id,
+                'user_id'    => $user->id,
                 'torrent_id' => $torrent->id,
             ]);
 
--- a/app/Http/Controllers/TorrentDownloadController.php
+++ b/app/Http/Controllers/TorrentDownloadController.php
@@ -26,8 +26,6 @@ use App\Models\FreeleechToken;
 use App\Services\Unit3dAnnounce;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Storage;
-use Illuminate\Support\Facades\Log;
-use Throwable;
 
 class TorrentDownloadController extends Controller
 {
@@ -90,45 +88,24 @@ class TorrentDownloadController extends Controller
         $torrentDownload->save();
 
         // Auto-apply a freeleech token if the user has enabled the setting
-        try {
-            $settings = $user->getSettingsAttribute();
+        $settings = $user->settings;
 
-            $minTokens = $settings?->auto_freeleech_min_tokens ?? 0;
+        if (
+            $settings?->auto_freeleech_apply &&
+            $user->fl_tokens >= max(1, $settings->auto_freeleech_min_tokens) &&
+            !cache()->get("freeleech_token:{$user->id}:{$torrent->id}")
+        ) {
+            $freeleech_token = new FreeleechToken();
+            $freeleech_token->user_id = $user->id;
+            $freeleech_token->torrent_id = $torrent->id;
+            $freeleech_token->save();
 
-            if (($settings?->auto_freeleech_apply ?? false) && $user->fl_tokens >= max(1, (int) $minTokens)) {
-                $activeToken = cache()->get('freeleech_token:'.$user->id.':'.$torrent->id);
+            Unit3dAnnounce::addFreeleechToken($user->id, $torrent->id);
 
-                if ($user->fl_tokens < 1) {
-                    Log::debug('Skipping auto-freeleech - user has no tokens', ['user_id' => $user->id, 'torrent_id' => $torrent->id]);
-                } elseif ($activeToken) {
-                    Log::debug('Skipping auto-freeleech - active token exists', ['user_id' => $user->id, 'torrent_id' => $torrent->id]);
-                } else {
-                    Log::debug('Applying auto-freeleech token', ['user_id' => $user->id, 'torrent_id' => $torrent->id]);
+            $user->decrement('fl_tokens');
+            cache()->put("freeleech_token:{$user->id}:{$torrent->id}", true);
 
-                    $freeleechToken = new FreeleechToken();
-                    $freeleechToken->user_id = $user->id;
-                    $freeleechToken->torrent_id = $torrent->id;
-                    $freeleechToken->save();
-
-                    Unit3dAnnounce::addFreeleechToken($user->id, $torrent->id);
-
-                    $user->fl_tokens -= 1;
-                    $user->save();
-
-                    cache()->put('freeleech_token:'.$user->id.':'.$torrent->id, true);
-
-                    // Ensure search index reflects the change
-                    $torrent->searchable();
-
-                    Log::info('Auto-freeleech token assigned', ['user_id' => $user->id, 'torrent_id' => $torrent->id, 'remaining_tokens' => $user->fl_tokens]);
-                }
-            } else {
-                Log::debug('Auto-freeleech conditions not met', ['user_id' => $user->id, 'torrent_id' => $torrent->id]);
-            }
-        } catch (Throwable $e) {
-            // Do not block download if auto-apply fails. Log and continue
-            report($e);
-            Log::debug('Auto-freeleech assignment failed', ['user_id' => $user->id, 'torrent_id' => $torrent->id, 'error' => $e->getMessage()]);
+            $torrent->searchable();
         }
 
         return response()->streamDownload(
--- a/app/Http/Controllers/TorrentDownloadController.php
+++ b/app/Http/Controllers/TorrentDownloadController.php
@@ -27,6 +27,7 @@ use App\Services\Unit3dAnnounce;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Facades\Log;
+use Throwable;
 
 class TorrentDownloadController extends Controller
 {
@@ -124,7 +125,7 @@ class TorrentDownloadController extends Controller
             } else {
                 Log::debug('Auto-freeleech conditions not met', ['user_id' => $user->id, 'torrent_id' => $torrent->id]);
             }
-        } catch (\Throwable $e) {
+        } catch (Throwable $e) {
             // Do not block download if auto-apply fails. Log and continue
             report($e);
             Log::debug('Auto-freeleech assignment failed', ['user_id' => $user->id, 'torrent_id' => $torrent->id, 'error' => $e->getMessage()]);
--- a/app/Http/Controllers/TorrentDownloadController.php
+++ b/app/Http/Controllers/TorrentDownloadController.php
@@ -22,8 +22,11 @@ use App\Models\Scopes\ApprovedScope;
 use App\Models\Torrent;
 use App\Models\TorrentDownload;
 use App\Models\User;
+use App\Models\FreeleechToken;
+use App\Services\Unit3dAnnounce;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Storage;
+use Illuminate\Support\Facades\Log;
 
 class TorrentDownloadController extends Controller
 {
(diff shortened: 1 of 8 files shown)


resources/views/user/general-setting/edit.blade.php
```diff
@@ -550,6 +550,38 @@ class="form__checkbox"
...
+                                id="auto_freeleech_min_tokens"
+                                class="form__text"
+                                name="auto_freeleech_min_tokens"
+                                placeholder=" "
+                                type="number"
+                                min="0"
+                                step="1"
+                                value="{{ $user->settings->auto_freeleech_min_tokens }}"
+                            />
+                            <small class="text-muted">
```
HDVinnie: This text seem redundant. I would just remove it.

database/migrations/2026_01_14_120000_add_auto_freeleech_to_user_settings.php
```diff
@@ -0,0 +1,25 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('user_settings', function (Blueprint $table): void {
```
HDVinnie, line 15: You can remove both if conditions.

app/Http/Controllers/TorrentDownloadController.php
```diff
@@ -85,6 +89,48 @@ public function store(Request $request, int $id, ?string $rsskey = null): \Illum
         $torrentDownload->type = $rsskey ? 'RSS/API using '.$request->header('User-Agent') : 'Site using '.$request->header('User-Agent');
         $torrentDownload->save();
 
+        // Auto-apply a freeleech token if the user has enabled the setting
```
HDVinnie, line 90: This code seems over complicated. Why all the logging? Untest and done from phone but something like this seems more appropriate.

```php
$settings = $user->getSettingsAttribute();
$minTokens = (int) ($settings?->auto_freeleech_min_tokens ?? 0);

if (
    ($settings?->auto_freeleech_apply ?? false) &&
    $user->fl_tokens >= max(1, $minTokens) &&
    !cache()->get("freeleech_token:{$user->id}:{$torrent->id}")
) {
    FreeleechToken::create([
        'user_id' => $user->id,
        'torrent_id' => $torrent->id,
    ]);

    Unit3dAnnounce::addFreeleechToken($user->id, $torrent->id);

    $user->decrement('fl_tokens');
    cache()->put("freeleech_token:{$user->id}:{$torrent->id}", true);

    $torrent->searchable();
}
```

HDVinnie (changes requested):
@Roardom can review as well as im on mobile and might not be best.

app/Http/Controllers/TorrentDownloadController.php
Roardom: Probably something more like this since settings has default values:

```php
$settings = $user->settings;

if (
    $settings->auto_freeleech_apply &&
    $user->fl_tokens >= max(1, $settings->auto_freeleech_min_tokens) &&
    !cache()->get("freeleech_token:{$user->id}:{$torrent->id}")
) {
    FreeleechToken::create([
        'user_id' => $user->id,
        'torrent_id' => $torrent->id,
    ]);

    Unit3dAnnounce::addFreeleechToken($user->id, $torrent->id);

    $user->decrement('fl_tokens');
    cache()->put("freeleech_token:{$user->id}:{$torrent->id}", true);

    $torrent->searchable();
}
```

resources/views/user/general-setting/edit.blade.php
```diff
@@ -550,6 +550,38 @@ class="form__checkbox"
...
+                        <p class="form__group">
+                            <label class="form__label" for="auto_freeleech_min_tokens">
+                                Minimum tokens to keep when auto-applying freeleech
+                            </label>
+                            <input
+                                id="auto_freeleech_min_tokens"
+                                class="form__text"
+                                name="auto_freeleech_min_tokens"
+                                placeholder=" "
+                                type="number"
```
Roardom: Don't use `type="number"` here but `type="text"` with the numeric text input and regex validation. You should be able to find examples of this pattern used throughout the codebase.

Roardom (changes requested):
Left some additional feedback

nikoresu:
@Roardom @HDVinnie 
Comments should be resolved. Thanks for the quick reviews, it's the first time i write PHP :p

resources/views/user/general-setting/edit.blade.php
```diff
@@ -550,6 +550,35 @@ class="form__checkbox"
...
+                                    type="checkbox"
+                                    name="auto_freeleech_apply"
+                                    value="1"
+                                    @checked($user->settings->auto_freeleech_apply)
+                                />
+                                Automatically apply freeleech tokens when applicable
+                            </label>
+                        </p>
+                        <p class="form__group">
+                            <label class="form__label" for="auto_freeleech_min_tokens">
```
HDVinnie: This <label> should be using `form__label form__label--floating` class and should be placed after the input.

Example:

```
                    <p class="form__group">
                        <input
                            id="standalone_css"
                            class="form__text"
                            name="standalone_css"
                            placeholder=" "
                            type="url"
                            value="{{ $user->settings->standalone_css }}"
                        />
                        <label class="form__label form__label--floating" for="standalone_css">
                            Standalone CSS stylesheet (no site theme used)
                        </label>
                    </p>
```

HDVinnie (approved):

nikoresu:
This could be related to these issues [#5170](https://github.com/HDInnovations/UNIT3D/issues/5170) and [#4732](https://github.com/HDInnovations/UNIT3D/issues/4732)

Might steal the idea for a threshold for applying the FL token depending on the size of a torrent

resources/views/user/general-setting/edit.blade.php
HDVinnie: Just this last issue @nikoresu

nikoresu:
Moved the auto-freeleech section down so it takes the floating label style properly:
image

app/Http/Controllers/TorrentDownloadController.php
```diff
@@ -85,6 +87,27 @@ public function store(Request $request, int $id, ?string $rsskey = null): \Illum
...
+
+        if (
+            $settings?->auto_freeleech_apply &&
+            $user->fl_tokens >= max(1, $settings->auto_freeleech_min_tokens) &&
+            !cache()->get("freeleech_token:{$user->id}:{$torrent->id}")
+        ) {
+            $freeleech_token = new FreeleechToken();
+            $freeleech_token->user_id = $user->id;
+            $freeleech_token->torrent_id = $torrent->id;
+            $freeleech_token->save();
```
Roardom:
```suggestion
            FreeleechToken::query()->create([
                'user_id'    => $user->id,
                'torrent_id' => $torrent->id,
            ]);
```

resources/views/user/general-setting/edit.blade.php
```diff
@@ -564,7 +564,37 @@ class="form__checkbox"
...
+                        <input
+                            pattern="[0-9]*"
+                            id="auto_freeleech_min_tokens"
+                            class="form__text"
+                            name="auto_freeleech_min_tokens"
+                            placeholder=" "
+                            type="text"
+                            min="0"
+                            step="1"
+                            value="{{ $user->settings->auto_freeleech_min_tokens }}"
```
Roardom:
```suggestion
                            id="auto_freeleech_min_tokens"
                            class="form__text"
                            inputmode="numeric"
                            name="auto_freeleech_min_tokens"
                            pattern="[0-9]*"
                            placeholder=" "
                            type="text"
                            required
                            value="{{ $user->settings->auto_freeleech_min_tokens }}"
```

Roardom (changes requested):
Can you also run `./node_modules/.bin/prettier resources --write` to format the blade?

As well as `php artisan schema:dump` and committing the new schema

app/Models/FreeleechToken.php
```diff
@@ -39,6 +39,16 @@ final class FreeleechToken extends Model
...
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var list<string>
+     */
+    protected $fillable = [
+        "user_id",
+        "torrent_id"
+    ];
+
```
Roardom, line 48: Oh whoops, I didn't realize that the guards weren't set up here yet. It should be done this way instead:

```suggestion
    /**
     * The attributes that aren't mass assignable.
     *
     * @var list<string>
     */
    protected $guarded = [];

```

database/schema/mysql-schema.sql
```diff
@@ -2531,16 +2531,18 @@ CREATE TABLE `user_settings` (
...
   `torrent_layout` tinyint unsigned NOT NULL DEFAULT '0',
   `torrent_filters` tinyint(1) NOT NULL DEFAULT '0',
-  `custom_css` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
-  `standalone_css` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `custom_css` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `standalone_css` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT NULL,
   `show_poster` tinyint(1) NOT NULL DEFAULT '0',
   `unbookmark_torrents_on_completion` tinyint(1) NOT NULL,
+  `auto_freeleech_apply` tinyint(1) NOT NULL DEFAULT '0',
+  `auto_freeleech_min_tokens` int unsigned NOT NULL DEFAULT '0',
```
Roardom, line 2543: These two lines here (and the migration one at the very bottom) are the only lines we need. I'm not sure why the character set is being explicitly added when you run the command. Can you ensure only those 3 lines are committed?

Roardom (changes requested):
Thanks! A few more touchups

Roardom:
Oh and can you run `./vendor/bin/pint` to format the PHP?

database/schema/mysql-schema.sql
Roardom: Still need the migration line at the very bottom to fix the ci

nikoresu:
I hope this is the good one! 😅

database/schema/mysql-schema.sql
```diff
@@ -3103,3 +3105,4 @@ INSERT INTO `migrations` (`id`, `migration`, `batch`) VALUES (369,'2025_11_29_10
 INSERT INTO `migrations` (`id`, `migration`, `batch`) VALUES (370,'2026_01_06_231535_remove_unnecessary_bigints',1);
 INSERT INTO `migrations` (`id`, `migration`, `batch`) VALUES (371,'2026_01_07_040502_mark_columns_as_unsigned',1);
 INSERT INTO `migrations` (`id`, `migration`, `batch`) VALUES (372,'2026_01_09_015532_alter_table_reports_make_verdict_nullable',1);
+INSERT INTO `migrations` (`id`, `migration`, `batch`) VALUES (373,'2026_01_14_120000_add_auto_freeleech_to_user_settings',2);
```
Roardom:
```suggestion
INSERT INTO `migrations` (`id`, `migration`, `batch`) VALUES (373,'2026_01_14_120000_add_auto_freeleech_to_user_settings',1);
```

I think this is the missing change

Roardom:
Thank you!

nikoresu:
Thanks for the help! and sorry for all the back-and-forths

shashwat001:
This should also be more dynamic that freeleech tokens should only be applied when there is no existing freeleech condition like global freeleech or personal freeleech or torrent specific freeleech otherwise freeleech token usage would be a waste.

nikoresu:
> This should also be more dynamic that freeleech tokens should only be applied when there is no existing freeleech condition like global freeleech or personal freeleech or torrent specific freeleech otherwise freeleech token usage would be a waste.

right, i was also thinking that maybe tokens will be consumed when downloading the file generated after an upload cause I believe it ends up using the same controller. i'll need to expand this

PR #5230, opened 2026-01-17
(Fix) Hide news block when no articles exist

thatonePRguy:
The news block currently renders an empty section when there are no existing articles.

Adding `@if ($articles->isNotEmpty())` follows the same pattern used by featured.blade.php to prevent rendering an empty article section when there are no articles to display.

Note: Diff looks messy since everything was indented one level. Only additions are L1: `@if ($articles->isNotEmpty())` and L70: `@endif`

### DIFF
--- a/resources/views/blocks/news.blade.php
+++ b/resources/views/blocks/news.blade.php
@@ -19,7 +19,10 @@
             </h2>
             <div class="panel__actions">
                 <div class="panel__action">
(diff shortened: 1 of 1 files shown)


HDVinnie:
You need to run prettier per contribution guidelines. This is why your diff is a mess. I assume your IDE tried to auto format when that is not the style we follow

HDVinnie:
Thank you for the contribution

PR #5235, opened 2026-01-22
(Fix) Long username overflow in online users block

sudo2k16:
## Description
Fixes long usernames causing horizontal overflow in the online users block.

## Changes
- Added `max-width: 100%`, `overflow: hidden`, and `text-overflow: ellipsis` to `.user-tag` and `.user-tag__link` in SCSS
- Added `flex-shrink: 0` to icons to prevent them from shrinking

## Before
Long usernames would overflow and break the column layout.

## After
Long usernames are truncated with ellipsis (`...`) while preserving the full username in the title attribute on hover.


Roardom (changes requested):
I don't agree with hiding portions of usernames. The online users css should be changed instead to accommodate. For context, the max length of a username is validated to 25 characters upon user creation.

Roardom:
> while preserving the full username

Is this AI speaking? Because you definitely didn't do this.

sudo2k16:
> > while preserving the full username
> 
> Is this AI speaking? Because you definitely didn't do this.

The AI is a human dev and it’s speaking. Having a user online with a 25-character-long name will result in massive overflowing. This happens really rarely as almost no one will choose a name that long, but if someone does and there are a lot of users online, it results in issues.

Lokilicious:
> I don't agree with hiding portions of usernames. The online users css should be changed instead to accommodate. For context, the max length of a username is validated to 25 characters upon user creation.

> > while preserving the full username
> 
> 
> 
> Is this AI speaking? Because you definitely didn't do this.



How about you try having a username with 25 characters and see how it overflows into other usernames before being snarky at contributors who are just trying to make your product better ...

AnabolicsAnonymous:
> How about you try having a username with 25 characters and see how it overflows into other usernames before being snarky at contributors who are just trying to make your product better ...

Literally gave an alternative suggestion in a calm matter btw.

HDVinnie:
> > I don't agree with hiding portions of usernames. The online users css should be changed instead to accommodate. For context, the max length of a username is validated to 25 characters upon user creation.
> 
> > > while preserving the full username
> > 
> > 
> > 
> > Is this AI speaking? Because you definitely didn't do this.
> 
> 
> 
> How about you try having a username with 25 characters and see how it overflows into other usernames before being snarky at contributors who are just trying to make your product better ... 

It doesn't hurt to ask. You know how many people don't know what there doing and just rely on AI to do everything for them.

Roardom:
I'm not being snarky... I explained the solution I wanted to see.

> The online users css should be changed instead to accommodate

In the current state, it hides all usernames, everywhere, globally. This does a lot of damage for everywhere *except* the online users block.

My other comment was specifically about how the PR text mentioned that the username is preserved in the title tag, but the PR code says otherwise. This is another problem needing solved still.

HDVinnie:
> > > while preserving the full username
> 
> > 
> 
> > Is this AI speaking? Because you definitely didn't do this.
> 
> 
> 
> The AI is a human dev and it’s speaking. Having a user online with a 25-character-long name will result in massive overflowing. This happens really rarely as almost no one will choose a name that long, but if someone does and there are a lot of users online, it results in issues. 

I agree this is a ux bug and does need to be fixed.

sudo2k16:
> I'm not being snarky... I explained the solution I wanted to see.
> 
> > The online users css should be changed instead to accommodate
> 
> In the current state, it hides all usernames, everywhere, globally. This does a lot of damage for everywhere _except_ the online users block.

In the current state it would not show the full name everywhere where the name would overflow and I'm pretty you overflowing should never happen 

> My other comment was specifically about how the PR text mentioned that the username is preserved in the title tag, but the PR code says otherwise. This is another problem needing solved still.

the default behavior of `text-overflow: ellipsis` is to show the full text as title tag.

Roardom:
image

This is what your code does on hover (latest firefox)

Lokilicious:
> image
> This is what your code does on hover (latest firefox)

Speaking as an outsider (who is affected by this bug and was happy someone tackled it):

Obviously the code did not change the hover text, but IMHO the change is otherwise good and is only affecting places where there is not enough place for the username (IMHO that’s good behaviour and I would have done it the same way to be honest) 

so instead of telling a new contributor he is AI because of one (!!!) word in an otherwise good description another more constructive way to communicate this would make more sense.

PR #5238, opened 2026-01-23
(Add) WebAuthn/FIDO2 Security Key Support for 2FA

sudo2k16:
This PR adds hardware security key support (FIDO2/WebAuthn) as an additional two-factor authentication method.

### Changes

- Adds `asbiin/laravel-webauthn` package for WebAuthn functionality
- Users can register multiple security keys in their 2FA settings
- Security keys are prioritized over TOTP codes during login
- Falls back to TOTP code if user doesn't have a security key or cancels

### New Files

- `RedirectIfTwoFactorAuthenticatable` - Extends Fortify's 2FA detection to check for security keys
- `WebAuthnTwoFactorController` - Handles WebAuthn authentication during login
- `SecurityKeyForm` - Livewire component for managing security keys
- Migration for `webauthn_keys` table

### How It Works

1. User registers a security key in Settings -> Two-Factor Authentication
2. On next login, if the user has a security key registered, they're prompted to use it
3. If TOTP is also enabled, user can switch to code entry as fallback

### Testing

- Tested with YubiKey 5
- Works in Chrome, Firefox, Safari
- Requires HTTPS


Roardom (changes requested):
Wow. Thanks for the PR. This is quite big (and might take some time to review). I've done a quick glance over so far, and first thing I notice is that 2500 of the 4000 lines of code changed are in the lock file. I think you might have accidentally updated all the dependencies instead of just adding the new ones. Can you avoid updating unnecessary packages to make it easier to review? While I'd like to trust you, any added package could be malicious, and I'd rather not go through the extra 2500 lines.

And also on that note, are you using in prod and trust the code or this is a proof of concept? Did you write this yourself following docs or did you use an LLM? Just trying to figure out how fine of a comb I need to go over this considering how critical it is to ensure this code is secure.

sudo2k16:
> Wow. Thanks for the PR. This is quite big (and might take some time to review). I've done a quick glance over so far, and first thing I notice is that 2500 of the 4000 lines of code changed are in the lock file. I think you might have accidentally updated all the dependencies instead of just adding the new ones. Can you avoid updating unnecessary packages to make it easier to review? While I'd like to trust you, any added package could be malicious, and I'd rather not go through the extra 2500 lines.
> 
> And also on that note, are you using in prod and trust the code or this is a proof of concept? Did you write this yourself following docs or did you use an LLM? Just trying to figure out how fine of a comb I need to go over this considering how critical it is to ensure this code is secure.

Good catch on the lock file - that was unintentional, I'll clean that up.

To answer your question: I wrote this myself following Laravel docs, though I did use AI as a coding assistant to help speed things up and catch edge cases. I've reviewed and tested the code myself. Happy to address any specific concerns you have during review.

sudo2k16:
Just pushed a cleaned up version. The lock file is still ~1,500 lines but that's the webauthn package's dependencies - it pulls in 15 new packages (web-auth/webauthn-lib, spomky-labs crypto libs, some symfony components, etc.). Only 3 existing packages changed: league/uri got a minor bump (7.7 -> 7.8) and symfony/clock downgraded for compatibility. Can't really make it smaller since those are all needed for the FIDO2 spec implementation.

PR #5243, opened 2026-01-27
(Update) Prevent showing cspell progress in ci

Roardom:


Roardom:
Very nice. It caught a recent error.

PR #5249, opened 2026-01-29
(Add) Stale action

flowerey:
Adds an action that closes inactive issues.

Configured with the default options for now. LMK if you want a change.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,26 @@
+name: Close inactive issues
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+        actions: write
+        contents: write
+        issues: write
+        pull-requests: write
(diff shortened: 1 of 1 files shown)


HDVinnie:
I'm not sure this is the best idea because 95% of the issues are requests. A project like this is never ending. These requests can be slowly chipped away as they are, but it really comes down to time and energy.

flowerey:
Maybe there is a way to ignore issues starting with (Request)? I will look further onto it

flowerey:
Fixed it so it won't close Requests.

Roardom:
I don't think automatically closing issues is the best idea at this point in time. It's usually only needed to close stale issues if there are so many issues that you can't keep track of them. But until there are over 200+ issues that likely isn't needed. Right now there are only 9.

flowerey:
Hmm, I see. Do I leave this open or close?

HDVinnie:
I think we can close this for now. It doesn't indeed seem unnecessary at this point when 95% of our issues are requests.

PR #5254, opened 2026-02-01
(Update) Improve highspeed tag query

Roardom:
More direct and a bit faster as well.


Roardom:
Doesn't work, the IPs are encrypted.

PR #5263, opened 2026-02-04
(Fix) Handle null release_date with collection movies

thatonePRguy:
The movie-meta.blade throws a "Call to a member function format() on null" error when attempting to view any item that is a part of a collection when one item in that collection lacks a release date.

Changed $movie->release_date->format('Y') to substr($movie->release_date ?? '', 0, 4) to match the same date handling format used elsewhere in the file.

NOTE: This will leave the empty parentheses when release_date is null. This is a pretty obscure edge case so the stability of matching existing patterns is probably preferable. Alternatively, the line could be wrapped in a ternary to also hide parentheses.

### DIFF
--- a/resources/views/torrent/partials/movie-meta.blade.php
+++ b/resources/views/torrent/partials/movie-meta.blade.php
@@ -366,13 +366,13 @@
                                             <h3 class="meta-chip__value">
                                                 <strong>
                                                     {{ $movie->title }}
-                                                    ({{ substr($movie->release_date ?? '', 0, 4) }})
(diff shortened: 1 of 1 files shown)


Roardom:
I prefer keeping the parentheses empty as well.

But, can you do `?->format('Y')` instead?

thatonePRguy:
> I prefer keeping the parentheses empty as well.
> 
> But, can you do `?->format('Y')` instead?

Sure thing. It's been updated.

Roardom:
Thank you!

PR #5265, opened 2026-02-05
(Refactor) Use typed middleware

Roardom:
Also removed the middleware aliases for the typed alternatives.

Laravel 12.45 adds a static constructor to RedirectIfAuthenticated which is should be used once we update.

Also, should middleware be imported or use the full path?


HDVinnie:
I think import is the way to go. It looks like in most files you're already doing that besides one.

HDVinnie (approved):

PR #5267, opened 2026-02-11
(Update) Laravel Debugbar to v4.0 (with dependencies)

Oha-you:
It should be safe to update this package for local development. Release notes: 
- https://github.com/fruitcake/laravel-debugbar/releases/tag/v4.0.0
- https://fruitcake.nl/blog/laravel-debugbar-v4-release

Remove the old version first
```console
composer remove barryvdh/laravel-debugbar --dev --no-scripts
```

Then do `composer install` and (re)generate `config/debugbar.php`
```console
php artisan vendor:publish --provider='Fruitcake\LaravelDebugbar\ServiceProvider'
```
Make sure environmental variables are set in `.env`
```dotenv
APP_ENV=local
APP_DEBUG=true
DEBUGBAR_ENABLED=true
DEBUGBAR_OPEN_STORAGE=true
CSP_ENABLED=false
```
CSP needs to be disabled for the current `config/secure-headers.php`, otherwise the required scripts won't load.
Don't use these settings in production environment, obviously.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,11 +43,11 @@
         "vstelmakh/url-highlight": "^3.1.3"
     },
     "require-dev": {
-        "barryvdh/laravel-debugbar": "^3.16.3",
         "brianium/paratest": "^7.16.0",
         "calebdw/larastan": "^3.8.7",
         "calebdw/larastan-livewire": "^2.3.0",
         "fakerphp/faker": "^1.24.1",
+        "fruitcake/laravel-debugbar": "^4.0",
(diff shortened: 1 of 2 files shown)


HDVinnie (approved):

HDVinnie:
@Roardom any objections?

Roardom:
> @Roardom any objections?

I wanted to double check that the dependencies added in the composer.lock file are the same ones as would be added from the composer.json, unless you've done so already.

Roardom (approved):

Roardom:
Thank you!

PR #1760, opened 2021-05-12
(Fix) Upload page: ajax preview description, save values, don't replace

Oha-you:
Some improvements for the upload page:
1. Preview description without page reload, fixes #1741
2. Correctly restore `old()` values after page reload (e.g. in case of errors). Current checks with `??` (`isset`) always return `true` and erase values because defaults are set in [TorrentController.php#L1208](https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/master/app/Http/Controllers/TorrentController.php#L1208)
3. Don't trigger `uploadExtension.hook()` if any of these inputs already have values: `name`, `tmdb`, `imdb`.
It might be frustrating to see an error after page reload, select torrent file again and have name/ID values replaced. Automatic values should be set only if these inputs are empty.

<details>
<summary>Changes in helper.js</summary>

</details>

PR #5269, opened 2026-02-11
(Remove) Redundant TMDB ID value check that's always true

Oha-you:
To reproduce:
- Open Torrent Upload form
- Set Title and some IDs: TMDB movie/TV ID, IMDB ID, MAL ID or TVDB ID
- Pick a .torrent file
- Observe that all IDs are reset

Note: this doesn't happen when you change the category first.
But it does happen if you set it on page load with `torrents/create?category_id=2`

I'm not sure why `cat = $refs.catId.value` was needed in `@change` originally for #2243
Because `cat` is assigned here anyway:
https://github.com/HDInnovations/UNIT3D/blob/735b4b7031a8374f5873f6f4ef4c237ca24d166e/resources/views/torrent/create.blade.php#L44-L48

Values are reset because they listen to `cat` changes:
https://github.com/HDInnovations/UNIT3D/blob/735b4b7031a8374f5873f6f4ef4c237ca24d166e/resources/views/torrent/create.blade.php#L316

When you change the category from the dropdown list, IDs should reset - this is correct.
Also IDs are retrieved through TMDB API when the Title is not set and .torrent is selected.
But after you fill all the inputs in the upload form yourself, they shouldn't be touched.

### DIFF
--- a/resources/js/unit3d/helper.js
+++ b/resources/js/unit3d/helper.js
@@ -184,11 +184,8 @@ class uploadExtensionBuilder {
     }
     hook() {
         let name = document.querySelector('#title');
-        let tmdb_movie = document.querySelector('#auto_tmdb_movie');
-        let tmdb_tv = document.querySelector('#auto_tmdb_tv');
-        let imdb = document.querySelector('#autoimdb');
 
-        if (!name.value.trim() && (!tmdb_movie.value.trim() || !tmdb_tv.value.trim())) {
+        if (!name.value.trim()) {
             let torrent = document.querySelector('#torrent');
             let release;
             if (!name.value) {


Oha-you:
Adding another commit here.
Condition in parenthesis always returns `true` because one of `#auto_tmdb_movie` or `#auto_tmdb_tv` values is always NULL depending on whether Movie or TV category is selected:
https://github.com/HDInnovations/UNIT3D/blob/735b4b7031a8374f5873f6f4ef4c237ca24d166e/resources/js/unit3d/helper.js#L191

Introduced after the split happened: #4563
Originally this logic was added by me back in 2021: #1760

But I think just checking if the title is set should be enough. I doubt people set DB IDs first when they upload a torrent anyway.

resources/views/torrent/create.blade.php
```diff
@@ -79,7 +79,7 @@ class="upload-form-file form__file"
                         name="torrent"
                         id="torrent"
                         required
-                        @change="uploadExtension.hook(); cat = $refs.catId.value"
+                        @change="uploadExtension.hook()"
```
Roardom: I believe the reason this code is needed is because the uploadExtension.hook() changes #autocat's value, and alpine.js's x-model down below doesn't register this change and update its value, so it has to be done manually.

Oha-you: You were right. But what if we trigger the `change` event in JS instead of the template?
Please see the updated commit.
Tested it with Movies and TV category selected by default, then added a TV/Movie torrent.
`@change` event was triggered and DB ID inputs got updated.
This discussion helped: https://github.com/alpinejs/alpine/discussions/1388

Roardom: I think I'd rather keep the explicitness for now. It's impossible to link the event emitters to the event listeners in the code, and at least the way it was before, it was all referenced on the same line. Maybe at some point it can be all moved into alpine. For now, can you revert the changes you made to the category changing?

Oha-you: Alright, first commit removed. Also changed the PR title.

Roardom:
Thank you!

PR #5275, opened 2026-02-17
(Fix) Make the hidden captcha non-tabbable

Oha-you:
Right now it's possible to focus the hidden captcha field by selecting the first input on the login page and pressing Tab three times.
You can even input something there to trigger an error on submit.
I don't think it should be tabbable, otherwise it's not truly hidden.
It'll still work as a honeypot for some dumb bots. And the more clever ones probably check for `translate` anyway.


Roardom (changes requested):
Can you format the blade files please (`./node_modules/.bin/prettier . --write`)

Oha-you:
> Can you format the blade files please (`./node_modules/.bin/prettier . --write`)

Sure! Done.

Roardom:
Thank you!

PR #5277, opened 2026-02-17
(Update) Replace and remove Thanks in password reset and emails

Oha-you:
1. It seems strange to thank a user who forgot a password and used the password reset form.
Make it consistent with other Success messages:
https://github.com/HDInnovations/UNIT3D/blob/6af8f5c94f5185f8986abe0f1f068a875fc4ef99/resources/views/auth/forgot-password.blade.php#L48-L52

2. Replace "Thanks" with "Regards" in Deny Application and Test emails.
Same idea, plus make it consistent with `email.blade.php`:
https://github.com/HDInnovations/UNIT3D/blob/6160f12d818ac18b3765cfc0e3f601c1c0f7d22e/resources/views/vendor/notifications/email.blade.php#L43-L48
3. Add `<br>` tags since this text goes into a single `<p>` tag.

### DIFF
--- a/app/Notifications/MassEmail.php
+++ b/app/Notifications/MassEmail.php
@@ -70,8 +70,7 @@ class MassEmail extends Notification implements ShouldQueue
         return (new MailMessage())
             ->subject($this->subject)
             ->line($this->message)
-            ->action('Login Now', route('login'))
-            ->line('Thank you for using 🚀'.config('other.title'));
+            ->action('Login Now', route('login'));
     }
 
     /**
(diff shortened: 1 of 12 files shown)


flowerey (approved):

flowerey:
Seems good to me

HDVinnie (approved):

Oha-you:
Question: what do you think about removing all these lines from the emails as well?

no-thank-you

---

thank-emails

---
I think just "Regards" is fine. There's no need to thank anyone for getting a (pre)warning or being banned.
Of course, each tracker can adjust this text as needed. But such defaults are better, in my opinion.

If you agree with it, I can add a new commit to this PR.

HDVinnie:
Makes sense to me. Will review and merge once updated.

HDVinnie (approved):

PR #4157, opened 2024-09-12
(Add) Donation System

HDVinnie:

PR #5282, opened 2026-02-18
(Fix) Donations: prevent time loss, fix sorting, count pending, support multiple same user donos

Oha-you:
1. Using `date` type for donation's `starts_at` and `ends_at` times leads to up to 24 hours of lost paid time for all donors. Example:
    - Admin notices a "7 days VIP" donation at 23:59 UTC on Monday, February 16th and instantly approves it
    - `starts_at` value is set to `2026-02-16` without a time (`00:00:00` internally); server time is also UTC
    - `ends_at` value is set to `2026-02-23` without a time (`00:00:00` internally)
    - On February 23rd a scheduled script `auto:remove_expired_donors` starts at `00:00:01` and sets `is_donor` for this user to `false`: https://github.com/HDInnovations/UNIT3D/blob/fd63b96269b7c17396e968ab65450d2e4222c3f1/app/Console/Commands/AutoRemoveExpiredDonors.php#L54-L56
    - Result: user only got 6 whole days of the VIP status, from Tuesday to Sunday
    - Note: `timestamp` type is chosen instead of `datetime` because it automatically adjusts values to the server timezone in case the admin decides to change it (not a single second will be lost)
    - Note: After this change users will get extra donor time depending on when the admin approves a donation and when the scheduled script starts. If it's undesirable, replace `->daily()` in `Console/Kernel.php` with `twiceDaily()`, `everySixHours()`, etc.
    - donate-lost

2. Donations should be sorted by `updated_at` property in sum and goal calculations. Example:
    - User sends a donation on February 28th (`created_at` = `2026-02-28`)
    - Admin notices it on March 1st and approves it (`updated_at` = `2026-03-01`)
    - It is counted towards the previous month in the charts and not counted towards the updated goal at all
    - Also change it for the profile view because "Latest donation" should mean "Latest approved donation" in case a user sends multiple ones and the admin approves them in a random order *(this is also needed for the next fix)*

3. Correctly set start and end times for donations made by the same user when the current donation is still active
    - Also display `updated_at` instead of `starts_at` for "Latest donation date" because after this change `starts_at` might be in the future
    - If there's only one active donation, these values will be the same
    - donate-multi

4. Remove some hardcoded `USD` and `$` text from the donation modal
    - Also add currency inside the Cost label
    - Fix some grammar mistakes; "etc." should always have a dot
    - Use the same letter case in two places

5. Add "Pending donations count" to the Staff Dashboard, same as for Applications (with an animated badge)

### DIFF
--- a/app/Http/Controllers/Staff/DonationController.php
+++ b/app/Http/Controllers/Staff/DonationController.php
@@ -73,18 +73,30 @@ class DonationController extends Controller
 
         $donation = Donation::query()->with(['user', 'package'])->findOrFail($id);
         $donation->status = ModerationStatus::APPROVED;
+
+        $latestDonation = Donation::query()
+            ->where('status', '=', ModerationStatus::APPROVED)
+            ->where('user_id', '=', $donation->user->id)
+            ->latest('updated_at')
+            ->first();
+
+        $isLifetime = $donation->package->donor_value === null ||
+            ($latestDonation && $latestDonation->ends_at === null);
+
         $donation->starts_at = $now;
 
-        if ($donation->package->donor_value > 0) {
-            $donation->ends_at = $now->addDays($donation->package->donor_value);
-        } else {
-            $donation->ends_at = null;
+        if ($latestDonation && !$isLifetime) {
+            $donation->starts_at = $latestDonation->ends_at;
+        }
+
+        if (!$isLifetime) {
+            $donation->ends_at = $donation->starts_at->copy()->addDays($donation->package->donor_value);
         }
 
         $donation->user->invites += $donation->package->invite_value ?? 0;
         $donation->user->uploaded += $donation->package->upload_value ?? 0;
         $donation->user->is_donor = true;
-        $donation->user->is_lifetime = $donation->package->donor_value === null;
+        $donation->user->is_lifetime = $isLifetime;
         $donation->user->seedbonus += $donation->package->bonus_value ?? 0;
         $donation->user->save();
 
--- a/app/Http/Controllers/Staff/DonationController.php
+++ b/app/Http/Controllers/Staff/DonationController.php
@@ -39,7 +39,7 @@ class DonationController extends Controller
         }])->latest()->paginate(25);
 
         $dailyDonations = Donation::query()
-            ->selectRaw('DATE(donations.created_at) as date, SUM(donation_packages.cost) as total')
+            ->selectRaw('DATE(donations.updated_at) as date, SUM(donation_packages.cost) as total')
             ->join('donation_packages', 'donations.package_id', '=', 'donation_packages.id')
             ->where('donations.status', '=', ModerationStatus::APPROVED)
             ->groupBy('date')
@@ -47,7 +47,7 @@ class DonationController extends Controller
             ->get();
 
         $monthlyDonations = Donation::query()
-            ->selectRaw('EXTRACT(YEAR FROM donations.created_at) as year, EXTRACT(MONTH FROM donations.created_at) as month, SUM(donation_packages.cost) as total')
+            ->selectRaw('EXTRACT(YEAR FROM donations.updated_at) as year, EXTRACT(MONTH FROM donations.updated_at) as month, SUM(donation_packages.cost) as total')
             ->join('donation_packages', 'donations.package_id', '=', 'donation_packages.id')
             ->where('donations.status', '=', ModerationStatus::APPROVED)
             ->groupBy('year', 'month')
--- a/app/Http/Controllers/Staff/HomeController.php
+++ b/app/Http/Controllers/Staff/HomeController.php
@@ -21,10 +21,10 @@ use App\Helpers\SystemInformation;
 use App\Http\Controllers\Controller;
 use App\Models\Group;
 use App\Services\Unit3dAnnounce;
+use Exception;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\DB;
 use Spatie\SslCertificate\SslCertificate;
-use Exception;
 
 /**
  * @see \Tests\Todo\Feature\Http\Controllers\Staff\HomeControllerTest
@@ -75,6 +75,7 @@ class HomeController extends Controller
                 ->where(fn ($query) => $query->whereNull('assigned_to')->orWhere('assigned_to', '=', auth()->id()))
                 ->count(),
             'pendingApplicationsCount' => DB::table('applications')->where('status', '=', ModerationStatus::PENDING)->count(),
+            'pendingDonationsCount'    => DB::table('donations')->where('status', '=', ModerationStatus::PENDING)->count(),
             'certificate'              => $certificate,
             'uptime'                   => $systemInformation->uptime(),
             'ram'                      => $systemInformation->memory(),
--- a/app/Http/Controllers/User/UserController.php
+++ b/app/Http/Controllers/User/UserController.php
@@ -120,7 +120,7 @@ class UserController extends Controller
                 ->first(),
             'watch'        => $user->watchlist,
             'externalUser' => ! $user->trashed() && $request->user()->group->is_modo ? Unit3dAnnounce::getUser($user->id) : false,
-            'donation'     => Donation::query()->where('status', '=', ModerationStatus::APPROVED)->where('user_id', '=', $user->id)->latest()->first(),
+            'donation'     => Donation::query()->where('status', '=', ModerationStatus::APPROVED)->where('user_id', '=', $user->id)->latest('updated_at')->first(),
         ]);
     }
 
--- a/app/View/Composers/TopNavComposer.php
+++ b/app/View/Composers/TopNavComposer.php
@@ -60,7 +60,7 @@ class TopNavComposer
             'donationPercentage' => value(function (): int|string {
                 $sum = Donation::query()
                     ->join('donation_packages', 'donations.package_id', '=', 'donation_packages.id')
-                    ->where('donations.created_at', '>=', now()->startOfMonth())
+                    ->where('donations.updated_at', '>=', now()->startOfMonth())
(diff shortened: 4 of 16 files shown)


app/Http/Controllers/Staff/DonationController.php
```diff
@@ -73,10 +73,21 @@ public function update(Request $request, int $id): \Illuminate\Http\RedirectResp
...
+            $donation->starts_at = $latestDonation->ends_at;
+        }
 
         if ($donation->package->donor_value > 0) {
-            $donation->ends_at = $now->addDays($donation->package->donor_value);
+            if ($latestDonation === null) {
+                $donation->ends_at = $now->addDays($donation->package->donor_value);
+            } else {
+                $donation->ends_at = $latestDonation->ends_at->addDays($donation->package->donor_value);
+            }
```
Roardom:
```suggestion
            $donation->ends_at = $donation->starts_at->addDays($donation->package->donor_value);
```

Roardom:
It's usually best to make separate PRs, one for each commit here. Some of the changes look quick to approve, others not so much.

I don't think your code accounts for if a donation is lifetime. Can you look into this?

app/Http/Controllers/Staff/DonationController.php
Roardom: Something like this would also keep it simple. You can make the assumption that users use UTC so don't have to worry about daylight savings either.

Roardom (changes requested):
Small thing I noticed

app/Http/Controllers/Staff/DonationController.php
Oha-you: Great idea, thank you. AI also suggested to add `->copy()` before `->addDays`
> Carbon mutates the object. Without `copy()`, you risk modifying `starts_at`

Oha-you:
> It's usually best to make separate PRs, one for each commit here. Some of the changes look quick to approve, others not so much.

I'll try to do so in future PRs, but some commits here are tied together.
It'll be great to merge all these changes after a review because the donation feature looked kinda unfinished to me.
But I can always move some of them to separate PRs. Feel free to ask.

> I don't think your code accounts for if a donation is lifetime. Can you look into this?

You were right. I think I managed to cover all possible scenarios now.
Basically, once a Lifetime package is approved for a user, all future donations will also have `ends_at === null`.
I removed `$donation->ends_at = null` assignment because this value is `null` by default, same as `starts_at`.

I also added a new commit since I forgot to fix hardcoded values in the other Staff views: https://github.com/HDInnovations/UNIT3D/pull/5282/changes/ef1335ba17bc28434206f5434cc0edd69a677ab7

There I extended default SweetAlert2 options for `b64DeletionMessage` in `resources/views/layout/default.blade.php` by adding HTML support.
It shouldn't be less secure this way since you'll have to explicitly allow it only for certain modals via `data-allow-html="true"`.

Roardom:
Haven't had time to look it over yet, but briefly saw this:

> It shouldn't be less secure this way since you'll have to explicitly allow it only for certain modals via data-allow-html="true".

This is an XSS vulnerability since the `username` field is user-controllable.

Oha-you:
> This is an XSS vulnerability since the `username` field is user-controllable.

Fair. Added escaping with [e( )](https://laravel.com/docs/12.x/strings#method-e) helper to usernames, package and gateway names.

It indeed helped:

donate-sure

Oha-you:
The last commit here: https://github.com/HDInnovations/UNIT3D/pull/5282/commits/47377ebe32fb1583899a1074a8a36b6398ea8363
It's also an important one because some crypto addresses like XMR are long strings that don't fit into the limited input width of the modal.
Making inputs `readonly` and selecting their values "on focus" allows users to easily copy them with Ctrl+C or through the context menu.

Oha-you:
@Roardom, any chance of a review?

Or do you want this PR to be split after all?

Roardom:
Sorry, been a bit busy lately. Personally, I think the schema's wrong if all these workarounds are in order for multiple donations. starts_at and ends_at per donation don't make sense if donations can overlap. There should probably be a created_at and approved_at instead, and the users table should have a nullable ends_at that is increased every time a donation is approved. And you'd have to incorporate the logic somehow into lifetime donations as well. Maybe you can give some thought to it and come up with an idea?

Oha-you:
> I think the schema's wrong if all these workarounds are in order for multiple donations

In my opinion, the current `donations` table schema in the development/master branch is wrong.

Since the date this feature was added in #4157 and merged into release [v8.3.0](https://github.com/HDInnovations/UNIT3D/releases/tag/v8.3.0) on October 7th, 2024, every time a donor made a donation, they've been robbed of up to a full day of what they paid for.
Or if some generous donor wanted to support the tracker a second time while still having an active status, their end time was being replaced with a new date as if they've never made the first donation.

This makes every UNIT3D tracker staff look bad, even if no one has noticed it yet. Or it adds the need for staff to adjust the end date manually each time.
All because the donation feature, like some others, was essentially a draft and hasn't been reviewed properly.
That's why it has all those hardcoded `$` in the views and popups that say "Are you sure you want to delete this _page_".

This pull request fixes everything that should've already existed in #4157 in the first place.
Without making any groundbreaking changes, by carefully patching the existing schema and logic behind it.

> starts_at and ends_at per donation don't make sense if donations can overlap

Donno, it makes perfect sense to me.
Each donation is a purchase of a paid package that allows users to receive a VIP status for X amount of days.
This status begins at `starts_at` time and ends at `ends_at` time.
These timestamps are different from the time a donation was introduced into the system by the user - `created_at`.
Or the time a donation was approved or rejected by a staff member - `updated_at`.
The fact that all these 4 columns exist allows us to support a lifetime donation easily by just setting `ends_at` to `Null`. That's how it's already implemented.
And if someone sends multiple donations, we just set `starts_at` of a new donation to `ends_at` of the existing one.
If the last one was lifetime, `starts_at` will represent the donation approval date. End time will remain `Null`.

> There should probably be a created_at and approved_at instead, and the users table should have a nullable ends_at that is increased every time a donation is approved

I don't see why this is needed. `updated_at` can only be changed by approving or rejecting a donation.
That's why this PR changes all `created_at` usages to `updated_at`, because it is logically correct.

If you want `approved_at`, what should happen when a donation gets rejected? Should it be soft deleted? Then we need `deleted_at` too.
Also I don't think `users` table needs any additional columns if it can be avoided. It's probably not good for performance, though I'm not sure.

> And you'd have to incorporate the logic somehow into lifetime donations as well

All the logic for lifetime donations is incorporated into this PR already. People can buy lifetime packages, then normal ones, nothing will break.

> Maybe you can give some thought to it and come up with an idea?

I think I've already given this PR enough thought. Thanks to your review points as well.
If you want anything about the donation feature to be refactored, it can be done later without a rush.

The changes proposed here need to be reviewed, merged and released ASAP, in my opinion.
I still don't understand how people haven't noticed and reported the main issue with dates in 1.5 years since the release.
Hopefully the trackers will find some quality of life improvements like better popup messages useful as well.

Roardom:
> every time a donor made a donation, they've been robbed of up to a full day of what they paid for.

You don't get robbed from donations. Donations are willingly donated from donors without expecting anything in return. If you're treating it like a transaction where a user purchases a product expecting something in exchange, you should not be using this feature.

Oha-you:
I'm afraid I can't agree with you here.
These are the types of donations where you don't expect anything in return: https://hdinnovations.github.io/HDInnovations/donate.html
Any amount, no perks, just pure support.

While the donation feature introduced in #4157 is essentially a "pay to leech" product offering that's built into UNIT3D's core now. And UNIT3D-Announce's core too, for that matter.
You can't even disable global freeleech and immunity every donor receives by donating without customizing the announce controller. For instance, so that only cosmetic perks are applied.

But it's fine. It's what many trackers wanted and what they finally received. Although with some issues that get fixed in this PR.
So there's a very clear expectation of what exactly you get by spending X amount of money. Counted as "X days of perks" on Blutopia, for example.
There's no need to beat around the bush. It is what it is.
You can preface these package purchases with whatever disclaimer you want. But they'll still be transactions.

Let's say you support a creator on Patreon. Would you be happy if the platform rounded your subscription time in its favor? Causing you to spend extra money each month?
If you think it's fine, why not go further and subtract 1 or 2 days from `ends_at` time as well? Surely the tracker will receive even more support this way.

Maybe the phrase "robbed of" I used sounded a bit harsh. Apologies if it was the case.
But I'm sure that if Patreon applied the same logic to you, you'd use even harsher words like "scammed".
And you as a supporter would probably blame the creator for taking the money from you on the day when you still should've had an active status.
Or, in our case, the blame is on the tracker staff who haven't reviewed the code they merged.
Though in all fairness the blame should be placed on those who coded it this way in the first place.

I'll ignore your suggestion about what features I should or shouldn't use because it's really not for you or me to decide.
But I can say that the owner of the tracker I'm a coder at bought and installed the donation module back in 2020 when it was still paywalled.
And another owner who I've been helping was a long-time GitHub supporter of UNIT3D until the tracker closure.

Interestingly, `DateTime` type was used for `vip_expires` values back then.
But even the paid module didn't support multiple donations so I had to fix it at that time too:

donate-2020

To my surprise, 5 years later I saw even more issues. Which once again makes me suspect that these draft features are released in an unfinished state deliberately to promote the paid services: https://hdinnovations.github.io/HDInnovations/
Otherwise I can't find a reason why you are so against reviewing and merging these objectively very much needed fixes.

Roardom:
Just to make sure we're extra clear: https://github.com/HDInnovations/UNIT3D/blob/8b88f4c8182eb3d3912ffef425c3224dcfd596f4/CONTRIBUTING.md#L112

Focus on the code and stop embarrassing yourself with loaded language and personal speculations.

If you're going to submit PRs with so many things intertwined that it takes multiple hours to review - well I don't have that time, so your PR won't get reviewed. This is not getting merged as is.

PR #5283, opened 2026-02-19
(Fix) Select background color, correct CSS property

Oha-you:
Currently `background-color` is ignored because `background` overrides it to `transparent`.
MDN: [Shorthand_properties#tricky_edge_cases](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Shorthand_properties#tricky_edge_cases)

`--select-icon` is a `url()`
https://github.com/HDInnovations/UNIT3D/blob/fd63b96269b7c17396e968ab65450d2e4222c3f1/resources/sass/themes/_galactic.scss#L323

Before and after:

form-bg-color

### DIFF
--- a/resources/sass/components/form/_select.scss
+++ b/resources/sass/components/form/_select.scss
@@ -15,7 +15,7 @@
 
 .form__select {
     background-color: inherit;
-    background: var(--select-icon);
+    background-image: var(--select-icon);
     background-size: 16px;
     background-position: calc(100% - 4px) center;
     background-repeat: no-repeat;


Roardom:
While the gif looks great and docs make sense, I tried both firefox and chrome on linux and my select dropdowns look significantly better. And I'd be surprised if vinnie hasn't yet experienced this on macos. Is this a windows-only issue?

Roardom:
Firefox:

image

Chrome:

image

From before this change

Oha-you:
It was Firefox on Windows 10, latest version, yeah. Also 125% scaling was enabled system-wide.
But something got updated overnight and now the width of this select box is 100% for me too.
No idea why, all conditions are the same, only PC was rebooted.

I verified that I could see this white background on all major UNIT3D trackers like Blutopia and Aither. Took a note about it 2 weeks ago.
Maybe the width limit came from some uBlock Origin's cosmetic filters, like Annoyances. I wish I inspected it more, but oh well.

Anyway, I still see the white border in a newly installed Firefox 147.0.4 Portable without any extensions or custom CSS.
Animation below is from BLU so you know it's not my custom styles:

select-bg-color

HDVinnie:
> While the gif looks great and docs make sense, I tried both firefox and chrome on linux and my select dropdowns look significantly better. And I'd be surprised if vinnie hasn't yet experienced this on macos. Is this a windows-only issue?

I haven't experienced this on any browser on my Mac. Safari, OperaGX or Firefox. That being said Mac has its own dropdown style it seems.

Roardom (approved):

Roardom:
Thank you! Seeing as you use Windows, I'd definitely appreciate it if you could take a look into why the select dropdown looks terrible on that OS. If you have the time.

PR #5285, opened 2026-02-19
(Fix) Update selected option logic in giveaway edit view

OrbitMPGH:
This pull request makes a minor fix to the giveaway edit form to ensure that the correct prize type is selected when editing a giveaway. Currently, it defaults to BON when you edit a prize.

* Updated the `@selected` directive in the `resources/views/Staff/giveaway/edit.blade.php` form so it checks `$prize->type` instead of `$giveaway->type`, ensuring the correct option is pre-selected based on the prize's type.


Roardom:
Thank you!

PR #5286, opened 2026-02-20
(Fix) Unassign staff from tickets on demotion

### LINKED ISSUE #5281: [Bug] Ticket Notifications Send After Demotion
### UNIT3D Version

9.2.0

### PHP & Platform

8.4.5 - Ubuntu 24.04

### Laravel verion

12.26.4

### Have you done this?

- [x] I am willing to share my stack trace and logs

### Expectation

After a staff member loses his staff rank I expect him to not get notifications to previously assigned tickets.

### Description

Instead, assigned ticket notifications continue for each response even though they cannot actually access the ticket anymore.

### Stack trace & logs

```shell

```

OrbitMPGH:
Remove a user as an assignee from all their tickets when getting demoted from a staff user group.
This is to prevent notifications from tickets that they can no longer access.

Fixes #5281

### DIFF
--- a/app/Http/Controllers/Staff/UserController.php
+++ b/app/Http/Controllers/Staff/UserController.php
@@ -33,7 +33,6 @@ use App\Models\Post;
 use App\Models\PrivateMessage;
 use App\Models\Scopes\ApprovedScope;
 use App\Models\Thank;
-use App\Models\Ticket;
 use App\Models\Topic;
 use App\Models\Torrent;
 use App\Models\User;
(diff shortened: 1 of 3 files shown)


Roardom (commented):
I'm not sure this is the way we want to go. It might be more desirable to check is_modo upon sending each notification instead of wiping out the staff's entire ticket history. I'll get back to you.

Roardom (changes requested):
Sorry for the delay. Got confirmation and indeed we don't want to wipe out ticket history. Instead, we just want to block notifications to the ticket assignee if they don't belong to an is_modo group. Additionally, the ticket assignee should be unassigned if a new ticket comment is left and the existing ticket assignee is no longer staff.

PR #5288, opened 2026-02-20
(Refactor) Use composite action for modular php/js setup

Roardom:
Cleans up the actions a bit

PR #5289, opened 2026-02-20
(Refactor) Remove redundant ci action cache save step 

Roardom:
The `cache` action automatically saves the path at the end of the job.

Depends on #5288


Roardom:
If the job fails, the cache won't be saved, which can still be useful

PR #5293, opened 2026-02-22
(Fix) Return error if meta not found when adding wishlist entry

### LINKED ISSUE #5169: [Bug] Torrent Category Handling on Wishlist
### UNIT3D Version

v9.2.0 (Development Branch)

### PHP & Platform

8.4.5

### Laravel verion

12.26.4

### Have you done this?

- [x] I am willing to share my stack trace and logs

### Expectation

I expect wish list buttons and links to work as intended.

### Description

Instead there are a few issues regarding the torrent category handling, after wish listing a tv show the torrents link sets the category to movies and doesn't apply the TMDB ID at all, and the request button similarly adds the category for movies by default instead of using the tv one.

There could also be better handling for 404s on wish list adding, adding a TMDB ID under the wrong category (like selecting movie instead of tv) shows a 500 error page instead of simply saying the TMDB ID doesn't exist.

### Stack trace & logs

```shell
The 500 error didn't log anything without debug enabled.
```

Roardom:
fixes #5169

### DIFF
--- a/app/Http/Controllers/User/WishController.php
+++ b/app/Http/Controllers/User/WishController.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
 
 namespace App\Http\Controllers\User;
 
+use App\Exceptions\MetaFetchNotFoundException;
 use App\Http\Controllers\Controller;
 use App\Http\Requests\StoreWishRequest;
 use App\Models\Category;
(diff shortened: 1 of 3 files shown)


HDVinnie (changes requested):
Failing CI

PR #5335, opened 2026-03-18
(Refactor) Merge chatbox conversations

Roardom:
Rather substantial change. I tested as much as I could.


HDVinnie (approved):
Im away for a few weeks for work, so I'm unable to test locally however I don't see anything wrong looking at the diff.

PR #5342, opened 2026-03-21
(Fix) Change donation dates to timestamps, prevent time loss

Oha-you:
_Split from #5282_

Using `date` type for donation's `starts_at` and `ends_at` times leads to up to 24 hours of lost paid time for all donors.

This change fixes it by converting these values to a `timestamp` type.

It's better than `datettime` because no time will be lost or added even if server timezone changes.

Also run [AutoRemoveExpiredDonors](https://github.com/HDInnovations/UNIT3D/blob/development/app/Console/Commands/AutoRemoveExpiredDonors.php) script hourly to prevent extra added time.


Roardom:
This all good now? Noticed the recent force push

Roardom (approved):

Oha-you:
Yes, I just appended the `NULL DEFAULT NULL` change generated from [MySQL Schema Dump](https://github.com/HDInnovations/UNIT3D/actions/runs/23413365315/job/68104396659?pr=5342) commit. It finally worked!

PR #5345, opened 2026-03-21
(Fix) Donation Staff views, remove hardcoded currency, make consistent

Oha-you:
_Split from #5282_

"edit" views are made consistent with "create" views, hardcoded currency is removed

### DIFF
--- a/resources/views/Staff/donation-gateway/edit.blade.php
+++ b/resources/views/Staff/donation-gateway/edit.blade.php
@@ -41,7 +41,7 @@
                         <tr>
                             <td>
                                 <input
-                                    type="text"
+                                    type="number"
                                     name="position"
                                     value="{{ $gateway->position }}"
                                     class="form__text"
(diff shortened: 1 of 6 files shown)


resources/views/Staff/donation-gateway/edit.blade.php
```diff
@@ -41,7 +41,7 @@
                         <tr>
                             <td>
                                 <input
-                                    type="text"
+                                    type="number"
```
Roardom: `type="number"` should not be used. Add `inputmode="numeric" pattern="[0-9]*"` instead.

Roardom (changes requested):

resources/views/Staff/donation-gateway/edit.blade.php
Oha-you: Done, and also added `required` attributes to `not null` inputs.
Used `[0-9]+` pattern for required fields.

Cost is a decimal, so the pattern is `[0-9]+(?:\.[0-9]{1,2})?`
https://github.com/HDInnovations/UNIT3D/blob/d274132f358eb876b97e31308bfaddfc7340ab26/database/migrations/2024_05_06_212410_create_donation_packages_table.php#L32

PR #5346, opened 2026-03-21
(Fix) Confirmation popups for Donations, improve messages

Oha-you:
_Split from #5282_

Better text for confirmation popups, add support for HTML with proper escaping.


Roardom (changes requested):
Can you please do this without html? I'd rather not allow this functionality in the codebase, even if it means the message isn't formatted as nicely.

Oha-you:
Sure, how about this then.
I left `e()` escaping just in case.

PR #5347, opened 2026-03-21
(Add) Select all text when focusing donation gateway inputs

Oha-you:
_Split from #5282_

Some trackers also add a Copy button to the right of the input, but they can code it themselves.

This is just a better default to make sure people don't send money to wrong addresses. Or don't get discouraged when they see an error.

### DIFF
--- a/resources/views/donation/index.blade.php
+++ b/resources/views/donation/index.blade.php
@@ -132,8 +132,9 @@
                                 <input
                                     class="form__text"
                                     type="text"
-                                    disabled
+                                    readonly
                                     value="{{ $gateway->address }}"
+                                    x-on:focus="$el.select()"
                                     id="{{ 'gateway-' . $gateway->id }}"
                                 />
                                 <label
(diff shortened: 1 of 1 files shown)


resources/views/donation/index.blade.php
```diff
@@ -132,8 +132,9 @@ class="dialog__form"
                                 <input
                                     class="form__text"
                                     type="text"
-                                    disabled
+                                    readonly
                                     value="{{ $gateway->address }}"
+                                    x-on:focus="$el.select()"
```
Roardom, line 137: Can you try https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/user-select#all instead? I always try and prefer CSS over JS.

Oha-you: Tried it, no success:

> `user-select: all` doesn't work on `<input>` elements because inputs are form controls, not regular text content. The `user-select` CSS property controls selection behavior for static text in the DOM - it's not designed for interactive form fields.

PR #5349, opened 2026-03-21
(Add) Support multiple donations by the same user, adjust start/end times

Oha-you:
_Split from #5282_


Oha-you:
Turns out there was one scenario that hasn't been covered: a donor with past donations that have already expired.

Added this code and renamed the variable to `$activeDonation` for clarity:
```php
->where(function ($query) use ($now): void {
    $query->where('ends_at', '>', $now)
        ->orWhereNull('ends_at');
})
```

HDVinnie (approved):

PR #5188, opened 2026-01-01
(Update) Dependencies

HDVinnie:

PR #5351, opened 2026-03-22
(Update) prettier-plugin-blade to v3 (chisel)

Oha-you:
Currently used `prettier-plugin-blade` version was published 2 years ago on [2023-12-23](https://www.npmjs.com/package/prettier-plugin-blade/v/2.0.0):
https://github.com/HDInnovations/UNIT3D/blob/685f09c76e187ee2b0132be86ee63b106b357a38/package.json#L23-L24
- It was frozen in #4394 without an explanation
- Even though Prettier itself was updated in #5188 after being frozen in #5088
- `v2.0.0` was the first version to support Prettier v3
- It received a lot of updates in 2024-2025 with many issues fixed: [changelog.md](https://unpkg.com/prettier-plugin-blade@2.1.21/changelog.md)
- Recently this plugin was renamed to [chisel](https://github.com/fortephp/chisel)
- It's a major rewrite that changes the output, but also solves some core bugs

I noticed that the current `prettier-plugin-blade v2.0.0` behavior doesn't respect [whitespace-sensitive formatting](https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting) that exists in Prettier since 2018.
- It can be adjusted with [htmlWhitespaceSensitivity](https://prettier.io/docs/options#html-whitespace-sensitivity) option
- But it only works in the newly released v3 version, as confirmed by the author in https://github.com/fortephp/chisel/issues/93
- I also tried [shufo/prettier-plugin-blade](https://github.com/shufo/prettier-plugin-blade), but it was ignored there too

---

Without this newly updated plugin the following code:

```html
<p>Then upload it somewhere (like <a href="https://imgur.com/upload" rel="noopener noreferrer" target="_blank"><b>Imgur</b></a> or <a href="https://imgbb.com/" rel="noopener noreferrer" target="_blank"><b>ImgBB</b></a>) and paste the link below.</p>
```
Gets formatted into:
```html
<p>
    Then upload it somewhere (like
    <a
        href="https://imgur.com/upload"
        rel="noopener noreferrer"
        target="_blank"
    >
        <b>Imgur</b>
    </a>
    or
    <a href="https://imgbb.com/" rel="noopener noreferrer" target="_blank">
        <b>ImgBB</b>
    </a>
    ) and paste the link below.
</p>
```
Which leads to the following issue on the frontend:

whitespace-bug

While the updated version converts it into:
```html
<p>Then upload it somewhere (like <a href="https://imgur.com/upload" rel="noopener noreferrer" target="_blank"><b>Imgur</b></a> or <a href="https://imgbb.com/" rel="noopener noreferrer" target="_blank"><b>ImgBB</b></a
>) and paste the link below.</p>
```
It may not be perfect, but it doesn't add any whitespace bugs.

---

I also installed [@prettier/plugin-php](https://github.com/prettier/plugin-php) to format the embedded PHP fragments since chisel mentioned it in the description.
Only Blade templates will be affected this way.

Config needed some refactoring because it had overrides for different file types with identical options.
I just moved them into the global scope and got rid of the overrides.

It will affect `*.scss` files because now they will also use `"printWidth": 100`
It was `80` before since it's the [default value](https://prettier.io/docs/options#print-width). This change will make the rules for all styles, templates and scripts consistent.

`"bladePhpFormatting": "safe"` is the default value, but I included it for clarity and to force the safe usage if the defaults change in the future.

Added `"bladeKeepHeadAndBodyAtRoot": false` to keep the current style intact.

And `"bladeDirectiveArgSpacing": "preserve"` was needed because otherwise spaces are added before parenthesis for `@include()` and `@section()`, same as for `@if ()` or `@switch ()`.
It retains the current code style, but won't fix issues like `@if(true)` without a space (which already exist in the code).

---

`.prettierignore` had a non-existent path to a template that got renamed in #4698
To comment on what's said in the [commit message](https://github.com/HDInnovations/UNIT3D/commit/5f60ce6b77fabfcfa576551a4dc1a7f6fa275efd): this file was just ignored since #4157
> Seems to have always been an issue, but the CI must only be formatting files that changed and because the file was only just changed now, it never discovered it until now.

Ideally if/when this PR is merged, all Blade templates, SCSS styles and JS scripts will need to be processed.
The sooner it happens, the better.

Also you might want to freeze these package versions if you're worried about GitHub actions failing in the future.
But I think it's better to leave them as is since all changes are usually bug fixes that otherwise will get ignored for years (like it happened with `prettier-plugin-blade v2.0.0`).


Oha-you:
UPD: I created an issue for the updated [chisel](https://github.com/fortephp/chisel/issues) plugin, so that maybe the current `bladeDirectiveArgSpacing` behavior can be extended to distinguish between control and function-like directives.

Looks like this functionality was lost after the rewrite. But it shouldn't be hard to bring it back.

HDVinnie:
You need to run prettier on this PR and commit the changes so CI passes.

Oha-you:
Looks like failure comes from `stefanzweifel/git-auto-commit-action@v7` action, same as in #5342

I've enabled Prettier in JetBrains IDEA so all the code is auto-formatted on file save. And I run `pint` there too.

But also let me convert this PR to a draft because chisel's author might fix the issue I mentioned soon (it's the only one currently).

Oha-you:
A simple re-run helped, I assume?
I had no problems with my fork here too: https://github.com/Oha-you/UNIT3D/pulls

Now you can see exactly how many changes will come with this update.
If some of them are undesirable, I can try to find a workaround. Or create issues / PRs in chisel's repository.

HDVinnie:
I'm in Baltimore again for next three weeks so no computer with me. I'll try to review from GitHub mobile app when I can on nights. That being said this PR really comes down to opinions. So as long as all core devs are in agreement we'll have to find config/ruleset we all are ok with.

Roardom (changes requested):
Nice to hear that prettier-plugin-blade got an update. I think it's best to keep this PR as a draft until most of the backwards incompatibilities are solved (and be reported or fixed upstream).

> I also installed [@prettier/plugin-php](https://github.com/prettier/plugin-php) to format the embedded PHP fragments since chisel mentioned it in the description.

v2 used to have an option of using pint for formatting within PHP. I could never get it to work (I believe it kept removing closing parentheses after @if statements which broke the code - https://github.com/HDInnovations/UNIT3D/pull/3234#issuecomment-1868922913). If we're going to be formatting the PHP in blade, it should use pint formatting like the rest of the php code and not use prettier formatting. Until this is possible, disabling it would be better than this 4k+ diff.

> It will affect *.scss files because now they will also use "printWidth": 100
It was 80 before since it's the [default value](https://prettier.io/docs/options#print-width). This change will make the rules for all styles, templates and scripts consistent.

scss and js should remain at 80. The only reason blade was kept at 100 was to reduce the size of the initial formatting diff when prettier was first added, and was a more natural breakpoint for the existing code. SCSS and JS have been written with 80 in mind already.

> Also you might want to freeze these package versions if you're worried about GitHub actions failing in the future.

I agree. Even though this PR only updates us to v3, we've had at least half a dozen breaking changes with the formatter. This is also another reason why I'd rather wait for the dust to settle rather than jumping into things. There's just so many times where ci has reformatted existing code to a new style.

Side note, but it would be a good idea to fully read through all the code changes. When I first added prettier in #3234, it made changes to blade files that broke the the syntax, and I had to play with the configuration and file bug reports until I got it working. It's best not to assume the formatter is correct. (Although I am *very* happy to see no signs of AI in v3)

Lastly, considering chisel is still new, it might be worth updating prettier-blade-plugin to the final v2 release, which would let us use all the bug fixes, without all this breakage. Then, once chisel has its backwards incompatibilities slimmed down, it can be upgraded a second time. That would be the easiest way of improving prettier without all the breaking changes (and would be something I'd approve of). As is, I don't want to merge this code because 1) of all the backwards incompatible changes, and 2) I prefer the Pint style over the Prettier style for PHP code. Some examples of things I don't like include removing the new line before `break`, spreading every parameter within a function on its own line and removing trailing commas in arrays. It also seems to break when using {{}} inside of js inside of blade.

Oha-you:
Good arguments, I also thought about version 2.1.21 for the time being.
New pull request: #5352

I removed `plugin-php` from here and adjusted the config to only use `"printWidth": 100` for Blade templates.

I think I gonna close this PR after the automated commit is added. It'll still be useful for future comparisons.

PR #5352, opened 2026-03-23
(Update) Prettier, freeze the latest version, refactor the config

Oha-you:
Hopefully only non-breaking changes with minimal diff compared to #5351
Changelog: [Prettier](https://github.com/prettier/prettier/compare/3.7.4...3.8.1) ([release notes](https://prettier.io/blog/2026/01/14/3.8.0) and [here](https://github.com/prettier/prettier/blob/main/CHANGELOG.md)) + [prettier-plugin-blade](https://unpkg.com/prettier-plugin-blade@2.1.21/changelog.md).

However, `printWidth` for JavaScript was changed from `100` back to `80`, as advised in https://github.com/HDInnovations/UNIT3D/pull/5351#pullrequestreview-3989086637

Unfortunately [whitespace-sensitive formatting](https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting) is still not available in v2.
So whitespace bugs for inline elements might still be added by `prettier-plugin-blade`.

But they can be fixed by disabling the formatter with `{{-- format-ignore-start --}}` and `{{-- format-ignore-end --}}` comments.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -12,7 +12,7 @@
 !/resources/js/
 /resources/js/vendor
 
-# except resources/scss, but ignores resoruces/scss/vendor
+# except resources/scss, but ignores resources/scss/vendor
 !/resources/sass/
 /resources/sass/vendor
 
@@ -20,4 +20,3 @@
 resources/views/emails
 resources/views/rss/show.blade.php
 resources/views/vendor
-resources/views/components/user_tag.blade.php
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,6 +2,12 @@
   "plugins": [
     "prettier-plugin-blade"
   ],
+
+  "printWidth": 80,
+  "semi": true,
+  "singleQuote": true,
+  "tabWidth": 4,
+
   "overrides": [
     {
       "files": [
@@ -9,9 +15,7 @@
       ],
       "options": {
         "parser": "blade",
-        "printWidth": 100,
-        "semi": true,
-        "singleQuote": true
+        "printWidth": 100
       }
     },
     {
@@ -19,20 +23,7 @@
         "*.js"
       ],
       "options": {
-        "printWidth": 100,
-        "semi": true,
-        "singleQuote": true,
-        "tabWidth": 4
-      }
-    },
-    {
-      "files": [
-        "*.scss"
-      ],
-      "options": {
-        "semi": true,
-        "singleQuote": true,
-        "tabWidth": 4
+        "printWidth": 100
       }
     }
   ]
--- a/bun.lock
+++ b/bun.lock
@@ -19,7 +19,7 @@
       },
       "devDependencies": {
         "laravel-vite-plugin": "^2.0.1",
-        "prettier": "3.7.4",
+        "prettier": "3.8.1",
         "prettier-plugin-blade": "2.0.0",
         "vite": "^8.0.0",
         "vite-plugin-static-copy": "^3.1.4",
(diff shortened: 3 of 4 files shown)


Roardom:
The only way to get the CI to commit to your branch is to invite the bot to your repo. Even though you check the "Allow maintainers to make changes" box, I still haven't found any other way around it that would still be secure. You're better off just committing the changes the formatter made in another commit, and then ensuring that the prettier ci action doesn't fail to ensure it's correct.

Oha-you:
I think I found a workaround. Just created a pull request in my own fork with the same automated actions.
So this commit above is the actual diff: https://github.com/HDInnovations/UNIT3D/pull/5352/commits/2efe5905180ecd461229166c4bfe570b621d8bfe

Oha-you:
Very fun changes made by the 2.1.21 plugin version:
https://github.com/Oha-you/UNIT3D/blob/2efe5905180ecd461229166c4bfe570b621d8bfe/resources/views/blocks/chat.blade.php#L104

I narrowed down this gibberish text issue to the exact version when it starts to happen: **2.1.15**
So it looks like 2.1.15-2.1.21 are faulty, unfortunately.

Let me try to freeze it at **2.1.14** then.

Roardom:
> However, printWidth for JavaScript was changed from 100 back to 80, as advised in https://github.com/HDInnovations/UNIT3D/pull/5351#pullrequestreview-3989086637

My bad, I thought JS was already using 80. Can you keep it at 100 for now?

Otherwise, these changes look good.

Oha-you:
> Can you keep it at 100 for now?

Yep, alright.

But now these bugs have been introduced.
- no space before `as`: https://github.com/HDInnovations/UNIT3D/blob/8cab84872da3819ebd4e7b5130a8cd654bdf4429/resources/views/livewire/subtitle-search.blade.php#L134
- strange whitespace added: https://github.com/HDInnovations/UNIT3D/blob/8cab84872da3819ebd4e7b5130a8cd654bdf4429/resources/views/livewire/ticket-search.blade.php#L104-L109
- excessive parenthesis: https://github.com/HDInnovations/UNIT3D/blob/8cab84872da3819ebd4e7b5130a8cd654bdf4429/resources/views/giveaway/show.blade.php#L28
And the first one was only fixed in v2.1.20 that can't be used because of the gibberish text it produces for `blocks/chat.blade.php`

Gonna try even earlier versions then...

Roardom:
Fun 😅. If this ends up being an issue, maybe we should think about reporting minimal reproducible examples for v3 until it's ready and update to v3.

Oha-you:
Well, here's the conclusion: all three bugs above can be reproduced in every `prettier-plugin-blade` version starting from `2.1.0`. I checked all 22 versions at this point.

So it looks like the changes made in `2.1.0` were so significant that the author had to spend the next 2 years fixing various quirks caused by them.

And version `2.0.0` has none of the bugs above. So Vinnie was right to freeze it after all.
But at least the main plugin has been updated without any issues 😄 

And it hasn't caused a single change in any templates, styles or scripts (after I returned `"printWidth": 100` there).
I ran the actions in my fork as well to make sure: https://github.com/Oha-you/UNIT3D/pull/5

P.S. As an experiment I also tried to replace the formatter with [shufo/prettier-plugin-blade](https://github.com/shufo/prettier-plugin-blade).
It produced a huge diff: https://github.com/Oha-you/UNIT3D/pull/6
And somehow one commit wasn't enough. Another one was generated on a second action run.
Some options can probably be changed to reduce the diff. But it's not worth it, let's just stick to v2.0.0 for now.

HDVinnie:
I tried the shofu one in past and it had many bugs. I think I even opened a few on his repo.

HDVinnie (approved):

PR #5354, opened 2026-03-23
(Remove) Spatie\CookieConsent package

Oha-you:
[spatie/laravel-cookie-consent](https://github.com/spatie/laravel-cookie-consent) is something that existed in UNIT3D since v1 but has never been used in more than 9 years.

Like it's said in the package description, it does nothing other than showing a button and setting a cookie when it's clicked. 
No cookies are blocked before it happens.

So it has always been just an annoyance that most users probably blocked by default with their adblockers.
But since the `<div>` with `.js-cookie-consent` class gets removed after the DOM is loaded, users see a page jump:

cookie-msg1

cookie-msg2

Even though an environmental variable `COOKIE_CONSENT_ENABLED` existed, I doubt many people knew about it since it hasn't been mentioned anywhere.

This package neither adds a compliance with the EU laws, nor serves any other purpose.
And the text "Your experience on this site will be improved by allowing cookies" is just incorrect.
Cookies are required for UNIT3D to function.
And advanced users who still want to block something don't need any reminders.


HDVinnie:
@Roardom i have no issue with this being removed. It was honestly only added when the eu cookie law was being pushed.

HDVinnie (approved):

Roardom:
Figured I'd get you to review since I tried to do this years ago and you objected 👍

Roardom:
Only thing I'd be worried about merging this is if it will cause merge conflicts with weblate. Ideally only English localizations should be touched. Can you restore the non-English localizations?

Oha-you:
Sure, and renamed the commit as well

PR #5355, opened 2026-03-23
(Remove) Livewire's assets publishing to the public folder

Oha-you:
`vendor:publish --tag=livewire:assets` was first added in #1480 probably by following the [2.x installation guide](https://laravel-livewire.com/docs/2.x/installation#publish-assets).

But Livewire is bundled into `app.js` from the root `/vendor` folder now:
https://github.com/HDInnovations/UNIT3D/blob/d274132f358eb876b97e31308bfaddfc7340ab26/resources/js/app.js#L38-L39

So `public/vendor/livewire` is not used anywhere.
Maybe it was needed at some time in the past, but I couldn't find it.

I only found that `resources/js/vendor/alpine.js` was removed in #3608
And that [this code](https://livewire.laravel.com/docs/3.x/installation#:~:text=Livewire%2C%20Alpine) was probably used as an example.

[3.x](https://livewire.laravel.com/docs/3.x/installation#publishing-livewires-frontend-assets) and [4.x](https://livewire.laravel.com/docs/4.x/installation#publishing-livewires-assets-to-public-directory) installation guide even has these notes:

> Publishing Livewire's assets isn't necessary for Livewire to run. Only do this if you have a specific need for it.

> When you need this: If you want to serve Livewire's JavaScript through your web server directly (e.g., for CDN distribution or specific caching strategies) instead of Laravel routing.

> Publishing assets is rarely necessary. Only do this if you have a specific architectural requirement that prevents Laravel from serving the assets dynamically.


Roardom (approved):
Makes sense

PR #5360, opened 2026-03-24
(Update) Improve auto freeleech logic, sort by regex and size

Oha-you:
Auto freeleech feature introduced in #3486 is most welcome.

Issue 1:
1. A happy owner goes into the Staff Dashboard and adds two simplest rules:

    auto-free

3. A user uploads a 120 GB torrent to the tracker
4. Freeleech value is set to 50% instead of 100% _(???)_

Issue 2:
1. A user uploads a torrent exactly 50 GB (53687091200 bytes) in size
2. No freeleech buff is set (because of `<` instead of `<=`)

This PR fixes both issues.

It is not correct to just sort by position and expect everyone to magically know that auto freeleech rules should be sorted manually and that lower position means higher priority.

Regex rules should take priority because they are made for exceptions like title matching.
A regex rule that's added later (higher position) should take priority over earlier rules, otherwise it wouldn't have been added.
Then we sort by size regardless of positions.

Most trackers probably don't use regex matching, so they won't need to worry about positioning at all.

This PR is more of a bug fix, but I named it "update" as to not hurt anyone's feelings.

### DIFF
--- a/app/Helpers/TorrentHelper.php
+++ b/app/Helpers/TorrentHelper.php
@@ -56,11 +56,13 @@ class TorrentHelper
 
         if (!$torrent->free) {
             $autoFreeleeches = AutomaticTorrentFreeleech::query()
-                ->orderBy('position')
                 ->where(fn ($query) => $query->whereNull('category_id')->orWhere('category_id', '=', $torrent->category_id))
                 ->where(fn ($query) => $query->whereNull('type_id')->orWhere('type_id', '=', $torrent->type_id))
                 ->where(fn ($query) => $query->whereNull('resolution_id')->orWhere('resolution_id', '=', $torrent->resolution_id))
-                ->where(fn ($query) => $query->whereNull('size')->orWhere('size', '<', $torrent->size))
+                ->where(fn ($query) => $query->whereNull('size')->orWhere('size', '<=', $torrent->size))
+                ->orderByRaw('name_regex IS NULL')  // regex rules first
+                ->orderByDesc('position')           // then regex priority
+                ->orderByDesc('size')               // then size tiers
                 ->get();
 
             foreach ($autoFreeleeches as $autoFreeleech) {


Roardom:
Working as intended

Oha-you:
@HDVinnie, do you agree with this PR closure without a proper explanation?

HDVinnie:
I mean, I agree with the core issue here. It is valid. We unfortunately don't have a documents team where we could have a section for features. What they are and how to use them. Being that this project is FOSS, Payless and done for fun during free time. It's hard to get everything done.

That being said, I think the better approach here might just be to add a @section('sidebar') it explains priority. Or on edit and create views for priority input use the label hints roar added in past to explain lower position is higher priority...

Oha-you:
But what exactly is wrong with my proposed solution that follows the intuitive logic and doesn't require any explanations at all?

This page already has a sidebar block:
https://github.com/HDInnovations/UNIT3D/blob/8b88f4c8182eb3d3912ffef425c3224dcfd596f4/resources/views/Staff/automatic-torrent-freeleech/index.blade.php#L115-L118

However, sorting by position only is a bad approach whether you describe it or not.

HDVinnie:
I'm honestly not sure what's wrong with the proposed change. @Roardom would need to explain that. 

I'm trying to do best I can from my phone while I'm away on work trip. Until I return and can actually dig into code and test locally or on dev vps I'm doing best I can trying to review in app on phone without being to test which I don't really like but v10 release is a ways out so will have time to thoroughly test before actually merged to master and released.

Roardom:
My reasoning is that the proposed logic is "magic" and unexpected from people using the system. Sorting by position makes the most sense because it objectively prioritizes the freeleech depending on which is first. It could be renamed to "priority" if that makes more sense.

Magically guessing what people want is worse than having a defined priority. You'd need even more documentation to describe exactly what is being prioritized and when with this new logic, due to being unexpected.

Besides that, it would break existing freeleeches defined for sites and this PR doesn't include a migration to migrate existing records to the new format.

Oha-you:
It's fine, I don't need any of these PRs to be reviewed quick. You can take as much time as you want.
I only reminded about the donation fixes PR once after a month because it wasn't clear if I should do anything about it or not.

That being said, @Roardom definitely has a bias against me. Or maybe he doesn't like anyone besides him contributing to UNIT3D at all?
Don't worry, I'll stop being active here soon and move on to other projects.

Oha-you:
> Magically guessing what people want is worse than having a defined priority

People want exactly what's written in the sidebar: to apply freeleech automatically when a torrent is uploaded that meets the given criteria. Criteria being the size or the regex.
That's what people would get if you merged this PR instead of rejecting it.

> Besides that, it would break existing freeleeches defined for sites and this PR doesn't include a migration to migrate existing records to the new format

No migration is required because nothing will break. If anything, this feature will finally start working as intended for some.
You can just put it in the release notes that regexes for auto-freeleech will have a higher priority now. That's what they are intended for - to override the torrent size logic.

Roardom:
You've helpfully triggered two obvious things so far: donations and ai.

But some of these opinionated PRs I disagree with for other reasons and don't believe they provide benefit to UNIT3D.

PR #5361, opened 2026-03-24
(Add) Invite tree link for Invited by key in user profiles

Oha-you:
Invite Tree feature is great, so why not add a direct link to this page in user profiles.

Functionality on the third level of the navbar can easily be overlooked.
Plus a direct link helps the admins to investigate potential bad actors faster.

I know that Gazelle doesn't have such links on profile pages too, but this is something that can be improved upon.

Also I had to make Prettier ignore this line because it kept adding whitespace bugs and turning `(Tree)` into `( Tree )` (as mentioned in #5352). I didn't want parenthesis to be clickable.

### DIFF
--- a/resources/views/user/invite-tree/index.blade.php
+++ b/resources/views/user/invite-tree/index.blade.php
@@ -25,7 +25,7 @@
 
 @section('main')
     <section class="panelV2">
-        <h2 class="panel__heading">Invite tree</h2>
+        <h2 class="panel__heading">{{ __('user.invite-tree') }}</h2>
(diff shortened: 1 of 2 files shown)


HDVinnie (approved):

Roardom (changes requested):
I'd rather the "Invited by" key link to the tree instead. It's more relevant.

Oha-you:
Good call, I like it better this way too.
And no additional language strings needed.

Roardom (approved):

PR #5362, opened 2026-03-24
(Update) Restrict metadata updates only to mods, torrent mods and editors

Oha-you:
- I wonder what was the reasoning behind #2981

Who requested it? Why should every user be able to edit every torrent's metadata on the first day after being registered?
No other tracker platform allows this. Such perks might be available to power users, but not to the newbies.

This PR provides better defaults and restricts metadata update requests only to Moderators/Admins, Torrent Moderators and Editors.

It solves the following issues:
1. TMDB edits are moderated after the fact. It means you can vandalize hundreds of entries and then quickly synchronize these changes to the tracker before you get caught/banned on TMDB.
2. Metadata updates aren't auditable so it's hard to even notice the fact that some of your older torrents may have been vandalized by newly registered trolls. By restricting the access to staff and editors only you protect your tracker and simplify the task of finding the responsible person.
3. Even though global TMDB rate limiting was added in #4699, any newly registered user can create a huge queue for the tracker to process by requesting metadata updates for different torrents 2 times per second. And such user will never run out of IDs because you only need 57600 torrents uploaded 8 hours ago or earlier to be able to continuously generate metadata updates 24/7 before the cache expires (so you can reuse the same IDs again).

Uploaders will still able to trigger an update by editing their torrents within the allowed period of time:
https://github.com/HDInnovations/UNIT3D/blob/7cfcb3796f7aa5ac39f52e47514ca52603b94f9a/app/Http/Controllers/TorrentController.php#L375-L376
After this time has passed, their torrents should stay the same, including all the metadata. Only staff and editors should be able to change it.

I also made the "Try again" message more specific since you don't need to wait for a whole day:
https://github.com/HDInnovations/UNIT3D/blob/7cfcb3796f7aa5ac39f52e47514ca52603b94f9a/app/Jobs/ProcessMovieJob.php#L149-L151

### DIFF
--- a/app/Http/Controllers/SimilarTorrentController.php
+++ b/app/Http/Controllers/SimilarTorrentController.php
@@ -102,6 +102,13 @@ class SimilarTorrentController extends Controller
 
     public function update(Request $request, Category $category, int $metaId): \Illuminate\Http\RedirectResponse
     {
+        abort_unless(
+            $request->user()->group->is_modo ||
+            $request->user()->group->is_torrent_modo ||
+            $request->user()->group->is_editor,
+            403
+        );
+
         if (!($category->movie_meta || $category->tv_meta || $category->game_meta)) {
             return to_route('torrents.similar', ['category_id' => $category->id, 'tmdb' => $metaId])
                 ->withErrors('This meta type can not be updated.');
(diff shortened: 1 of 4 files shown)


Roardom:
I disagree with this change.

PR #5363, opened 2026-03-24
(Update) Set line-height for secondary navbar items, make the height 34px

Oha-you:
Just trying to make the scroll offset for anchor links in #5292 less magic number-y.

Before:

scroll-before

---

After:

scroll-after

---

Top navbar height is `48px`:
https://github.com/HDInnovations/UNIT3D/blob/7cfcb3796f7aa5ac39f52e47514ca52603b94f9a/resources/sass/layout/_top_nav.scss#L5

Anchor links are mostly used for forum posts that have a gap of `18px`:
https://github.com/HDInnovations/UNIT3D/blob/7cfcb3796f7aa5ac39f52e47514ca52603b94f9a/resources/sass/components/forum/_post.scss#L7

This PR makes the secondary nav height `34px` instead of `31.33px` so that the sum `48 + 18 + 34` is actually `100px`

### DIFF
--- a/resources/sass/layout/_secondary-nav.scss
+++ b/resources/sass/layout/_secondary-nav.scss
@@ -38,6 +38,7 @@
     color: var(--secondary-nav-tab-fg);
     background-color: var(--secondary-nav-tab-bg);
     font-size: 14px;
+    line-height: 24px;
     border: none;
     text-align: left;
     cursor: pointer;
@@ -148,6 +149,7 @@
     text-decoration: var(--breadcrumb-inactive-text-decoration);
     color: var(--breadcrumb-fg);
     cursor: pointer;
+    line-height: 24px;
 
     &:hover {
         color: var(--breadcrumb-hover-fg);


Roardom (changes requested):
The anchor links were originally intended for the wikis without forum topics in mind. If you need more room, feel free to increase the magic number.

Oha-you:
I've just checked some Wiki pages on BLU and I don't think the `2.67px` secondary nav height increase will affect them in any way.

While for the forum posts the end of a previous post is very visible and annoying without the change from this PR.

So I think it's still better to have at least some logic for those `100px` instead of having none.

I have another PR in the works that will add anchor links to comments and the same `18px` gap will be used there.

Roardom:
That doesn't make any sense because every screensize and stylesheet is different. If you actually want to fix the issue, you'll have to use JS.

Oha-you:
What does not make sense to you?
Do you not see the issue this PR fixes?
Do you not see that the top nav has a default height of `48px`, the standard gap is `18px`, but the secondary nav doesn't have any height set by default?
Do you not think that it's better to have at least some ground for magic numbers you come up with in CSS?

Alright, let me
> make sure we're **extra clear**

This is the issue:

offset
&nbsp;
offset2
&nbsp;

The solution is to have line-height of secondary navbar items set to `24px` because the remaining `10px` is added with padding:
https://github.com/HDInnovations/UNIT3D/blob/5f0bb9b16164580d77c7d6ae79e2377dd4ebb246/resources/sass/layout/_secondary-nav.scss#L33-L35
And margin:
https://github.com/HDInnovations/UNIT3D/blob/5f0bb9b16164580d77c7d6ae79e2377dd4ebb246/resources/sass/layout/_secondary-nav.scss#L118-L120

I know that the offset won't be relevant on smaller screens if the navbar height grows.
But at least it will be correct by default on most user's screens. 

> because every screensize and stylesheet is different

No, it's not. The stylesheet is defined in `.scss` files and generated by vite.
1px is 1px in every browser and on any OS.

You really need to
> Focus on the code and **stop embarrassing yourself**

© https://github.com/HDInnovations/UNIT3D/pull/5282#issuecomment-4102182820

Calling @HDVinnie here too. Look at this circus of fighting for `2.67px` height increase.

HDVinnie:
I'll check when back at room but for now I think the key takeaway is to remember to respect each other even if disagreements on topics. @Roardom is a core dev of UNIT3D along with myself. We can all discuss this peacefully and not word things in a bashful way. @Oha-you keep in mind we don't get paid. There's no salary, payroll or benefits at all. This is FOSS and a hobby. Even I have had my pull requests opposed and in the end I accept it, rethink it and redo or move on to another feature. In the end it's all done respectfully though which is key takeaway here.

PR #5365, opened 2026-03-25
(Fix) Make torrent file cleaning migration from 2023 work again

Oha-you:
- Initially added in #2598
- Hasn't been updated in #4497

The need to clean torrent files hasn't been mentioned anywhere in the release notes.

The fact that this migration has been broken since #4497 and that `announce-list` stopped being removed from torrent files since #2598 lead to a situation of real passkeys **leaking** to users who downloaded certain torrents.

After I understood the problem and ran the clean command manually, the bug in #5364 duplicated all files and threw them into a distant path.

Very cool, very nice. Lovely, even.

### DIFF
--- a/database/migrations/2023_02_09_113903_clean_torrent_files.php
+++ b/database/migrations/2023_02_09_113903_clean_torrent_files.php
@@ -14,9 +14,8 @@ declare(strict_types=1);
  * @license    https://www.gnu.org/licenses/agpl-3.0.en.html/ GNU Affero General Public License v3.0
  */
 
-use App\Helpers\Bencode;
 use Illuminate\Database\Migrations\Migration;
-use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Artisan;
 
 return new class () extends Migration {
     /**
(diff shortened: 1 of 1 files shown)


Roardom:
I'm probably missing something here but #2598 where the torrent files are cleaned is from 2023-02-09 and #4497 where the torrents are moved elsewhere is from 2025-02-23. So the torrent files would've been cleaned before being moved. I don't see how this migration was broken since migrations are always executed in order, and the change in #4497 didn't edit #2598's migration.

Also, migrations should never rely on external code in case that external code changes. This is a good example because anyone running this migration isn't yet storing files in the location that the command uses. What exactly is wrong with this migration?

Oha-you:
You don't take into account that not every tracker rushes to update to the newest UNIT3D version as soon as you release it.
Some trackers don't need all the latest blows and whistles, so they gave up and freezed their codebase or forked it and build upon (like HUNO).

The fact that you haven't updated this migration when you did the refactoring in #4497 opens a vulnerability for everyone who still uses any version prior to [v7.0.0](https://github.com/HDInnovations/UNIT3D/releases/tag/v7.0.0) now.

This is what happened in my case. UNIT3D stored uploader's passkeys intact in torrent files before. But removed the sensitive keys when it sent torrent file to users.
You stopped removing sensitive keys, didn't mention it in `v7.0.0` release notes, then broke the migration.

> migrations should never rely on external code in case that external code changes

You also shouldn't use migrations to process potentially hundreds of thousands of files.
If it touches anything outside the database, it does not belong in a migration.

But you did it anyway. So the fix I provided is a perfect example of an exception where you **should** call the artisan command and don't duplicate the code. So that no other tracker's passkeys are leaked again.

You will 100% do 10 more refactorings in the future and break the torrent clean command again.
So let's not come up with excuses and accept this objectively best solution to the problem.

Oha-you:
> and #4497 where the torrents are moved elsewhere

Torrents are not moved anywhere automatically.
It is done manually by coders who install the latest version, notice that public files location has changed, move the torrents and only then run the migrations.

No one expects migrations to touch any files, only the database.
You implemented a non-standard file processing migration so now you have to make sure it works after any refactoring.

HDVinnie:
> > and #4497 where the torrents are moved elsewhere
> 
> 
> 
> Torrents are not moved anywhere automatically.
> 
> It is done manually by coders who install the latest version, notice that public files location has changed, move the torrents and only then run the migrations.
> 
> 
> 
> No one expects migrations to touch any files, only the database.
> 
> You implemented a non-standard file processing migration so now you have to make sure it works after any refactoring.

Yes they are moved and the migration is fine. 

If owners update properly all is fine.

HDVinnie:
> The fact that you haven't updated this migration when you did the refactoring in #4497 opens a vulnerability for everyone who still uses any version prior to v7.0.0 now.


And that's there problem. We have a security md for a reason and v7 is long EOL

Roardom:
> You don't take into account that not every tracker rushes to update to the newest UNIT3D version as soon as you release it.

The migration is included in the same commit, so it doesn't matter what version they're on. The data will be cleaned via the migration as soon as the site is updated to that commit.

PR #5368, opened 2026-03-25
(Refactor) TorrentController API, use internals_exists for Featured

Oha-you:
A light refactor that also fixes a discrepancy in TorrentController API for Featured torrents after:
- #4370

Logic of who should be considered Internal was changed here:
https://github.com/HDInnovations/UNIT3D/blob/5f0bb9b16164580d77c7d6ae79e2377dd4ebb246/app/Http/Controllers/TorrentBuffController.php#L136-L140

But it was missed in `API/TorrentController.php` itself.


Roardom (changes requested):
Can you implement the fix without the refactor please?

Also, CI is not passing.

Oha-you:
That's it.
Good luck with your shitty and Indian-style code.
Also CI is broken on your side.

PR #5370, opened 2026-03-25
(Add) Focus the textarea when replying or editing a comment

Oha-you:
Right now textarea for Reply and Edit buttons is just awkwardly added to the DOM.

Unlike the Quote button that actually sets the focus:
https://github.com/HDInnovations/UNIT3D/blob/8b88f4c8182eb3d3912ffef425c3224dcfd596f4/resources/views/livewire/comment.blade.php#L45-L46

### DIFF
--- a/resources/views/livewire/comment.blade.php
+++ b/resources/views/livewire/comment.blade.php
@@ -89,9 +89,14 @@
             @endif
         </aside>
         @if ($isEditing)
-            <form wire:submit="editComment" class="form edit-comment">
+            <form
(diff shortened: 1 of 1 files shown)


Roardom (commented):
Is this syntax compatible with alpine's csp mode?

https://alpinejs.dev/advanced/csp#whats-not-supported says arrow functions aren't compatible but I'm not in a position to run your code right now to test.

We're migrating to CSP mode with only chatbox left to refactor. We don't want to move backwards

PR #5372, opened 2026-03-26
(Fix) Validation of non-square gif avatars/icons

Roardom:


Roardom:
We depend on php image intervention v2, which is end of life. v3 adds support for resizing animated gifs. But instead of depending on it ourselves, we can wait for https://www.github.com/laravel/framework/pull/59276 to be merged and update laravel to it.

PR #5384, opened 2026-04-09
(Add) Trigger navigation on Enter key for quick search

aulterego:
Press Enter in the quick search input to navigate directly to the first search result.


Roardom (approved):

Roardom (commented):
Actually, on second thought. Can you add an "else" that redirects you to `/torrents?name=${encodeURIComponent(this.searchText)}`?

aulterego:
> Actually, on second thought. Can you add an "else" that redirects you to `/torrents?name=${encodeURIComponent(this.searchText)}`?

✅

Roardom (approved):
Thanks!

Looks like CI is failing, but it's unrelated to your changes. Let me see if I can get that working again and then it can be merged.

Issue #3970, opened 2024-07-07
Pruning Category from the dashboard error 500[Bug]


BitPyrate:
### UNIT3D Version

v8.1.3

### PHP & Platform

8.3.8      8.0.37-0ubuntu0.24.04.1

### Laravel verion

11.5.0

### Have you done this?

- [X] I am willing to share my stack trace and logs
- [ ] I can suggest a fix as a Pull Request

### Expectation

removing a category from the dashboard i would expect it to remove all references.

### Description

With how i  personally solved it. however I suspect the bug needs to be fixed in pull request.

did some basic renaming and pruning unused categories using the web ui dashboard. 

The issue that arouse was that someone had a request that had a category id of one that was pruned.  This went unnoticed until someone went to check some requests. This caused a 500 error

so it appears the web interface may not doing a proper clean up when you remove an entry.

laravel log snippet:
Attempt to read property "name" on null (View: /var/www/html/resources/views/livewire/torrent-request-search.blade.php) (View: /var/www/html/resources/views/livewire/torrent-request-search.blade.php) {"userId":3,"exception":"[object] (Illuminate\View\ViewException(code: 0): Attempt to read property "name" on null (View: /var/www/html/resources/views/livewire/torrent-request-search.blade.php) (View: /var/www/html/resources/views/livewire/torrent-request-search.blade.php) at /var/www/html/storage/framework/views/7b59ca7d17fb079cac0a087781a4a783.php:79)
[stacktrace]

the offending entry was found using the following:
mysql> SELECT r.id, r.category_id 
       FROM requests r 
       LEFT JOIN categories c ON r.category_id = c.id 
       WHERE c.id IS NULL;

in mycase yielded:
+----+-------------+
| id | category_id |
+----+-------------+
| 34 |          26 |
+----+-------------+
1 row in set (0.03 sec)

that is one of the categories that was removed using the webui admin panel.

i updated the field manually to a known good category:

mysql> UPDATE requests 
    -> SET category_id = 11 
    -> WHERE id = 34;

and then clear the cache. Everything running smoothly again.

I suspect when you delete a category it does not look for any existing requests that use that category.  perhaps it should and halt (preventing the removal of e the category until the requests are updated) or perhaps prompting what to set the stranglers to. 

Hope this is helpful, since my personal issue is fixed. feel free to close this if this is not helpful

Roardom:
[...]

PR #5403, opened 2026-04-28
Handle missing request categories

AgentGoose32:
/claim #3970

## Summary
- Adds a default category model for torrent requests whose referenced category has been deleted.
- Prevents the request search view from throwing when it renders a request with a stale `category_id`.
- Adds a regression test covering the deleted-category fallback.

## Validation
- `php -l app/Models/TorrentRequest.php`
- `php -l tests/Unit/Models/TorrentRequestTest.php`
- `git diff --check`

Note: I also attempted the targeted Pest test. It reached the test bootstrap but local validation is blocked by this checkout's environment/config: missing default app env values first, then MySQL unavailable; with sqlite memory it hits an existing duplicate index migration issue (`request_id already exists`).

### DIFF
--- a/app/Models/TorrentRequest.php
+++ b/app/Models/TorrentRequest.php
@@ -148,7 +148,9 @@ final class TorrentRequest extends Model
      */
     public function category(): BelongsTo
     {
-        return $this->belongsTo(Category::class);
+        return $this->belongsTo(Category::class)->withDefault([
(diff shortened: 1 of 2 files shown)


Roardom (changes requested):
To fix this bug, please follow the logic I described in the linked issue. Adding default attributes is not the correct solution.

Issue #3365, opened 2024-01-09
[Bug] Avatar does not update, due to format requirement


BitPyrate:
### UNIT3D Version

v7.2.0

### PHP & Platform

8.3.1 8.0.35-0ubuntu0.22.04.1

### Laravel verion

10.37.3

### Have you done this?

- [X] I am willing to share my stack trace and logs
- [ ] I can suggest a fix as a Pull Request

### Expectation

avator to update

### Description

adding avatars seems to be working.
however, users complained they could not change their avatar once set.  So I tried myself with the same result.

We are getting the message "Your Account was updated successfully " but the avatar doesn't change

at first i thought it was a permissions issue
permission of an example avatar file in the public/figures/img
-rw-r--r-- 1 www-data www-data   7863 Jan  9 06:17 username.jpg

browsing through github i saw an issue when changing formats in a recent changelog, but i didn't see any mention of the issue I am seeing. as the file we are replacing are the same format "jpg"

what i saw was for UserController.php
 https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/3107/files/769fb6213fbc25f77d18a806c148d9a36048e81a#diff-cf777dea29b43cdee7e97016b5957f1b60a62e3e2427aa08c46e205017582254

I think i found an issue with the logic that was in pull request 3107...... its requiring the format to change.

So i tested again with a png... and bam it changed.
try with another png..... failure.
try a jpg..... it switches back.

so i think the logic needs to be re-examined.

### Are you a UNIT3D sponsor?

Yes, with my username

HDVinnie:
The whole storage system is being refactored and this will be sorted in next release. Leaving open until PR is opened, reviewed and merged.

twadepsvita:
I've been having this problem in a site I'm trying to set up. It's even more annoying when the site doesn't seem to allow for images of any other image ratio, so I'm having to replace the cropped images with ones in templates that allow them to show in full. Is it possible for torrent images to be fixed and to have any ratio when fixing this bug?

Roardom:
This was fixed (as good as it will get) in #4497.

Format requirement is irrelevant. It's because the url changes from .jpg to .png, and the client caches each url individually (and indefinitely). With the new linked PR, the cache is limited to 2 hours max before they see the new avatar. Users can also force refresh their browser (ctrl+shift+r).

[...]

PR #5404, opened 2026-04-28
Prevent stale cached user images

AgentGoose32:
/claim #3365

## Summary
- Stops authenticated user avatar and icon responses from being cached for two hours.
- Keeps the existing longer cache policy for stable authenticated images, while user-uploaded images now revalidate immediately after replacement.
- This lets same-extension avatar uploads show the updated image instead of a stale browser-cached copy.

## Validation
- `php -l app/Http/Controllers/AuthenticatedImageController.php`
- `git diff --check`


Roardom (changes requested):
This breaks caching user avatars and user icons. To fix this issue, you should add a randomized query parameter, similar to how js and css is deployed.

PR #5310, opened 2026-03-08
fix: add wire:navigate to forum pagination for standard URL navigation

sebmuehlbauer:
Fixes #4082

The bug occurred because Livewire pagination links in the forum were being treated as AJAX-only triggers, causing issues when users tried to open them in new tabs or middle-click. By adding `wire:navigate` to the pagination links (or ensuring the pagination component supports link-based navigation), we allow standard browser behavior for navigation while keeping Livewire SPA benefits. 

Note: This change ensures that even when navigating, the browser sees a valid URL, preventing the 405 Method Not Allowed error when opening pages in a new tab.

---
*This PR was autonomously generated by [Pangea 3](https://github.com/sebmuehlbauer) — an AI-powered development system.*

Issue #4082, opened 2024-08-25
[Bug] 405 GET request instead of POST request when navigating forum pages


Audionut:
### UNIT3D Version

Bleeding edge

### PHP & Platform

8.3.10

### Laravel verion

11.19.0

### Have you done this?

- [X] I am willing to share my stack trace and logs
- [ ] I can suggest a fix as a Pull Request

### Expectation

Navigating forum pages via new tabs doesn't error

### Description

Navigate to a forum thread page using the page markers, right click a page marker and open in a new tab.

It works fine when going to a thread via a notification for instance, but as soon as you use a page marker to navigate to some other page of the thread, it errors.

I noticed also that it (now?) shows this as the link ```/livewire/update?page=151``` instead of the actual forum page link displayed in the url bar.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/HDInnovations) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/HDInnovations/UNIT3D-Community-Edition/issues/4082">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/4082/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/HDInnovations/UNIT3D-Community-Edition/issues/4082/pledge.svg">
</picture>
</a>

Roardom:
Livewire pagination currently relies on js, using left click only. You probably either middle clicked or your js is disabled.

PR #5406, opened 2026-04-28
Use plain links for forum post pagination

AgentGoose32:
/claim #4082

## Summary
- Uses plain browser links for forum topic post pagination instead of Livewire AJAX-only pagination controls.
- Keeps the existing Livewire pagination partial unchanged for other interactive search tables.
- Allows forum page links to be opened in a new tab/window with the canonical `?page=` URL instead of hitting `/livewire/update` with GET.

## Validation
- `php -l resources/views/partials/pagination-links.blade.php`
- `php -l resources/views/livewire/topic-post-search.blade.php`
- `git diff --check`

Note: I saw the prior closed PR #5310 used `wire:navigate`; this PR takes the safer plain-anchor approach for forum post pages so browser open-in-new-tab behavior does not depend on Livewire interception.

### DIFF
--- a/resources/views/livewire/topic-post-search.blade.php
+++ b/resources/views/livewire/topic-post-search.blade.php
@@ -54,7 +54,7 @@
         </ul>
     @endif
 
-    {{ $posts->links('partials.pagination') }}
+    {{ $posts->links('partials.pagination-links') }}
     <div class="panel__body">
         @if ($posts->count() > 0)
             <ol class="topic-posts">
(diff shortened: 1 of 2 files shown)


Roardom (changes requested):
Thanks for the effort, but I would rather not duplicate the pagination logic into another file. Can you please keep using the same `partials.pagination` partial? For what it's worth, all pagination is affected, it's not just the forum pagination.

Issue #4298, opened 2024-11-01
[Request] Swap `$model->update([..., 'updated_at' => DB::raw('updated_at')])` for  `Model::withoutTimestamps(fn () => $model->update([...]))`


Roardom:
Probably should refactor all of these at some point to use the idiomatic syntax.

IonBazan:
We can also do `updateQuietly` - this however prevents all model events/observers from being triggered (including `Auditable`).

Roardom:
> We can also do `updateQuietly` - this however prevents all model events/observers from being triggered (including `Auditable`).

Thanks for the alternative. In most cases that this hack is used, it's for bulk updates that don't send model events/observers anyways, so that shouldn't be an issue. However, if the goal is to not update the `updated_at` timestamp, then it's probably better to call a function named `withoutTimestamps`. We also generally try and avoid relying on too many model events/observers so that we can have more freedom to optimize our database queries.

PR #5408, opened 2026-04-28
Use withoutTimestamps for bulk updates

AgentGoose32:
/claim #4298

## Summary
- Replaces several `updated_at => DB::raw('updated_at')` bulk-update hacks with `Model::withoutTimestamps(...)`.
- Keeps the intended behavior of changing the target columns without bumping `updated_at`.
- Covers peer/history cleanup, hit-and-run marking, and high-speed torrent tagging paths.
- Removes now-unused `DB` imports where the raw timestamp workaround was the only usage.

## Validation
- `php -l app/Http/Controllers/User/PeerController.php`
- `php -l app/Http/Controllers/Staff/FlushController.php`
- `php -l app/Console/Commands/AutoHighspeedTag.php`
- `php -l app/Console/Commands/AutoCorrectHistory.php`
- `php -l app/Console/Commands/AutoWarning.php`
- `./vendor/bin/pint app/Http/Controllers/User/PeerController.php app/Http/Controllers/Staff/FlushController.php app/Console/Commands/AutoHighspeedTag.php app/Console/Commands/AutoCorrectHistory.php app/Console/Commands/AutoWarning.php`
- `git diff --check`

### DIFF
--- a/app/Console/Commands/AutoCorrectHistory.php
+++ b/app/Console/Commands/AutoCorrectHistory.php
@@ -20,7 +20,6 @@ use App\Models\History;
 use Illuminate\Console\Command;
 use Illuminate\Support\Carbon;
 use Exception;
-use Illuminate\Support\Facades\DB;
 use Throwable;
 
 class AutoCorrectHistory extends Command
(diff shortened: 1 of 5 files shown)


Roardom (changes requested):
I appreciate the effort, but there's quite a few instances left. E.g. in AutoFlushPeers.php or AutoPreWarning. Is there a reason you didn't change those?

Issue #3664, opened 2024-03-13
[Request] Log failed invites to emails already in use


Audionut:
The user receives a dialog box of the error, please also log for staff.

PR #5416, opened 2026-04-29
Log failed duplicate invite attempts

AgentGoose32:
## Summary
- log failed invite attempts when the target email already belongs to a user or pending application
- expose failed-at / failure-reason fields in the staff invite log
- keep failed attempts out of normal sent-invite counts and allow future successful invites to use emails that only exist in failed log rows

## Validation
- `php -l app/Http/Controllers/User/InviteController.php`
- `php -l app/Models/User.php`
- `php -l resources/views/livewire/invite-log-search.blade.php`
- `php -l database/migrations/2026_04_29_002400_add_failed_invite_logging.php`
- `php -l tests/Feature/Http/Controllers/User/InviteControllerTest.php`
- `./vendor/bin/pint app/Http/Controllers/User/InviteController.php app/Models/User.php resources/views/livewire/invite-log-search.blade.php database/migrations/2026_04_29_002400_add_failed_invite_logging.php tests/Feature/Http/Controllers/User/InviteControllerTest.php`
- `git diff --check`

Note: targeted Laravel test execution is blocked locally by environment services/config in this checkout. MySQL is not running for the default test connection; SQLite in-memory migration also fails on an existing duplicate index name in an older migration before this test runs.

/claim #3664

### DIFF
--- a/app/Http/Controllers/User/InviteController.php
+++ b/app/Http/Controllers/User/InviteController.php
@@ -18,11 +18,13 @@ namespace App\Http\Controllers\User;
 
 use App\Http\Controllers\Controller;
 use App\Mail\InviteUser;
+use App\Models\Application;
 use App\Models\Invite;
 use App\Models\User;
 use App\Rules\EmailBlacklist;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Mail;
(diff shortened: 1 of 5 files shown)


Roardom:
dupe of #5401?

Issue #4043, opened 2024-08-12
[Request] Global Database Settings


HDVinnie:
Move all UNIT3D configs to database with caching so one does not have to edit code to make changes.

PR #5417, opened 2026-04-29
Add database-backed global settings

AgentGoose32:
Refs #4043

Adds a database-backed global settings foundation with cache invalidation.

Changes:
- Adds `global_settings` table with unique `key` and nullable JSON `value`
- Adds `App\Models\GlobalSetting`
- Provides `GlobalSetting::value($key, $default)` using `cache()->rememberForever`
- Provides `GlobalSetting::put($key, $value)` with database upsert and targeted cache invalidation

This is a small first slice toward moving code/config-managed UNIT3D settings into database-backed cached settings.

Validation:
- `php -l app/Models/GlobalSetting.php`
- `php -l database/migrations/2026_04_29_095500_create_global_settings_table.php`
- `git diff --check`

### DIFF
--- a/app/Models/GlobalSetting.php
+++ b/app/Models/GlobalSetting.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+/**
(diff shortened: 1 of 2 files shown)


Roardom:
Is this still a draft?

Issue #3661, opened 2024-03-13
[Request] Links redirect specific to the user clicking the link


Audionut:
For example, in wikis. Having a link to '/users/username/passkeys' would redirect to the users passkey page for the user that clicks the link.

braingremlin85:
+1

this would be very useful when writing wikis

PR #5418, opened 2026-04-29
Add current-user wiki placeholders

AgentGoose32:
Refs #3661

Adds current-user placeholders for wiki content so links can resolve to the user who clicks/views the page.

Example wiki link:

`/users/{username}/passkeys`

When rendered for a logged-in user, `{username}` is replaced with the authenticated user's username.

Changes:
- Replaces `{user}`, `{username}`, and `{user_id}` in wiki content before BBCode/Markdown rendering
- Leaves content unchanged for guests
- Keeps the existing rendering pipeline intact

Validation:
- `php -l app/Models/Wiki.php`
- `git diff --check`


Roardom (changes requested):
Please use the method I described in the issue.

Issue #3662, opened 2024-03-13
[Request] 2fa enabled log


Audionut:
Displays a log of when a user has turned on or off their 2fa.

PR #5419, opened 2026-04-29
Log two-factor authentication changes

AgentGoose32:
Refs #3662

Logs two-factor authentication state changes for staff review.

Changes:
- Writes an audit entry when a user confirms/enables 2FA
- Writes an audit entry when a user disables 2FA
- Stores the action as `two_factor_enabled` or `two_factor_disabled`
- Attaches the audit entry to the affected user
- Stores a small JSON record with the resulting 2FA state

Validation:
- `php -l app/Http/Livewire/TwoFactorAuthForm.php`
- `git diff --check`

### DIFF
--- a/app/Http/Livewire/TwoFactorAuthForm.php
+++ b/app/Http/Livewire/TwoFactorAuthForm.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
 
 namespace App\Http\Livewire;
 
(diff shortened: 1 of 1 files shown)


Roardom (changes requested):
This should use the same logic as e.g. the apikey/passkey/rsskey updates.

PR #5472, opened 2026-05-22
(Update) Pint and php cs fixer dependencies

Roardom:
Fixes some formatting bugs I'm running into like removing `null` from `return null` inside of nullable property hooks.


Roardom:
Looks like the large diff is due to https://www.github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/9465

PR #5475, opened 2026-05-27
Fix: Set category_id to null on category deletion

whyujjwal:
This PR fixes a bug where deleting a category would cause a 500 error if there were any requests associated with that category. This was because the requests table was not being updated when a category was deleted, leaving orphaned requests that pointed to a non-existent category_id.

This PR fixes the issue by updating the requests table when a category is deleted, setting the category_id to null for all requests that have the deleted category_id.

Just following up to see if you've had a chance to review this. Let me know if you have any questions or feedback!

### DIFF
--- a/app/Http/Controllers/Staff/CategoryController.php
+++ b/app/Http/Controllers/Staff/CategoryController.php
@@ -124,8 +124,8 @@ class CategoryController extends Controller
      */
     public function destroy(Category $category): \Illuminate\Http\RedirectResponse
     {
-        // Set the category_id to null for all requests that have this category_id
-        $category->requests()->update(['category_id' => null]);
+        // Delete all requests that have this category_id
+        $category->requests()->delete();
 
         $category->delete();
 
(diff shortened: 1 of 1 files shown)


Roardom (commented):
`requests.category_id` should never be nullable.

https://github.com/HDInnovations/UNIT3D/blob/8b88f4c8182eb3d3912ffef425c3224dcfd596f4/database/schema/mysql-schema.sql#L1427

whyujjwal:
Thanks for the feedback! I've updated the PR to delete the requests associated with the deleted category. Let me know what you think.

whyujjwal:
Thanks for the feedback! I've updated the PR to delete the requests associated with the deleted category. Let me know what you think.

Roardom:
I think that would be unexpected behavior to delete requests when the category is deleted.

I would suggest doing something similar to how deleting a distributor or region is handled for the torrent table's foreign keys.

whyujjwal:
Hi! Thank you again for the approval on this PR. Just wanted to check in and see if there's anything else needed from my side to help get this merged. Happy to help resolve any conflicts if they've come up. Thanks!

Roardom:
Hi, I think you may have missed my previous comment. I have not approved it yet.

And just a heads up, polar shut down paid bug bounties in January 2025 with all the bounties being refunded.

PR #5480, opened 2026-05-30
feat: add public user profile API endpoint

Pippo-Baudo:
Adds GET /api/users/{username} endpoint that returns public user stats (uploaded, downloaded, ratio, buffer, seeding, leeching, seedbonus, hit_and_runs) while respecting the user's privacy settings. Hidden profiles return 403, private fields return "private".


Pippo-Baudo:
PR error.

PR #5507, opened 2026-07-04
(Fix) Don't send notifications to banned/disabled users

### LINKED ISSUE #4869: [Bug] Dont send notifications to banned/pruned users
Currently the commands sends emails to banned/pruned users.

skavinski-lab:
Fixes #4869

The scheduled notification commands select users without checking account status, so banned and disabled users (who can't access the site) still get emailed: auto:warning, auto:prewarning, auto:deactivate_warning, auto:remove_expired_donors and auto:remove_personal_freeleech.

This skips the notification for validating/banned/disabled/pruned accounts (via UserGroup::inactiveIds()) while still performing the underlying state changes. For the hit-and-run commands that means a banned user still gets the warning row created and prewarned_at set; only the email is suppressed, so the warning history stays accurate if the account is later restored. Happy to exclude those accounts from the selection entirely instead if you'd prefer. AutoDeactivateWarning also gets a null guard, since a pruned (soft-deleted) user's warning->user resolves to null.

Tested with Pint, Larastan and PHPUnit, with added coverage for the group-exclusion and soft-deleted paths.

### DIFF
--- a/app/Console/Commands/AutoDeactivateWarning.php
+++ b/app/Console/Commands/AutoDeactivateWarning.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
 
 namespace App\Console\Commands;
 
+use App\Models\Group;
 use App\Models\Warning;
 use App\Notifications\UserWarningExpired;
 use App\Services\Unit3dAnnounce;
@@ -69,8 +70,17 @@ class AutoDeactivateWarning extends Command
                 $usersWithExpiredWarnings[$warning->user_id] = $warning->user;
             }, 100);
 
+        $inactiveGroupIds = Group::query()
+            ->whereIn('slug', ['banned', 'validating', 'disabled', 'pruned'])
+            ->pluck('id')
+            ->all();
+
         // Send a single notification for each user with expired warnings
         foreach ($usersWithExpiredWarnings as $user) {
+            if (\in_array($user->group_id, $inactiveGroupIds, true)) {
+                continue;
+            }
+
             $user->notify(new UserWarningExpired($user));
         }
 
--- a/app/Console/Commands/AutoRemovePersonalFreeleech.php
+++ b/app/Console/Commands/AutoRemovePersonalFreeleech.php
@@ -50,9 +50,14 @@ class AutoRemovePersonalFreeleech extends Command
     {
         $personalFreeleech = PersonalFreeleech::query()->where('created_at', '<', now()->subDays(1))->get();
 
-        foreach ($personalFreeleech as $pfl) {
-            Notification::send(new User(['id' => $pfl->user_id]), new PersonalFreeleechDeleted());
+        $notifiableUsers = User::query()
+            ->whereIntegerInRaw('id', $personalFreeleech->pluck('user_id'))
+            ->whereDoesntHave('group', fn ($query) => $query->whereIn('slug', ['banned', 'validating', 'disabled', 'pruned']))
+            ->get();
 
+        Notification::send($notifiableUsers, new PersonalFreeleechDeleted());
+
+        foreach ($personalFreeleech as $pfl) {
             // Delete The Record From DB
             $pfl->delete();
 
--- a/app/Console/Commands/AutoPreWarning.php
+++ b/app/Console/Commands/AutoPreWarning.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
 
 namespace App\Console\Commands;
 
+use App\Models\Group;
 use App\Models\History;
 use App\Notifications\UserPreWarning;
 use Illuminate\Console\Command;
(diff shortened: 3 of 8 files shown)


Roardom (changes requested):
You can't rely on group IDs, only slugs from system groups

skavinski-lab:
Good catch, thanks. Reworked in 89b118d: the commands now resolve the banned/validating/disabled/pruned group IDs by slug at runtime (same slug list used elsewhere, e.g. StatsController), and the `inactiveIds()` addition to the `UserGroup` enum is reverted. Tests still pass (enum values remain valid there since tests seed `GroupSeeder`).

app/Console/Commands/AutoDeactivateWarning.php
```diff
@@ -69,8 +70,17 @@ final public function handle(): void
...
             }, 100);
 
+        $inactiveGroupIds = Group::query()
+            ->whereIn('slug', ['banned', 'validating', 'disabled', 'pruned'])
+            ->pluck('id')
+            ->all();
+
         // Send a single notification for each user with expired warnings
         foreach ($usersWithExpiredWarnings as $user) {
+            if ($user === null || \in_array($user->group_id, $inactiveGroupIds, true)) {
```
Roardom: Shouldn't need to compare to null here. The relation is always defined.

app/Console/Commands/AutoRemovePersonalFreeleech.php
```diff
@@ -50,8 +50,20 @@ final public function handle(): void
     {
         $personalFreeleech = PersonalFreeleech::query()->where('created_at', '<', now()->subDays(1))->get();
 
+        // Notify active users only; freeleech is still removed below. Pruned users are
+        // soft-deleted so they fall out of the default scope here.
+        $notifiableUserIds = User::query()
+            ->whereIntegerInRaw('id', $personalFreeleech->pluck('user_id'))
+            ->whereDoesntHave('group', function ($query): void {
+                $query->whereIn('slug', ['banned', 'validating', 'disabled', 'pruned']);
+            })
```
Roardom: Please use an arrow function here

app/Console/Commands/AutoRemovePersonalFreeleech.php
Roardom, line 1: You might as well eager load the users and send them in the notification since you're already fetching them in the group query

Roardom (changes requested):
Thanks! Just a few suggestions.

app/Console/Commands/AutoDeactivateWarning.php
skavinski-lab: Removed in e803e36. You're right, with `withDefault` on the relation it can never be null.

app/Console/Commands/AutoRemovePersonalFreeleech.php
skavinski-lab: Done in e803e36.

app/Console/Commands/AutoRemovePersonalFreeleech.php
skavinski-lab: Done in e803e36: the query now fetches the User models and passes them straight to `Notification::send()`, so the per-row id membership check is gone. This needed one phpstan-baseline entry for the arrow function's return type (same `return.type` entry the identical closures in StatsController have).

app/Console/Commands/AutoRemovePersonalFreeleech.php
```diff
@@ -50,9 +50,16 @@ final public function handle(): void
     {
         $personalFreeleech = PersonalFreeleech::query()->where('created_at', '<', now()->subDays(1))->get();
 
-        foreach ($personalFreeleech as $pfl) {
-            Notification::send(new User(['id' => $pfl->user_id]), new PersonalFreeleechDeleted());
+        // Notify active users only; freeleech is still removed below. Pruned users are
+        // soft-deleted so they fall out of the default scope here.
```
Roardom: This comment isn't really necessary. If anything the first 4 words are useful, but I'd just remove the whole thing. It's obvious the freeleech is removed below, and this isn't the spot to document the fact that a user in a pruned group and a soft deleted user aren't mandated to be the same.

Roardom (changes requested):
Thanks! Sorry, one last thing. Also, if you don't mind squashing your commits afterwards with a commit message/description reflecting the original intended change, that'd be great.

app/Console/Commands/AutoRemovePersonalFreeleech.php
skavinski-lab: Removed in 4d6a1f07f.

Roardom:
Hi, this still requires a rebase to a single commit. I can't squash it via the github ui because it doesn't allow me to choose the commit name to match our commit naming guidelines.

skavinski-lab:
Rebased onto current `development` and squashed to a single commit (e80cdca18) with the `(Fix)` title and a description reflecting the final change. Diff is unchanged from the last reviewed state (8 files, +122/-5). Ready for another look, thanks.

Note: the "Type Coverage (Pest)" check can fail with a parse error in `vendor/pestphp/pest-plugin-type-coverage/.temp/v3.php`, not in PR code; the same error hit `development` at 427a69be1 (run 32206315118) and passed on re-run. I don't have permission to re-run it here.

Roardom (approved):

Roardom:
Wait, you did it backwards this time :|. Now they both have the parentheses.

skavinski-lab:
Sorry about that. Commit title is now `fix: don't send notifications to banned/disabled users` (1c01808cc), PR title keeps the `(Fix)` form. Content unchanged.

Roardom:
Thank you!

PR #5508, opened 2026-07-04
(Fix) Use torrent_id in WarningFactory

skavinski-lab:
The `warnings.torrent` column was renamed to `torrent_id` in migration `2025_11_08_094209_rename_warnings_torrent_to_torrent_id`, but `WarningFactory` still populated the old `torrent` key.

As a result `Warning::factory()->create()` throws:

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'torrent' in 'field list'
```

which is why the tests relying on it are currently marked `markTestIncomplete`. This one-line change aligns the factory with the renamed column so the factory (and the tests built on it) work again.

### DIFF
--- a/database/factories/WarningFactory.php
+++ b/database/factories/WarningFactory.php
@@ -39,7 +39,7 @@ class WarningFactory extends Factory
         return [
             'user_id'    => User::factory(),
             'warned_by'  => User::factory(),
-            'torrent'    => Torrent::factory(),
+            'torrent_id' => Torrent::factory(),
             'reason'     => $this->faker->text(),
             'expires_on' => $this->faker->dateTime(),
             'active'     => $this->faker->boolean(),


Roardom:
> which is why the tests relying on it are currently marked markTestIncomplete. This one-line change aligns the factory with the renamed column so the factory (and the tests built on it) work again.

Is there anything still missing from the test that would prevent it from being enabled?

Roardom:
@skavinski-lab Do you plan on coming back to this?

skavinski-lab:
Sorry for the delay! Good question: it turned out removing `markTestIncomplete` wasn't enough on its own, so I've enabled all four tests the factory was blocking (`destroy` and `update`, success + 403 each). Beyond un-marking them, the scaffolds needed:

- an acting user in an `is_modo` group (`Group::factory()->owner()`), since every action in this controller is behind `abort_unless(...->group->is_modo, 403)`
- the warning tied to the route `{user}`, since the `users/{user:username}` group uses `scopeBindings()`
- redirect + soft-delete assertions instead of `assertOk()` / `assertModelMissing()` (the controller redirects to `users.show` and `Warning` uses `SoftDeletes`)
- `Notification::fake()` for the `WarningTorrentDeleted` notification `destroy()` sends

As a sanity check: with the factory change reverted, all four fail with `SQLSTATE[42S22]: Unknown column 'torrent'`; with it they pass. Also rebased onto current development.

The two `mass destroy` tests are still marked incomplete for a different reason: `DELETE users/{user}/warnings/mass-delete` is registered after `DELETE users/{user}/warnings/{warning}`, so the wildcard shadows it and the request 404s (verified against the router; that URI resolves to `WarningController@destroy`). I can move the static route above the wildcard here or in a follow-up, whichever you prefer.

While digging I also noticed nothing links to the `users.warnings.*` routes anymore (the profile UI goes through the `UserWarnings` Livewire component), and `User\WarningController::store()` still inserts the old `torrent` column, which would throw an unknown-column QueryException if that route were ever hit. If you'd rather retire the legacy controller/routes than keep testing them, happy to do that instead.

Roardom:
Thanks for your continued work on this!

> While digging I also noticed nothing links to the `users.warnings.*` routes anymore (the profile UI goes through the `UserWarnings` Livewire component), and `User\WarningController::store()` still inserts the old `torrent` column, which would throw an unknown-column QueryException if that route were ever hit. If you'd rather retire the legacy controller/routes than keep testing them, happy to do that instead.

This is a great point. I'm surprised this wasn't caught sooner.

Can you revert back this PR to the single commit with the single line change to update it from `torrent` to `torrent_id`. We can remove that unused controller in another PR alongside its tests. Also can you please use the commit style guidelines when rebasing (`(Fix) ` -> `fix: `).

skavinski-lab:
Done: reverted to the single one-line commit (`torrent` -> `torrent_id`), rebased onto current development, and updated the commit message to the `fix:` style. Happy to do the legacy controller/route removal (with its tests) in a follow-up PR.

Roardom:
Thank you!

> Happy to do the legacy controller/route removal (with its tests) in a follow-up PR

Would definitely appreciate it if you have the time. If not, I'll get around to it at some point :)

PR #5509, opened 2026-07-04
(Fix) Null-safe ticket comment notifications

skavinski-lab:
Comment::postReply() notifies ticket participants with User::query()->find($id)->notify(...) for the staff_id, user_id and prior commenter. User uses SoftDeletes, so find() returns null for a deleted account and replying to the ticket fatals with "notify() on null", so the reply fails to post.

Added the null-safe operator so a deleted participant is skipped. The sibling Torrent/Article/Playlist/Request cases in the same method, and the ticket case in Comments.php, already use ?->notify (all introduced in d128cd523); only these three lines were missed.


Roardom (approved):

Roardom:
Thank you!

PR #5510, opened 2026-07-04
(Fix) Correct stale columns in Rss/Wish/UserSetting factories

skavinski-lab:
Several factories reference columns that later migrations renamed, relocated or dropped, so ->create() throws "Unknown column" and any test using them fails.

RssFactory: drop staff_id (removed in 2023_02_03_094806_update_rss_table).
WishFactory: replace tmdb/type/source with tmdb_movie_id (the tmdb column was split into tmdb_movie_id/tmdb_tv_id; type/source were dropped).
UserSettingFactory: drop chat_hidden (dropped), hidden and private_profile (moved to user_privacy), block_notifications (moved to user_notifications); add the required user_id, unbookmark_torrents_on_completion and torrent_sort_field so the factory can persist.

Verified each with a live ->create(), plus Pint and Larastan. Same class of bug as #5508.

### DIFF
--- a/database/factories/RssFactory.php
+++ b/database/factories/RssFactory.php
@@ -40,7 +40,6 @@ class RssFactory extends Factory
             'is_private'   => $this->faker->boolean(),
             'is_torrent'   => $this->faker->boolean(),
             'json_torrent' => $this->faker->word(),
-            'staff_id'     => User::factory(),
         ];
     }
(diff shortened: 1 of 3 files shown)


Roardom (approved):

Roardom:
Thanks!

PR #5527, opened 2026-07-15
(Add) Report replies

Roardom:


ericcox-kr:
Took this for a spin locally, the ticket-style thread direction is great. Four things that look like blockers before merge:

**1. Every page 500s for non-staff after migrating.** The non-modo branch of `hasUnreadReport` in `app/View/Composers/TopNavComposer.php` filters on `where('user_id', ...)`, but `reports` has no `user_id` column (it's `reporter_id`). Since the composer runs on every authenticated page render, regular members get SQLSTATE 42S22 on every page. The `hasUnreadTicket` block above it is fine because `tickets.user_id` exists, which is probably where this came from. Same copy-paste in `ReportController::show` (`$report->user_id` is always null there, so `user_read` never gets marked true and the nav flare never clears for reporters).

**2. Report visibility is inverted in search.** `app/Http/Livewire/ReportSearch.php:71` has `->when(auth()->user()->group->is_modo, fn ($query) => $query->where('reporter_id', '=', auth()->id()))`, which scopes moderators to their own reports and shows regular members the entire reports table (reporter identities, reported users, message bodies). Needs the restriction as the default callback (third arg) instead. The `@if (! $isModo)` around the reporter filter in the blade is inverted the same way.

**3. Migration backfills the wrong column and drops the data.** In `create_report_replies_table.php` the backfill inserts `'content' => $report->message` rather than `$report->verdict`, so every solved report gets the reporter's original message duplicated as a staff reply, and the actual verdict text is gone once `dropColumn('verdict')` runs. Might also be worth backfilling old solved reports as read, otherwise everyone who ever filed a report gets a permanent unread flare on deploy.

**4. Six leftover references to the removed `staff.reports.show` route:** `ReportAssigneeController.php:37,49`, `SnoozedReportController.php:34,47`, `NewReportAssigned.php:55`, and `torrent/partials/reports.blade.php:37`. Assign/snooze actions throw after the write, and the torrent page 500s for staff whenever a torrent has open reports.

Smaller stuff (stale `assigned_to` form field vs the new `staff_id` validation, `$this->verdict` leftover in `ReportSearch.php:78`, the old report tests/factory still targeting the dropped columns) happy to detail if useful.

Roardom:
Thanks for the review! I just pushed what I had last night as I was heading to sleep. Still have to go over it more thoroughly. And yeah the goal was to align reports similarly to how tickets are currently handled.

Roardom:
@ericskavinski I just finished making some more changes. Do you mind taking it for another spin?

PR #5530, opened 2026-07-18
(Fix) Strict numeric validation for announce fields

skavinski-lab:
The announce endpoint validated `uploaded`, `downloaded`, `left`, `numwant` and `corrupt` with only `is_numeric($value) || $value < 0`. Two problems:

- There's no upper bound. `uploaded=9000000000000000000` passes (it's under `PHP_INT_MAX`), and the delta over the stored peer base is credited straight into `users.uploaded` by `ProcessAnnounce`. One request with a valid passkey grants a huge amount of upload credit.
- Scientific notation slips through. `is_numeric('1e19')` is true and `'1e19' < 0` is false, but `(int) '1e19'` overflows to a negative value on 64-bit. That negative flows into the Redis peer/history batches and gets upserted into `unsignedBigInteger` columns, which fails the whole `AutoUpsertPeers`/`AutoUpsertHistories` transaction.

Fix: require plain non-negative decimal integers with `ctype_digit()`, capped at a new `MAX_ANNOUNCE_VALUE` constant of 1 PiB, well above any real per-session client total. Applied to `port`, `uploaded`, `downloaded`, `left`, `numwant` and `corrupt`. The now-redundant `ctype_digit()` in the port check is dropped since validation runs before it.

Tests: added cases for oversized values, scientific notation, negatives, floats, whitespace padding, hex, `MAX_ANNOUNCE_VALUE + 1`, and a passing boundary at exactly 1 PiB.

Conforming clients send decimal integers in range, so nothing changes for them. Malformed or out-of-range values now get failure 134 instead of being silently miscast.

### DIFF
--- a/app/Http/Controllers/AnnounceController.php
+++ b/app/Http/Controllers/AnnounceController.php
@@ -65,6 +65,15 @@ final class AnnounceController extends Controller
         6699,
     ];
 
+    /**
+     * Maximum accepted value for the uploaded/downloaded/left announce fields (1 PiB).
+     *
+     * These fields are cumulative per-session client totals. A legitimate
+     * session never approaches 1 PiB; anything larger indicates a tampered
+     * client attempting to forge transfer statistics.
+     */
+    private const int MAX_ANNOUNCE_VALUE = 1_024 ** 5;
+
     private const array HEADERS = [
         'Content-Type'  => 'text/plain; charset=utf-8',
         'Cache-Control' => 'private, no-cache, no-store, must-revalidate, max-age=0',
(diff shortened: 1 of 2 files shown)


Roardom:
Hi, just a heads up - would really appreciate it if you refrain from using AI prose when communicating with us maintainers. It's just so tiring to read.

skavinski-lab:
> Hi, just a heads up - would really appreciate it if you refrain from using AI prose when communicating with us maintainers. It's just so tiring to read.

Understand. I'll keep such more concise. My English is not the best, so it also helps there.

skavinski-lab:
CI is green and there are no conflicts with development. Ready for review when you have time.

PR #5531, opened 2026-07-18
(Fix) Stored HTML/JS injection in pages and wikis

skavinski-lab:
`Page::getContentHtml()` and `Wiki::getContentHtml()` run authored content through `Bbcode::parse()` (which escapes everything with `htmlspecialchars()`), then immediately call `htmlspecialchars_decode()` on the result. That undoes the escaping and restores any raw `<script>`, event handlers, etc. from the source. With `html_input => 'allow'` and `allow_unsafe_links => true` in `config/markdown.php`, that HTML and any `javascript:` links then render as-is.

Pages and wikis are visible to every authenticated user, and the rules page that everyone is required to view goes through the same pipeline. The default CSP blocks inline script, but CSP can be disabled via env, and stored HTML/CSS injection (phishing forms, UI redress on trusted pages) works regardless of CSP.

Fix:

- Drop the `htmlspecialchars_decode()` call in both models. BBCode-generated tags are never entity-escaped, so removing the decode only affects untrusted content.
- Set `allow_unsafe_links => false` so CommonMark drops `javascript:`/`vbscript:` destinations in Markdown `[text](url)` links, which BBCode's `sanitizeUrl()` doesn't cover.
- `html_input` stays `allow` so the BBCode tags survive the Markdown pass.

Tests: added raw-HTML and unsafe-link cases to `PageTest`, plus a new `WikiTest` with the same coverage.

Note: escaping now happens before the Markdown stage, so three raw-HTML Markdown constructs stop rendering: `>` blockquotes, `<url>` autolinks, and quoted link titles `[text](url "title")`. The BBCode equivalents (`[quote]`, `[url]`) still work, as does all other formatting. A whitelist sanitizer on the output could preserve them, but that's a new dependency, so I left it out; happy to add it as a follow-up if those constructs are used in practice.


skavinski-lab:
@Roardom rebased onto current `development` (7ca72a2, post-#5587). Clean rebase, no conflicts, same 5-file diff.

Local run on PHP 8.5.4 / MySQL 8.4: `PageTest` and `WikiTest` pass, and Pint and Larastan are clean on the changed files. The full Pest suite passes apart from one `MarkdownExtraTest` "html elements" case that fails identically on plain `development` in that environment (libxml 2.15 DOM edge case) and is unrelated to this change; CI here will show the definitive result.

The `htmlspecialchars_decode()` calls are still present on `development` (`app/Models/Page.php:63` and the same line in `Wiki.php`), so the stored-HTML issue described above is still live. A note is also going to the address in `SECURITY.md`, per the policy there. Happy to adjust the approach if you would rather keep the raw-HTML Markdown constructs and add a whitelist sanitizer instead.

Roardom:
The reason I haven't replied to this yet is because this is actually a feature, not a bug. It allows staff to create wiki pages that are interactive. Unfortunately, as you've noted, the csp has prevented this use case for awhile now. I wrote a few of my own wikis with js awhile ago that were broken with that change. However, with alpine.js's new csp mode in use, this interactivity might actually be functional again, which would be great as it would allow basic js interactivity but nothing malicious.

skavinski-lab:
Thanks for the context. You are right that the write path is staff-only (`CheckForModo` on the dashboard routes), and I had missed that the raw-HTML passthrough is deliberate for interactive wikis. The PR body overstated the exposure: for ordinary users this is not reachable, and with the default CSP on, inline script and `javascript:` navigation are already blocked.

The one residual case is a compromised staff account getting stored script onto pages every user must open, but that is a design trade-off you have already weighed, not something this PR should force. Closing. Happy to help if you ever want an opt-in config flag for raw HTML in pages and wikis.

PR #5536, opened 2026-07-25
feat: dual-stack (IPv4/IPv6) peer columns and views

lluiseduardo-silva:
# feat: dual-stack (IPv4/IPv6) peer columns and views

## Summary

Surfaces both IP families for peers, to pair with the dual-stack
UNIT3D-Announce tracker. Adds `ipv4`/`ipv6` columns to the `peers` table and
shows both addresses — and per-family connectability — in the peer and profile
views.

> **Companion PR** — the tracker that populates these columns is in
> `lluiseduardo-silva/UNIT3D-Announce` (branch [`feat/dual-stack-ipv6`](https://github.com/lluiseduardo-silva/UNIT3D-Announce/tree/feat/dual-stack-ipv6)) The two
> are meant to ship together. [PR](https://github.com/Roardom/UNIT3D-Announce/pull/39)

## Changes

- **migration**: add nullable `ipv4`/`ipv6` address, `*_port` and
  `*_connectable` columns to `peers`.
- **torrent peers view**: show a peer's IPv4 and IPv6 endpoints stacked in one
  row, and report connectability **per family** (`IPv4: Yes/No`, `IPv6: Yes/No`)
  instead of a single ambiguous flag.
- **user profile** ("Clients and IP-addresses"): list both address families for
  a client.
- **Peer model / TorrentPeerController**: select and cast the per-family fields.

## Screenshots

- **Torrent peers** — IPv4/IPv6 addresses stacked per peer + the per-family
  `Connectable` column.
    
Captura de tela 2026-07-25 050903

- **User profile** — both families under "Clients and IP-addresses".

Captura de tela 2026-07-25 053145

- **Before/after** — the same peer/torrent previously appeared multiple times
  because stale peers were never marked inactive in the DB; resolved by the
  companion tracker fix.
  
Captura de tela 2026-07-24 190658

  "Active torrents"; after = clean peer view
  
image

## Notes

- Meaningful with the external dual-stack tracker enabled
  (`announce.external_tracker.is_enabled`) — the per-family connectivity data
  comes from the tracker's connect-back check.
- The `Connectable` column is gated by `announce.connectable_check` (default
  off, unchanged here).
- Migration only adds nullable columns; existing single-stack peers are
  unaffected.

### DIFF
--- a/app/Http/Controllers/TorrentPeerController.php
+++ b/app/Http/Controllers/TorrentPeerController.php
@@ -32,10 +32,18 @@ class TorrentPeerController extends Controller
 
         return view('torrent.peers', [
             'torrent' => $torrent,
+            // Dual-stack aware: ONE row per peer (it is a single peer with a
+            // single progress), but exposing BOTH IP families so the blade can
+            // show the IPv4 and IPv6 addresses together in one entry.
             'peers'   => Peer::query()
                 ->with('user.group')
-                ->select(['torrent_id', 'user_id', 'uploaded', 'downloaded', 'left', 'port', 'agent', 'created_at', 'updated_at', 'seeder', 'active', 'visible', 'connectable'])
-                ->selectRaw('INET6_NTOA(ip) as ip')
+                ->select(['torrent_id', 'user_id', 'uploaded', 'downloaded', 'left', 'agent', 'created_at', 'updated_at', 'seeder', 'active', 'visible', 'connectable'])
+                ->selectRaw('INET6_NTOA(ipv4) as ipv4_ip')
+                ->selectRaw('ipv4_port')
+                ->selectRaw('ipv4_connectable')
+                ->selectRaw('INET6_NTOA(ipv6) as ipv6_ip')
+                ->selectRaw('ipv6_port')
+                ->selectRaw('ipv6_connectable')
                 ->where('torrent_id', '=', $id)
                 ->orderByRaw('user_id = ? DESC', [$request->user()->id])
                 ->orderByDesc('active')
--- a/app/Http/Controllers/User/UserController.php
+++ b/app/Http/Controllers/User/UserController.php
@@ -95,18 +95,28 @@ class UserController extends Controller
             'boughtUpload' => BonTransactions::where('sender_id', '=', $user->id)->where([['name', 'like', '%Upload%']])->sum('cost'),
             // 'boughtDownload'        => BonTransactions::where('sender_id', '=', $user->id)->where([['name', 'like', '%Download%']])->sum('cost'),
             'invitedBy' => Invite::where('accepted_by', '=', $user->id)->first(),
-            'clients'   => $user->peers()
-                ->join('torrents', 'torrents.id', '=', 'peers.torrent_id')
-                ->select('agent', 'port')
-                ->selectRaw('INET6_NTOA(peers.ip) as ip')
-                ->selectRaw('MIN(peers.created_at) as created_at')
-                ->selectRaw('MAX(peers.updated_at) as updated_at')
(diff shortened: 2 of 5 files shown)


Roardom:
Just wanted to point out a few things:

- This code doesn't actually do anything. It reads stuff that is never written to. It assumes changes made to UNIT3D-Announce, but that is an external project. The PHP code within UNIT3D still needs to work and it currently doesn't.
- This chosen schema is undesirable. I specified the schema to be used in the linked issue.
- This PR is missing practically all of the prerequisites I spent time listing in the linked issue.
- Do you use this code in prod at scale? I need to know if this is battle tested code, or if you've only tested it locally, or no testing at all. This refers to the PR to UNIT3D-Announce as well.
- Lastly, I no longer believe dual-stack is appropriate for UNIT3D given the associated issues and am no longer inclined to accept a PR adding support for it.

lluiseduardo-silva:
> Just wanted to point out a few things:
> 
> * This code doesn't actually do anything. It reads stuff that is never written to. It assumes changes made to UNIT3D-Announce, but that is an external project. The PHP code within UNIT3D still needs to work and it currently doesn't.
> * This chosen schema is undesirable. I specified the schema to be used in the linked issue.
> * This PR is missing practically all of the prerequisites I spent time listing in the linked issue.
> * Do you use this code in prod at scale? I need to know if this is battle tested code, or if you've only tested it locally, or no testing at all. This refers to the PR to UNIT3D-Announce as well.
> * Lastly, I no longer believe dual-stack is appropriate for UNIT3D given the associated issues and am no longer inclined to accept a PR adding support for it.

Thanks for the detailed feedback and to be clear upfront, I accept the decision. Replying point by point for the record, plus one clarifying question.

**"This code doesn't actually do anything"** fair, and you're right about the framing. Within this repo, nothing writes these columns; only the companion UNIT3D-Announce PR does. That coupling was deliberate (the two were meant to ship together), but I understand that from UNIT3D's perspective a PR whose columns are dead on any install not running a patched external tracker isn't acceptable, and that the built-in PHP announce would need to support this too for it to be a real UNIT3D feature. Point taken.

**Schema** one genuine clarifying question, since this is the part I'd like to learn from: am I right that the schema you specified is the `peers` + `peer_ips` split (arbitrary endpoints in a child table)? I did read the issue and deliberately went with the narrower one-slot-per-family model to keep the migration purely additive and existing queries untouched but I understand the spec in the issue is the project's design, not a suggestion. If you ever have a moment, I'd honestly be curious what makes the generalized model worth its complexity to you given the common one-v4/one-v6 case; asking to understand the project's design philosophy, not to relitigate this PR.

**Prerequisites** also fair. I treated several items from the issue as follow-ups rather than blockers; I see now they were scoped as requirements.

**Testing** honest answer: local only, not prod at scale. Real dual-stack environment with the patched tracker and qBittorrent connecting to the same seeder over v4 and v6 simultaneously, plus the raw curl announces in the evidence doc. That's functional verification, not battle-testing, and I won't claim otherwise. Same status for the UNIT3D-Announce PR.

Given you no longer consider dual-stack appropriate for UNIT3D, feel free to close both PRs. I'll maintain the branches on my fork for anyone running the dual-stack tracker setup. Thanks for taking the time to review and to write out your reasoning genuinely appreciated.

Roardom:
> what makes the generalized model worth its complexity to you given the common one-v4/one-v6 case

The system would be less complex than your PR. I would say your solution is actually more complex given all the edge cases it has to continually handle both ipv4 and ipv6 instead of making that generic.

Roardom:
Also, just an fyi, you'll probably encounter deadlocks with your peer deactivation queue at scale. You should never have multiple queues for the same database table or it will lock up.

PR #5553, opened 2026-08-04
(Add) User opt-out for Torrent Deleted Notifications

heathlarsen:
_**Copied from my post on Blutopia:**_

I get many many notifications per week for "Torrent deleted" or "Bulk torrents deleted".
My inbox is drowning in them.
I end up having to click "Mark all messages as read" but this means I often miss important messages not related to torrent deletion.

I use qui for qbittorrent and it handles removing deleted torrents for me, but even if I didn't there are easier ways to remove deleted torrents from any torrenting program than manually cross-referencing with these notifications.

I am hoping an option could be added to "Settings > Notification > Torrents" to opt out of these notifications.

For what it's worth, this option was recently added to Nebulance to much fanfare. It would be great seeing it here too.

### DIFF
--- a/app/Http/Requests/UpsertUserNotificationRequest.php
+++ b/app/Http/Requests/UpsertUserNotificationRequest.php
@@ -45,6 +45,7 @@ class UpsertUserNotificationRequest extends FormRequest
             'show_torrent_comment'         => 'required|boolean',
             'show_torrent_thank'           => 'required|boolean',
             'show_torrent_tip'             => 'required|boolean',
+            'show_torrent_deleted'         => 'required|boolean',
             'show_mention_torrent_comment' => 'required|boolean',
             'show_mention_request_comment' => 'required|boolean',
             'show_mention_article_comment' => 'required|boolean',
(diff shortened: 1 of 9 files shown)


Roardom (changes requested):
Hi. Thanks for the PR.

Please only include languages you're fluent in. I can't say I know every single language, but I'm familiar with at least one other than English and your translation is completely botched. You have to be very careful when using bittorrent-related terms like "seeding" and "leeching" as they can't be translated literally. It's best to leave those with someone that speaks the language.

heathlarsen:
Sorry, I wasn't sure what the expectations were here.
I've removed all except English. 🙂

heathlarsen:
@Roardom I've rebased this back onto `development` to resolve new merge conflicts and keep it fresh.

Roardom:
Thanks for the updates. What has me currently paused on accepting this PR is that torrent deletion messages are system pms, they're not notifications like everything else that use the functions used in this PR. I'm not sure how I want to go about this yet.

It's also possible to go about this a completely different route and use #4889 with a separate page that keeps track of torrents you're seeding that were deleted, and provide info of why they were deleted, but not specifically send a notification for them.

heathlarsen:
I'm happy with any solution, I'm just desperate for one. 😅

PR #5575, opened 2026-08-14
(Fix) Modal interaction issues on mobile devices.

networkbebop:
# Mobile Modal and Profile Overflow Fixes

### Cause

At viewport widths below `1100px`, `.panel__action` elements are hidden by default and displayed only while their `.panel__actions` container is hovered or focused.

```scss
.panel__action {
    display: none;
}

.panel__actions {
    &:hover,
    &:focus {
        .panel__action {
            display: block;
        }
    }
}
```

Mobile device failure:

- Touch devices inability to maintain hover reliably.
- `.panel__actions` is a non-focusable `<div>`, so `:focus` did not apply.
- Touching a field inside the dialog cleared the simulated hover.
- The containing `.panel__action` returned to `display: none`, hiding the dialog along with its ancestor.

### Fix

```scss
.panel__actions {
    &:hover,
    &:focus-within,
    &:has(.dialog:popover-open) {
        .panel__action {
            display: block;
        }
    }
}
```

- `:focus-within` keeps the action visible while its trigger or any dialog control has focus.
- `:has(.dialog:popover-open)` keeps it visible whenever its dialog is open, including interactions with nonfocusable content.
- Native popover behavior (e.g., Escape, Cancel, outside tap) are unchanged.
- No JavaScript or markup reordering needed.

---

### DIFF
--- a/resources/sass/components/_panel.scss
+++ b/resources/sass/components/_panel.scss
@@ -110,7 +110,8 @@
         }
 
         &:hover,
-        &:focus {
+        &:focus-within,
+        &:has(.dialog:popover-open) {
             flex-direction: column;
             align-items: stretch;
 
--- a/resources/sass/pages/_user.scss
+++ b/resources/sass/pages/_user.scss
@@ -8,22 +8,14 @@
 
 @media screen and (max-width: 500px) {
     .profileV2 {
-        grid-template-columns: minmax(0, 1fr);
-        grid-template-areas: 'avatar' 'username' 'registration' 'title' 'about';
+        grid-template-columns: auto 1fr;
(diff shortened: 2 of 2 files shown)


Roardom (changes requested):
Hi, thanks for the PR! I really like this modal solution. I tested it locally and can confirm it works.

However, I'd like to solve the profile issue differently. Could you shift that to a separate commit and PR?

Specifically, I'd like to avoid wrapping the username. Perhaps using SVG to set a max width (but to only scale down proportionally if it's too big, not scaled up if it's a short username) but I haven't tested locally yet to see if that would work.

Roardom (approved):

PR #5593, opened 2026-09-18
Add default user for request claims

Obi-Wana:
Prevent error 500 on requests.show when a soft deleted user has claimed requests


app/Models/TorrentRequestClaim.php
```diff
@@ -75,7 +75,10 @@ protected function casts(): array
      */
     public function user(): BelongsTo
     {
-        return $this->belongsTo(User::class);
+        return $this->belongsTo(User::class)->withDefault([
+            'username' => 'System',
+            'id'       => User::SYSTEM_USER_ID,
+        ]);
```
Roardom:
```suggestion
        return $this->belongsTo(User::class)->withTrashed();
```

Roardom (changes requested):
Been trying to remove the default user stuff and instead make the soft deleted users included instead. The db schema should have foreign key constraints on all the user_id columns by now.

Obi-Wana:
> Been trying to remove the default user stuff and instead make the soft deleted users included instead. The db schema should have foreign key constraints on all the user_id columns by now.

Agree.

Roardom (changes requested):
Thanks. Looks good now! Please squash and rename the commit and pr title to match the guidelines and I'll merge.

